PR #2, opened 2015-10-09
More tslint rules

ef4:


stefanpenner:
https://github.com/tildeio/glimmer/pull/27 includes these (but updated to current active branch/code)

PR #7, opened 2015-11-07
Correct invalid import paths in `index.d.ts` files.

### LINKED ISSUE #6: Invalid import paths in `index.d.ts` files.
The import paths for the `index.d.ts` files are incorrect in the transpiled output.

The transpiled structure is:

```
% tree dist/es6/ -L 1
dist/es6/
├── htmlbars-compiler
├── htmlbars-compiler.js
├── htmlbars-object
├── htmlbars-object.js
├── htmlbars-reference
├── htmlbars-reference.js
├── htmlbars-runtime
├── htmlbars-runtime.js
├── htmlbars-syntax
├── htmlbars-syntax.js
├── htmlbars-test-helpers
├── htmlbars-test-helpers.js
├── htmlbars-util
├── htmlbars-util.js
├── htmlbars.js
├── simple-html-tokenizer
└── simple-html-tokenizer.js
```

The contents of `htmlbars-compiler.js` is:

```js
import {
  compile,
  compileSpec,
  template,
} from "./lib/compiler";

export { default as TemplateCompiler } from './lib/template-compiler';
export { default as TemplateVisitor } from './lib/template-visitor';

export {
  compile,
  compileSpec,
  template
};
```

There is no `./lib` folder in this context.  The tests happen to work now because the loader being used had a bug.

rwjblue:
Prior to this change, the `index.d.ts` files were moved up one directory (and named for their package) but the import paths were not being updated so that the resulting ES6 or AMD output referenced invalid paths.

This change adds define.alias shims for each of the packages, and leaves the `index.d.ts` files within the same level as they are on disk (therefore keeping the import paths correct).

In Ember we will need a little more work during integration (specifically the Ember loader will need to either gain aliases or support for `index` fallback).

Fixes #6.

---

Tested via:

```
ember test --server --no-launch
```

```
open http://localhost:7357/3586/tests/index.html?hidepassed&packages=htmlbars-runtime&module=HTMLBarsComponent%20-%20invocation
```

### DIFF
--- a/build-support/shims/index-aliases.js
+++ b/build-support/shims/index-aliases.js
@@ -1,0 +1,10 @@
+/* globals define */
+
+enifed('htmlbars', enifed.alias('htmlbars/index'));
+enifed('htmlbars-compiler', enifed.alias('htmlbars-compiler/index'));
+enifed('htmlbars-syntax', enifed.alias('htmlbars-syntax/index'));
+enifed('htmlbars-util', enifed.alias('htmlbars-util/index'));
+enifed('htmlbars-runtime', enifed.alias('htmlbars-runtime/index'));
+enifed('htmlbars-reference', enifed.alias('htmlbars-reference/index'));
+enifed('htmlbars-test-helpers', enifed.alias('htmlbars-test-helpers/index'));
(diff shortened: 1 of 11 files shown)


rwjblue:
Closing and reworking

PR #17, opened 2015-12-07
Replace references of "HTMLbars" with "Glimmer"

HeroicEric:
- Replaces references of "HTMLbars" with "Glimmer"


chancancode:
Unfortunately, `ARCHITECTURE.md` is largely outdated at this point :P Maybe we should just delete it for now so no one would be confused? We can always get it back from version control and update it when things stabilizes

HeroicEric:
@chancancode makes sense to me if you think it will be misleading. I just read through it and assumed it was up to date so I was mislead :smile: 

I moved the change outside of `ARCHITECTURE.md` to https://github.com/tildeio/glimmer/pull/18

HeroicEric:
I put up another PR that removed this doc entirely. The choice is yours :smiley:  https://github.com/tildeio/glimmer/pull/19

PR #21, opened 2015-12-07
Don't alloc during opcode creation

chadhietala:
Probably minimal impact but doesn't degrade code quality a ton. Also replaced some `var` -> `let`.

### DIFF
--- a/packages/glimmer-compiler/lib/template-compiler.ts
+++ b/packages/glimmer-compiler/lib/template-compiler.ts
@@ -252,9 +252,13 @@ export default class TemplateCompiler {
 
   openElement([action]) {
     this.opcode('openElement', action, action.tag, action.blockParams);
+    for (let i = 0; i < action.attributes.length; i++) {
+      this.attribute([action.attributes[i]]);
+    }
 
-    action.attributes.forEach(attr => this.attribute([attr]));
-    action.modifiers.forEach(modifier => this.modifier([modifier]));
+    for (let i = 0; i < action.modifiers.length; i++) {
+      this.modifier([action.modifiers[i]]);
+    }
   }
 
   closeElement() {
@@ -264,10 +268,11 @@ export default class TemplateCompiler {
   component([action]) {
     let { attributes, tag } = action;
 
-    attributes.forEach(({ name, value }) => {
+    for (let i = 0; i < attributes.length; i++) {
+      let { name, value } = attributes[i];
       this.prepareAttributeValue(value);
       this.opcode('pushLiteral', name, name);
-    });
+    }
 
     this.opcode('prepareObject', null, attributes.length);
 
@@ -294,10 +299,10 @@ export default class TemplateCompiler {
   }
 
   modifier([action]) {
-    let { path } = action;
+    let { path: { parts } } = action;
 
     this.prepareHelper(action);
-    this.opcode('modifier', action, path.parts);
+    this.opcode('modifier', action, parts);
   }
 
   mustache([action]) {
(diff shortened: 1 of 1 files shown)


packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -407,16 +412,16 @@ export default class TemplateCompiler {
...
     if (!pairs.length) {
       this.opcode('pushLiteral', null, null);
       return;
     }
 
-    for (var i = pairs.length - 1; i >= 0; i--) {
-      var key = pairs[i].key;
-      var value = pairs[i].value;
+    for (let i = pairs.length - 1; i >= 0; i--) {
+      let key = pairs[i].key;
```
stefanpenner: `let { key, value } = pairs[i];`

chadhietala: Fixed.

stefanpenner:
LGTM, mergable when CI is green.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -264,10 +268,11 @@ export default class TemplateCompiler {
   component([action]) {
     let { attributes, tag } = action;
 
-    attributes.forEach(({ name, value }) => {
+    for (let i = 0, l = attributes.length; i < l; i++) {
```
stefanpenner: lets not cache `length`

stefanpenner:
one comment, then it should be good to go.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -252,9 +252,13 @@ export default class TemplateCompiler {
 
   openElement([action]) {
     this.opcode('openElement', action, action.tag, action.blockParams);
+    for (let i = 0, l = action.attributes.length; i < l; i++) {
+      this.attribute([action.attributes[i]]);
+    }
 
-    action.attributes.forEach(attr => this.attribute([attr]));
-    action.modifiers.forEach(modifier => this.modifier([modifier]));
+    for (let i = 0, l = action.modifiers.length; i < l; i++) {
```
stefanpenner: lets not cache length

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -252,9 +252,13 @@ export default class TemplateCompiler {
 
   openElement([action]) {
     this.opcode('openElement', action, action.tag, action.blockParams);
+    for (let i = 0, l = action.attributes.length; i < l; i++) {
```
stefanpenner: lets not cache length

chadhietala:
I've been sort of working on a bunch of babel transforms to hopefully make this a non manual endeavor. https://github.com/chadhietala/babel-array-method-zero-cost-abstractions

stefanpenner:
@chadhietala cool, the tricky bit is knowing if the given `forEach` is being called on an array or not.

chadhietala:
:grimacing: yea...

stefanpenner:
@chadhietala a different approach, i took some time ago was to use sweet.js style macro, i think the same approach could work here nicely.

```js
import { forEach } from 'zero-cost-abstractions';

forEach(array, callback) 
```

Then use babel's matcher stuff to detect which forEach it is, now the developer has basically annotated which `forEach`/`map` etc it want's to optimize. such macro fun.

chadhietala:
Yea annotation is what I was thinking

chadhietala:
So on my flight I took a whack at the macros. Was more of experiment in writing sweet.js macros then anything that I would recommending using at this point. https://github.com/chadhietala/array-method-macros

PR #25, opened 2015-12-21
Tmp fix

stefanpenner:
Will move away from the ember-cli branch once it is merged  + released.


stefanpenner:
failures seem unrelated, if they turn out to be let me know i'll dig in further

PR #27, opened 2015-12-22
More build things

stefanpenner:
- speed up build
- remove unused deps
- cleanup build
- upgrade deps
- fix as many tslint errors as appear to not be bugs in tslint.


stefanpenner:
shrinkwrap was removed, because it appears NPM can't shrinkwrap the version of typescript....

PR #30, opened 2016-01-06
Integration

stefanpenner:
- [ ] phantom tests need TLC
- [ ] ...

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -13,9 +13,8 @@ function transpile(tree, label) {
 }
 
 module.exports = function() {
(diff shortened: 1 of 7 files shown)


ember-cli-build.js
```diff
@@ -137,7 +139,10 @@ module.exports = function() {
     destDir: '/assets'
   });
 
-  var es6Tree = find(packagesTree, {
+  var es6Tree = find(merge([
```
chadhietala: Doesn't `find` do a merge when passed an array?

PR #49, opened 2016-01-24
Make consumable from Ember without `npm link`.

rwjblue:
Prior to these changes a few things were missing from a build when just `npm install`'ed (`benchmark` node library, all of bower components libs, etc). 

This PR changes things around so that if devDeps or bower components are missing the main package libs still build properly.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -1,5 +1,7 @@
 /* globals __dirname */
 
+var path = require('path');
+var existsSync = require('exists-sync');
 var concat = require('broccoli-concat');
 var merge = require('broccoli-merge-trees');
 var typescript = require('broccoli-typescript-compiler');
@@ -17,6 +19,7 @@ function transpile(tree, label) {
 module.exports = function() {
   var packages = __dirname + '/packages/node_modules';
   var bower = __dirname + '/bower_components';
+  var hasBower = existsSync(bower);
 
   var tsOptions = {
     tsconfig: {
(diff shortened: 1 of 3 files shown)


rwjblue:
@stefanpenner r?

ember-cli-build.js
```diff
@@ -64,15 +67,22 @@ module.exports = function() {
...
       include: ['benchmark.js'],
       destDir: 'demos'
-    })
-  ]);
+    }));
+  }
+  var demos = merge(demoTrees);
 
+
+  var tokenizerPath = path.join(require.resolve('simple-html-tokenizer'), '..', '..', 'lib');
```
stefanpenner: https://github.com/stefanpenner/broccoli-stew/blob/master/lib/npm.js#L17 may fit here.

ember-cli-build.js
```diff
@@ -176,19 +189,24 @@ module.exports = function() {
...
     demos,
     glimmerCommon,
     glimmerCompiler,
     glimmerRuntime,
     glimmerTests
-  ]);
+  ];
+
+  if (hasBower) {
+    var loader = find(bower, {
```
stefanpenner: not required for this PR, but maybe we should grab loader from npm .

rwjblue: yes, agreed

stefanpenner:
LGTM, left some nit-picks but they shouldn't block this.

rwjblue:
Thanks @stefanpenner!  I'll poke at Stew's `npm` for a minute to see if I can just use it out of the box.

Issue #44, opened 2016-01-20
Lazy component compilations


chancancode:
Currently, `TestEnvironment#registerComponent` eagerly compiles the component layout, which causes it to try to lookup other component definitions that might not have been loaded yet (which then causes the compiler to emit `OpenPrimitiveElement`).

(This is just a `TestEnvironment` implementation issue, so it shouldn't require any Glimmer changes)

chadhietala:
Working on this one.

chadhietala:
Closed by #50

PR #50, opened 2016-01-25
Lazy layout compilation

chadhietala:
WIP: This causes test failures but not entirely sure why. Would like some :eyes: .

Instead of eagerly compiling the template, compile the layout and create
the definition lazily. Will address #44

### DIFF
--- a/packages/node_modules/glimmer-test-helpers/lib/environment.ts
+++ b/packages/node_modules/glimmer-test-helpers/lib/environment.ts
@@ -240,9 +240,17 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
 
 const EMBERISH_CURLY_COMPONENT_MANAGER = new EmberishCurlyComponentManager();
 
+interface IDefinitionLookup {
+  definition: BasicComponentDefinition | EmberishCurlyComponentDefinition | EmberishGlimmerComponentDefinition;
+  layout: string;
+  factory: BasicComponentFactory | EmberishCurlyComponentFactory | EmberishGlimmerComponentFactory,
+  manager: BasicComponentManager | EmberishCurlyComponentManager | EmberishGlimmerComponentManager
+}
+
 export class TestEnvironment extends Environment {
   private helpers = {};
   private components = dict<ComponentDefinition<any>>();
+  private componentDefinitions = dict<IDefinitionLookup>();
 
   constructor(doc: HTMLDocument=document) {
     super(new DOMHelper(doc), Meta);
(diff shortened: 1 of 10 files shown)


chancancode:
I think I know why this isn't working: this patch defers adding the definitions to the `this.components` array and do that lazily at `getComponentDefinition`. However, the check in `hasComponentDefinition` is still based on the `this.components` array. So you will still have the original problem – when compiling a layout that references other components, those referenced components might not have been added to the array yet, causing them to be compiled into `OpenPrimitiveComponents`.

Here is what I propose instead: we can keep

chancancode:
Oops!

Here is what I propose instead: we can keep the lazy compilation logic encapsulated inside the `ComponentDefinition`:

1. Remove the public layout field in the abstract ComponentDefinition interface
2. Remove the layout field in the interface constructor
3. Remove the layout field in the abstract `compile` method (I think this is wrong, I don't think it's actually used and it's probably just a bug)

Now, the only public contract for the abstract `ComponentDefinition` interface is that it needs to be able to compile itself through the `compile` method, which takes the environment (among other things).

Then, in the concrete implementation (`EmberishGlimmerComponentDefinition`, etc), we can add a mandatory constructor option that takes the raw layout string and store it as a field on the instance. Then, on `compile`, we can used that stored string to compile the Layout (and memorize it).

Does that make sense?

chancancode:
That approach would require less code, I think. The main benefit is that we will be able to construct the actual `ComponentDefinition` objects and add them to the `this.components` array without actually compiling them. Then we can avoid introducing another new kind of object just to hold the parts, and all the existing code (`hasComponentDefinition` etc) would just work.

chancancode:
We might have to add a `.getLayout(env: Environment)` on the `ComponentDefinition` interface if there are other code that currently access `.layout` directly.

chadhietala:
So I refactored to account for your suggestions and it does seem better, however there are some breaking tests. I think they are all related to these lines https://github.com/tildeio/glimmer/blob/master/packages/node_modules/glimmer-runtime/lib/syntax/core.ts#L542-L549. Still investigating.

chadhietala:
So the test in question is https://github.com/tildeio/glimmer/blob/master/packages/node_modules/glimmer-runtime/tests/ember-component-test.ts#L1273. Since items are compiled lazily the code I linked to above will never drop into the `OpenPrimitiveElement` case as we eagerly place items into the component dict. On master the first occurrence of the custom element will always default to the `OpenPrimitiveElement` then the next occurrence will fall into the registered case.

Is this even a valid test? I was under the assumption that component layouts needed an outer wrapping element. If it is a valid test should the output be `<div>In layout</div>` since `non-block` is a registered component?

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -403,8 +403,24 @@ interface BasicComponentFactory {
...
-    let { program } = layout;
+  public compiledLayout : CompiledLayout;
+  public layout : string;
+
+  constructor(name: string, manager: ComponentManager<any>, ComponentClass: ComponentClass, layout) {
+    super(name, manager, ComponentClass);
+    this.layout = layout;
+  }
+
+  getLayout(env) {
```
chancancode: Can you type this? `getLayout(env: Environment): Layout {` or so

chancancode: Does this not have to be on the interface/abstract class? (If we call it outside of the test-helper file, then it probably needs to be on the abstract class)

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -403,8 +403,24 @@ interface BasicComponentFactory {
 }
 
 abstract class GenericComponentDefinition<T extends Component> extends ComponentDefinition<T> {
-  compile({ layout }: { layout: CompiledLayout }): ComponentParts {
-    let { program } = layout;
+  public compiledLayout : CompiledLayout;
+  public layout : string;
```
chancancode: these two can be private now that we have a getter

PR #54, opened 2016-01-28
Add Conditional References

### LINKED ISSUE #38: Revert and improve 0e83fa3
0e83fa3 

The check should happen inside the `TestOpcode`, and it should delegate to the `Environment` for the specific definition of "truthy-ness"

chadhietala:
This introduces the concept of a conditional reference. This is needed to properly support an `EmberishConditionalReference`. An example of this would be that Ember considers empty arrays in `if` blocks as falsy. This also includes `{{#unless}}` syntax for testing.
Closes #38

### DIFF
--- a/packages/node_modules/glimmer-runtime/lib/environment.ts
+++ b/packages/node_modules/glimmer-runtime/lib/environment.ts
@@ -2,8 +2,8 @@ import { Statement as StatementSyntax } from './syntax';
 import { Layout as CompiledLayout } from './compiled/blocks';
 
 import { DOMHelper } from './dom';
-
-import { NULL_REFERENCE } from './references';
+import { Reference } from 'glimmer-reference';
+import { NULL_REFERENCE, ConditionalReference } from './references';
 
 import {
   Component,
@@ -98,6 +98,10 @@ export abstract class Environment {
   constructor(dom: DOMHelper, meta: MetaLookup) {
     this.dom = dom;
     this.meta = meta;
+  }
+
+  toConditionalReference(reference: Reference): ConditionalReference {
+    return new ConditionalReference(reference);
   }
 
   getDOM(): DOMHelper { return this.dom; }
--- a/packages/node_modules/glimmer-runtime/tests/updating-test.ts
+++ b/packages/node_modules/glimmer-runtime/tests/updating-test.ts
@@ -120,6 +120,21 @@ test("a simple implementation of a dirtying rerender", function() {
   QUnit.notStrictEqual(root.firstChild.firstChild.firstChild, valueNode, "The text node was not blown away");
 });
 
+test('The if helper should consider an empty array falsy', function() {
+  let object: any = { condition: [], value: 'hello world' };
+  let template = compile('<div>{{#if condition}}<p>{{value}}</p>{{else}}<p>Nothing</p>{{/if}}</div>');
+  let result = render(template, object);
+
+  equalTokens(root, '<div><p>Nothing</p></div>');
+
+  object.condition.push('thing');
+  result.rerender();
+  equalTokens(root, '<div><p>hello world</p></div>', "Initial render");
+  object.condition.pop();
+  result.rerender();
+  equalTokens(root, '<div><p>Nothing</p></div>');
+});
+
 test("a simple implementation of a dirtying rerender without inverse", function() {
   let object = { condition: true, value: 'hello world' };
   let template = compile('<div>{{#if condition}}<p>{{value}}</p>{{/if}}</div>');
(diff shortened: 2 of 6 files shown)


chancancode:
Yeah, I agree it is a bit weird.

The alternative is that we make the `toBool` (`toConditional`? `toConditionalReference`?) method on the environment take a reference and produce a `ConditionalReference` directly:

```typescript
// Interfaces in runtime
export class ConditionalReference implements Reference {
  private inner: Reference;

  constructor(inner: Reference) {
    this.inner = inner;
  }

  value(): boolean {
    return this.toBool(this.inner.value());
  }

  protected toBool(value: any): boolean {
    return !!value;
  }
}

abstract class Environment {
  toConditional(reference: Reference): ConditionalReference {
    return new ConditionalReference(reference);
  }
}

// Implementation in test-helper
class EmberishConditionalReference extends ConditionalReference {
  protected toBool(value: any): boolean {
    if (Array.isArray(value)) {
      return value.length > 0;
    } else {
      return super.toBool(value);
    }
  }
}

class TestEnvironment {
  toConditional(reference: Reference): ConditionalReference {
    return new EmberishConditionalReference(reference);
  }
}
```

That way, the `ConditionalReference` won't have to hang on to the environment (unless the implementation chose to do it).

By the way, we rarely use the bare `Reference` interface, so I suspect we will eventually have to change it. However, I don't know what it should be yet and that part of the system is due for a revisit soon anyway.

Also, I don't think we need to put the `ConditionalReference` interface/super class in `glimmer-references` (that is for the generic reference interfaces, but this is very specific to our need in the environment, so it can probably just stay in `glimmer-runtime`).

packages/node_modules/glimmer-runtime/tests/updating-test.ts
```diff
@@ -120,6 +120,18 @@ test("a simple implementation of a dirtying rerender", function() {
...
+  let template = compile('<div>{{#if condition}}<p>{{value}}</p>{{else}}<p>Nothing</p>{{/if}}</div>');
+  let result = render(template, object);
+
+  equalTokens(root, '<div><p>hello world</p></div>', "Initial render");
+
+  object.condition = [];
+  result.rerender();
+  equalTokens(root, '<div><p>Nothing</p></div>');
+});
+
```
chancancode: Can you rename this to `The if helper should consider an empty array falsy` and add one for the `unless` helper too?

Perhaps it might be better in this shape:

1. Start with condition = `[]`
2. Rerender with no change
3. Rerender with condition.push("stuff");
4. Rerender with condition.pop(); (or `condition = []`?)
```

That way we will have adequate coverage for both the initial render and updating passes

**Comment by chadhietala** [comment]
@chancancode I liked the suggestions and went with `toConditionalReference`.

File: packages/node_modules/glimmer-reference/index.ts [hunk 2]
```diff
@@ -6,4 +6,5 @@ export { PushPullReference } from './lib/references/push-pull';
 export * from './lib/types';
 export { default as UpdatableReference, referenceFromParts } from './lib/references/root';
 export { ConstReference } from './lib/references/const';
+export { ConditionalReference } from './lib/references/conditional';
```
chancancode: Doesn't need to be in `glimmer-reference` (see comment above)

packages/node_modules/glimmer-runtime/lib/environment.ts
```diff
@@ -1,6 +1,6 @@
 import { Statement as StatementSyntax } from './syntax';
 import { Layout as CompiledLayout } from './compiled/blocks';
-
+import { Reference, ConditionalReference } from 'glimmer-reference';
```
chancancode: ditto

packages/node_modules/glimmer-runtime/lib/environment.ts
```diff
@@ -163,6 +167,10 @@ export abstract class Environment {
     };
   }
 
+  toBool(value) {
+    return !!value;
+  }
```
chancancode: :scissors:

packages/node_modules/glimmer-runtime/tests/updating-test.ts
```diff
@@ -120,6 +120,21 @@ test("a simple implementation of a dirtying rerender", function() {
...
+  equalTokens(root, '<div><p>Nothing</p></div>');
+
+  object.condition.push('thing');
+  result.rerender();
+  equalTokens(root, '<div><p>hello world</p></div>', "Initial render");
+  object.condition.pop();
+  result.rerender();
+  equalTokens(root, '<div><p>Nothing</p></div>');
+});
```
chancancode: Can you add one for `#unless`?

chadhietala: Done.

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -248,7 +260,7 @@ export class TestEnvironment extends Environment {
     super(new DOMHelper(doc), Meta);
 
     this.registerHelper("if", ([cond, yes, no]) => cond ? yes : no);
-    this.registerHelper("unless", ([cond, yes, no]) => cond ? no : yes);
+    this.registerHelper("unless", ([cond, yes, no]) => cond ? yes : no );
```
chancancode: should this be `!cond ? yes : no` (or `cond ? no : yes`)?

chancancode: By the way, we should probably extract the Emberish `toBool` logic locally in this file and reuse it to convert the conditional here too

But these helpers were just added quickly for the demos and don't really have tests yet, so ¯\_(ツ)_/¯

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -686,6 +704,60 @@ class RenderInverseIdentitySyntax extends StatementSyntax {
...
+  }
+
+  compile(compiler: CompileInto, env: Environment) {
+    //        Enter(BEGIN, END)
+    // BEGIN: Noop
+    //        PutArgs
+    //        Test
+    //        JumpIf(ELSE)
+    //        Evaluate(default)
+    //        JumpUnless(END)
```
chancancode: should be an unconditional Jump?

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -686,6 +704,60 @@ class RenderInverseIdentitySyntax extends StatementSyntax {
...
+    let END = new LabelOpcode({ label: "END" });
+
+    compiler.append(new EnterOpcode({ begin: BEGIN, end: END }));
+    compiler.append(BEGIN);
+    compiler.append(new PutArgsOpcode({ args: this.args.compile(compiler, env) }));
+    compiler.append(new TestOpcode());
+    if (this.templates.inverse) {
+      compiler.append(new JumpIfOpcode({ target: ELSE }));
+      compiler.append(new EvaluateOpcode({ debug: "default", block: this.templates.default }));
+      compiler.append(new JumpUnlessOpcode({ target: END }));
```
chancancode: ditto

packages/node_modules/glimmer-test-helpers/lib/environment.ts
```diff
@@ -249,7 +261,7 @@ export class TestEnvironment extends Environment {
     super(dom || new DOMHelper(document), Meta);
 
     this.registerHelper("if", ([cond, yes, no]) => cond ? yes : no);
-    this.registerHelper("unless", ([cond, yes, no]) => cond ? no : yes);
+    this.registerHelper("unless", ([cond, yes, no]) => cond ? no : yes );
```
chancancode: actually, seems like the only thing that is changed here is the ` `?

chancancode:
Looks good to me otherwise!

chadhietala:
@chancancode Uhhh not sure what happened. I think messed up the rebase because there were other issues. Would be nice is if tsc errors were failing tests.

PR #56, opened 2016-01-31
Introduce assertions for glimmer component layout

mixonic:
These assertions are added to `packages/node_modules/glimmer-test-helpers/lib/environment.ts`, which seems a bit odd but I believe is ok since those classes (`EmberishGlimmerComponentDefinition`) will eventually move into Ember.

Let me know if there is a better way to go about this.

### DIFF
--- a/packages/node_modules/glimmer-runtime/tests/component-test.ts
+++ b/packages/node_modules/glimmer-runtime/tests/component-test.ts
@@ -14,7 +14,6 @@ function compile(template: string) {
 
 function commonSetup() {
   env = new TestEnvironment(); // TODO: Support SimpleDOM
-  env.registerBasicComponent('my-component', MyComponent, "<div>{{yield}}</div>");
   root = rootElement();
 }
 
@@ -55,7 +54,8 @@ class MyComponent {
   }
 }
 
-QUnit.skip('creating a new component', assert => {
+QUnit.test('creating a new component', assert => {
+  env.registerBasicComponent('my-component', MyComponent, "<div>{{yield}}</div>");
   let template = compile("<my-component color='{{color}}'>hello!</my-component>");
   render(template, { color: 'red' });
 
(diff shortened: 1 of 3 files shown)


packages/node_modules/glimmer-runtime/tests/component-test.ts
```diff
@@ -71,7 +71,7 @@ QUnit.skip('the component class is its context', assert => {
   equalTokens(root, "<div color='green'><p>456</p>hello!</div>");
 });
 
-QUnit.skip('attrs are available in the layout', assert => {
+QUnit.test('attrs are available in the layout', assert => {
   env.registerBasicComponent('my-component', MyComponent, '<div><p>{{attrs.color}}</p>{{yield}}</div>');
```
chancancode: should be @color

packages/node_modules/glimmer-runtime/tests/component-test.ts
```diff
@@ -61,7 +61,7 @@ QUnit.skip('creating a new component', assert => {
   equalTokens(root, "<div color='green'>hello!</div>");
 });
 
-QUnit.skip('the component class is its context', assert => {
+QUnit.test('the component class is its context', assert => {
   env.registerBasicComponent('my-component', MyComponent, '<div><p>{{testing}}</p>{{yield}}</div>');
```
chancancode: Seems weird to assume you could register the component twice

chancancode:
I think the `BasicComponent` stuff were just a boot strapping mechanism and we were planning on removing that, perhaps we should just make sure we have the equivalent coverage in `emberish-component-test` and remove the `component-test` file?

chancancode:
For the assertion part, we can probably steal from 

https://github.com/emberjs/ember.js/blob/c54ecab568e6a904c8c3acca02972efb654637a2/packages/ember-htmlbars/lib/node-managers/component-node-manager.js#L91-L101

and 

https://github.com/emberjs/ember.js/blob/5b686d44e3cadb174ed30a2f00e605eca1ebf37f/packages/ember-htmlbars/tests/integration/component_invocation_test.js#L811-L841

mixonic:
I'm not so sure much of that directly translates. `layout.meta.fragmentReason` is from [compile time](https://github.com/emberjs/ember.js/blob/1a2cd166a86219717a0cde1df4dab497d8dda3e3/packages/ember-template-compiler/lib/system/compile_options.js#L43) and unless I'm mistake I don't think glimmer has that. Or needs it.

I'll see what I can do to better align this with the assertions already used in Ember, and look at removing basic (@wycats mentioned it should go away as well)

chancancode:
Yeah, I just mean we should have the same message/coverage for the assertions, not the specific implementation. AFAICT, all of those cases are still invalid today.

rwjblue:
@mixonic - Should this be closed?

PR #58, opened 2016-01-31
SVG: Bring back tests for basic behaviors

mixonic:
A first pass at ensuring proper SVG support. There seems to be an outstanding issue with nesting html content inside a `foreignObject` at the least.

### DIFF
--- a/packages/node_modules/glimmer-runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/node_modules/glimmer-runtime/lib/compiled/opcodes/dom.ts
@@ -178,7 +178,11 @@ export class NamespacedAttribute implements ElementPatchOperation {
   }
 
   apply(dom: DOMHelper, element: Element, lastValue: string = null): any {
-    let { value: reference, namespace } = this;
+    let {
+      name,
+      value: reference,
+      namespace
+    } = this;
     let value = reference.value();
 
     if (value === lastValue) {
--- a/packages/node_modules/glimmer-runtime/lib/syntax/core.ts
+++ b/packages/node_modules/glimmer-runtime/lib/syntax/core.ts
@@ -77,6 +77,7 @@ import {
   CloseElementOpcode,
   StaticAttrOpcode,
   DynamicAttrOpcode,
+  DynamicAttrNSOpcode,
   DynamicPropOpcode,
   CommentOpcode
 } from '../compiled/opcodes/dom';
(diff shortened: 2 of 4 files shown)


packages/node_modules/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -420,8 +421,13 @@ export class DynamicAttr extends AttributeSyntax {
...
   compile(compiler: CompileInto, env: Environment) {
-    compiler.append(new PutValueOpcode({ expression: this.value.compile(compiler, env) }));
-    compiler.append(new DynamicAttrOpcode(this));
+    let {namespace, value} = this;
+    compiler.append(new PutValueOpcode({ expression: value.compile(compiler, env) }));
+    if (namespace) {
+      compiler.append(new DynamicAttrNSOpcode(this));
+    } else {
+      compiler.append(new DynamicAttrOpcode(this));
+    }
```
chancancode: Good catch! I guess that means we didn't have tests for this case before :fearful:

packages/node_modules/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -178,7 +178,11 @@ export class NamespacedAttribute implements ElementPatchOperation {
   }
 
   apply(dom: DOMHelper, element: Element, lastValue: string = null): any {
-    let { value: reference, namespace } = this;
+    let {
+      name,
+      value: reference,
+      namespace
+    } = this;
     let value = reference.value();
```
chancancode: :fearful:

packages/node_modules/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -845,103 +849,87 @@ test("error line numbers include multiple mustache lines", function() {
...
-//   let html = '<svg><path stroke="black" d="M 0 0 L 100 100"></path></svg>';
-//   let template = compile(html);
-//   let svgNode = render(template, {}).fragment.firstChild;
+module("Initial render (svg)");
 
-//   equal( svgNode.childNodes[0].namespaceURI, svgNamespace,
-//          "creates the path element with a namespace" );
-//   equalTokens(svgNode, html);
-// });
+test("Simple elements can have namespaced attributes", function() {
```
mixonic: Note to self: mmun sez these test names can be improved.

packages/node_modules/glimmer-runtime/lib/compiled/opcodes/dom.ts
chancancode: (Oops, this is one line off, I was again just surprised that the tests didn't catch the mistake!)

packages/node_modules/glimmer-runtime/tests/initial-render-test.ts
mmun: Specifically, what is a "simple element" vs an "SVG element"?

PR #64, opened 2016-02-06
#each built-in

chadhietala:


mixonic:
I presume there is discussion suggesting this change should happen. @chadhietala can you call out who should +1 this?

chadhietala:
I believe that the implementations from test-helpers were meant to be moved to the built in versions. @chancancode or @wycats can confirm or deny.

packages/node_modules/glimmer-runtime/lib/syntax/builtins/each.ts
```diff
@@ -0,0 +1,83 @@
...
+    compiler.append(new PushChildScopeOpcode());
+    compiler.append(new EvaluateOpcode({ debug: "default", block: this.templates.default }));
+    compiler.append(new PopScopeOpcode());
+    compiler.append(END);
+    compiler.append(new ExitOpcode());
+    compiler.append(new JumpOpcode({ target: ITER }));
+    compiler.append(BREAK);
+    compiler.append(new ExitListOpcode());
+  }
+}
```
chancancode: Newline?

chancancode:
@mixonic yeah, we don't feel comfortable exposing the opcodes as an API for Ember yet, so we decided to do it here (for now?). Seems okay to have these as built-in syntax

packages/node_modules/glimmer-runtime/lib/syntax/builtins/each.ts
chadhietala: Hmm not sure if github bug or not but I just added the new line and the diff did not update.

chancancode:
By the way, I don't think the iterators infrastructure is enough to support ember stuff in general, but should work fine for simple arrays

chadhietala:
@chancancode Yes, there are weird old deprecated behavior that I'm not sure how they should be handled. Basically old deprecated stuff like `itemViewClass`, `emptyViewClass` and a bunch of other stuff that should be go away. Not sure if the LTS stuff plays into supporting that stuff or not.

PR #65, opened 2016-02-06
Fix conways demo for the reference change

krisselden:
Update DOM version to not create a closure per rerender
and general cleanup for a more aspirational baseline by
using props versus attributes.

### DIFF
--- a/demos/conways.html
+++ b/demos/conways.html
@@ -8,11 +8,17 @@
         font-family: sans-serif;
       }
 
+      .world {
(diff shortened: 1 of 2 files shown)


chancancode:
> using props versus attributes

I personally would like to keep the DOM version as close to the glimmer version as possible, so that I can tell us how much overhead we are adding and how much is fixed DOM cost we cannot remove by micro-tuning the implementation. Maybe we can have both versions?

PR #66, opened 2016-02-06
more demo fixes

krisselden:
the ripples one doesn't error anymore but just the UpdatableReference thing did not seem to be enough.


chancancode:
Ripples is broken in chrome beta/canary due to https://github.com/whatwg/dom/issues/23

PR #67, opened 2016-02-06
Basic mem alloc optimizations

chadhietala:


mixonic:
:+1:

packages/node_modules/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -87,27 +91,29 @@ export class CloseElementOpcode extends Opcode {
...
+    }
 
     if (!classes.isEmpty()) {
       vm.updateWith(new NonNamespacedAttribute('class' as InternedString, classes).flush(dom, element));
     }
 
-    Object.keys(flattened).forEach(key => {
-      vm.updateWith(flattened[key].flush(dom, element));
-    });
+    for (let i = 0; i < flattenedKeys.length; i++) {
```
krisselden: if you let i the same name the de-conflicted name will not be correct in the source map names, so this hurts debugging, also, it can cause crazy code output from babel.

chadhietala: I'm assuming in the same outer scope?

stefanpenner: > it can cause crazy code output from babel.

this case should be fine, but ya the sourceMap lacking symbols blows...

chadhietala: So `k` :trollface: ?

stefanpenner: hm?

stefanpenner: their shouldn't be crazy output from this case, but the fact that babel doesn't produce symbols for the source map mapping does blow.

chancancode:
I think @wycats has thoughts on these kind of optimizations

PR #69, opened 2016-02-07
SVG + expression tests. Unsafe expressions in SVG

mixonic:
Unsafe expressions were disallowed inside the SVG context. This permits them to occur.


mixonic:
Please test on IE9, sez the powers that be.

mixonic:
This does not pass in IE9:



and it also fails in MS Edge :-( and all between.

mixonic:
Fighting sauce regarding a spurious Firefox failure on Sauce, but this is good to go (IE9 through Edge).

I can rewrite the tests in the style of the update tests in https://github.com/tildeio/glimmer/pull/70, but would prefer to see those approved and merged before going too far down that path.

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -721,8 +721,6 @@ test("error line numbers include multiple mustache lines", function() {
   }, /Closing tag `div` \(on line 3\) did not match last open tag `p` \(on line 2\)\./);
 });
 
-if (document.createElement('div').namespaceURI) {
```
mixonic: This was for IE8, not needed.

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -732,23 +730,6 @@ test("Namespaced attribute", function() {
   equal(svg.attributes[0].namespaceURI, XLINK_NAMESPACE);
 });
 
-test("Namespaced attribute with a quoted expression", function() {
```
mixonic: Moved to `updating-test.ts` as INUR

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -732,23 +730,6 @@ test("Namespaced attribute", function() {
...
-    "<svg xlink:title='{{title}}'>content</svg>",
-    "<svg xlink:title='svg-title'>content</svg>",
-    {title: 'svg-title'}
-  );
-  let svg = root.firstChild;
-  equal(svg.namespaceURI, SVG_NAMESPACE);
-  equal(svg.attributes[0].namespaceURI, XLINK_NAMESPACE);
-});
-
-test("Namespaced attribute with unquoted expression throws", function() {
```
mixonic: remains in this file since it it a compile-time assertion, just moved.

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
   compilesTo('<svg><linearGradient id="gradient"></linearGradient></svg>');
 });
 
-// QUnit.skip("svg can live with hydration", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `<svg> tag and expression as sibling`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-//   let template = compile('<svg></svg>{{name}}');
-
-//   let fragment = render(template, { name: 'Milly' }, env, { contextualElement: document.body }).fragment;
-
-//   equal(
-//     fragment.childNodes[0].namespaceURI, svgNamespace,
-//     "svg namespace inside a block is present" );
-// });
-
-// QUnit.skip("top-level unsafe morph uses the correct namespace", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `<svg> tag and unsafe expression as sibling`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-//   let template = compile('<svg></svg>{{{foo}}}');
-//   let fragment = render(template, { foo: '<span>FOO</span>' }).fragment;
-
-//   equal(getTextContent(fragment), 'FOO', 'element from unsafe morph is displayed');
-//   equal(fragment.childNodes[1].namespaceURI, xhtmlNamespace, 'element from unsafe morph has correct namespace');
-// });
-
-// QUnit.skip("nested unsafe morph uses the correct namespace", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `unsafe expression nested inside a namespace`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-//   let template = compile('<svg>{{{foo}}}</svg><div></div>');
-//   let fragment = render(template, { foo: '<path></path>' }).fragment;
-
-//   equal(fragment.childNodes[0].childNodes[0].namespaceURI, svgNamespace,
-//         'element from unsafe morph has correct namespace');
-// });
-
-// QUnit.skip("svg can take some hydration", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `expression nested inside a namespace`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-
-//   let fragment = render(template, { name: 'Milly' }).fragment;
-//   equal(
-//     fragment.firstChild.childNodes[0].namespaceURI, svgNamespace,
-//     "svg namespace inside a block is present" );
-//   equalTokens( fragment.firstChild, '<div><svg>Milly</svg></div>',
-//              "html is valid" );
-// });
-
-// QUnit.skip("root svg can take some hydration", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `expression nested inside a namespaced root element`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-//   let svgNode = fragment.firstChild;
-
-//   equal(
-//     svgNode.namespaceURI, svgNamespace,
-//     "svg namespace inside a block is present" );
-//   equalTokens( svgNode, '<svg>Milly</svg>',
-//              "html is valid" );
-// });
-
-// QUnit.skip("Block helper allows interior namespace", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `HTML namespace is created in child templates`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-//   fragment = render(template, { isTrue: false }, env, { contextualElement: document.body }).fragment;
-//   equal(
-//     firstChild(fragment).namespaceURI, xhtmlNamespace,
-//     "inverse block path has a normal namespace");
-//   equal(
-//     firstChild(fragment).firstChild.namespaceURI, svgNamespace,
-//     "svg namespace inside an element inside a block is present" );
-// });
-
-// QUnit.skip("Block helper allows namespace to bleed through", function() {
```
mixonic: Moved to `updating-test.ts` as INUR, renamed `HTML namespace is continued to child templates`

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -806,105 +785,8 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-
-//   let fragment = render(template, { isTrue: true }, env).fragment;
-//   let svgNode = fragment.firstChild.firstChild;
-//   equal( svgNode.namespaceURI, svgNamespace,
-//          "svg tag has an svg namespace" );
-//   equal( svgNode.childNodes[0].namespaceURI, svgNamespace,
-//          "circle tag inside block inside svg has an svg namespace" );
-// });
-
-// QUnit.skip("Block helper with root svg allows namespace to bleed through", function() {
```
mixonic: Was previously ported to `updating-test.ts` as `HTML namespace from root element is continued to child templates`

packages/glimmer-runtime/lib/dom.ts
```diff
@@ -82,7 +83,7 @@ class DOMHelper {
   }
 
   insertHTMLBefore(parent: HTMLElement, nextSibling: Node, html: string): Bounds {
-    let prev = nextSibling && nextSibling.previousSibling;
```
mixonic: When inserting into an element that already has children with `nextSibling=null`, the current `lastChild` of the parent should be considered the `prev` element. This fixed issues raise in tests below for `<svg></svg>{{{name}}}` on rerender. The bounds incorrectly included the `<svg>` tag (because `let first = prev ? prev.nextSibling : parent.firstChild;` at the end of this function), and thus on re-render the `<svg>` tag was being destroyed.

mixonic:
@chancancode @wycats this is really g2g now, I believe. :tada:

mixonic:
Grr, rewritten tests all pass in IE9 but IE9 wants to self-close SVG in a token comparison :sob: will fix.

chancancode:
Awesome work! :tada: :clap: :clap: :clap:

packages/glimmer-runtime/lib/dom.ts
chancancode: @mixonic Great catch. Can you add a non-svg test for this case? I think something like `<div>foo{{{bar}}}</div>` would reproduce the problem right? (I think the second update will wipe out the "foo" text node.) As you can see, the SVG tests were disabled for quite some time during the refactor. It would be good to notice something like this much earlier in future refactor cycles

mixonic: In https://github.com/tildeio/glimmer/pull/103

PR #70, opened 2016-02-07
foreignObject support

mixonic:
TODO:

* This implements foreignObject at dom render time instead of at least partially at parsing time (for example I feel at least flagging what elements are integration points and open SVG namespaces should happen at compile time, no?)
* Consequently there are parser tests for [integration points](https://github.com/tildeio/glimmer/blob/master/packages/node_modules/glimmer-syntax/tests/parser-node-test.ts#L76) and [svg](https://github.com/tildeio/glimmer/blob/master/packages/node_modules/glimmer-syntax/tests/parser-node-test.ts#L51) that are pretty poor, because nothing special is being tested (namespaces are not currently represented on the AST).

Needs a quick discussion.

### DIFF
--- a/packages/node_modules/glimmer-runtime/tests/initial-render-test.ts
+++ b/packages/node_modules/glimmer-runtime/tests/initial-render-test.ts
@@ -906,18 +906,6 @@ test("<foreignObject> tag has an SVG namespace", function() {
         "creates the foreignObject element with a namespace");
 });
 
-test("elements nested inside <foreignObject> have an XHTML namespace", function() {
-  compilesTo('<svg><foreignObject><div></div></foreignObject></svg>');
-  let svg = root.firstChild;
-  let foreignObject = svg.firstChild;
-  let div = foreignObject.firstChild;
-  equal(svg.namespaceURI, SVG_NAMESPACE);
-  equal(foreignObject.namespaceURI, SVG_NAMESPACE,
-        "creates the foreignObject element with a namespace");
-  equal(div.namespaceURI, XHTML_NAMESPACE,
-        "creates the div inside the foreignObject without a namespace");
-});
-
 test("Namespaced and non-namespaced elements as siblings", function() {
   compilesTo('<svg></svg><svg></svg><div></div>');
   let [first, second, third] = root.childNodes;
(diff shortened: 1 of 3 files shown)


stefanpenner:
Food for thought (reading this PR made me think of it), should we consider priming (by having the code when loaded take those paths) less commonly used paths like the SVG ones?

cc @krisselden

mixonic:
@chancancode @wycats we should confer. @mmun and I think this PR is a reasonable approach.

packages/node_modules/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -1047,17 +1047,26 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
+    '<svg>{{#unless}}<circle />{{/unless}}</svg>',
+    "<svg><circle /></svg>"
+  );
+  let svg = root.firstChild;
+  let circle = svg.firstChild;
+  equal(svg.namespaceURI, SVG_NAMESPACE);
+  equal(circle.namespaceURI, SVG_NAMESPACE);
+});
 
+test("root <foreignObject> tag is SVG namespaced" function() {
```
wycats: Can we add these tests in the updating test suite instead of adding it here? Going forward, we'd like all new tests to test both initial render and updating :smile:

mixonic:
@wycats should be good to go, added a new module to `updating-test` that implements these tests with updates (and they are accordingly removed from `initial-render-test`)

packages/node_modules/glimmer-runtime/tests/updating-test.ts
```diff
@@ -984,3 +987,87 @@ QUnit.skip("Pruned lists invoke a cleanup hook on their subtrees when removing e
...
+
+  object.hasCircle = true;
+  rerender();
+
+  equalTokens(root, "<svg><circle /></svg>");
+  equal(getSvg().namespaceURI, SVG_NAMESPACE);
+  equal(getCircle().namespaceURI, SVG_NAMESPACE);
+});
+
+test("root <foreignObject> tag is SVG namespaced" function() {
```
chancancode: How come this wasn't a syntax error on CI? o_0

PR #71, opened 2016-02-07
Disable failing tests in PhantomJS 1.9

mixonic:
Is there a unified story around browser hacks like this?

Should fix the CI build for Ember.js which runs Phantom 1.9 /cc @rwjblue 

(branch is named IE9, which is just incorrect)


packages/node_modules/glimmer-runtime/tests/updating-test.ts
```diff
@@ -2,6 +2,18 @@ import { Template, RenderResult } from "glimmer-runtime";
 import { TestEnvironment, equalTokens, stripTight } from "glimmer-test-helpers";
 import { UpdatableReference } from "glimmer-reference";
 
+/*
+ * IE9 does not serialize namespaced attributes correctly. The namespace
```
chancancode: Phantom 1.9?

rwjblue: https://github.com/tildeio/glimmer/pull/72

PR #72, opened 2016-02-07
Update comment for namespaced attribute detection.

rwjblue:


chadhietala:
Would it be worth just dump phantom using xvfb to launch chrome?

rwjblue:
@chadhietala - Phantom 1.9 and 2.0 are Ember supported platforms so we still need to test against them.  I do agree that we should test via chrome in Travis by default though (I am planning to do this in Ember too).

mixonic:
ugh thanks @rwjblue. I think I was having flashbacks to dec 2014 doing this SVG stuff.

PR #75, opened 2016-02-12
Emit ES6 & ES5 output also.

rwjblue:
Adds:

* dist/es6/
* dist/named-amd/

Ember will be updated to use named-amd (so that it can be more choosey with the files that it includes, to prevent duplication in the build output).


stefanpenner:
:+1:

rwjblue:
Thanks @stefanpenner

PR #82, opened 2016-02-20
Fix Glimmer Visualizer

chancancode:
This is to account for some of the changes we made to how components are compiled. This might not be a very good long term solution, but would allow the visualizer to work again for now.


mmun:
:tada:

PR #83, opened 2016-02-23
Fix to replay for uptime demo

tricknotes:
In [uptime boxes demo](https://github.com/tildeio/glimmer/blob/master/demos/uptime-boxes.html), the "Play" -> "Pause" switching works, but "Play" -> "Pause" -> "Play" keeps as paused against our expectation.
So I fixed to work toggling "Play" <-> "Pause".


mmun:
:bow:

PR #91, opened 2016-03-08
Fix loc info for content statements with leading newlines.

rwjblue:
Given the following:

```hbs
<div>
  {{#if isSingleError}}
    Single error here!
  {{else if errors}}
    Multiple errors here!
  {{else}}
    No errors found!
  {{/if}}
</div>
```

The `ElementNode` for the `<div>` was indicating that its `loc.end.column` was 12. This is ultimately due to how Handlebars handles trailing newlines after the `{{/if}}`. We were properly detecting that the node was `rightStripped` and adding the number of stripped newlines to the `tokenizer.line` number, but we did not reset the `tokenizer.column` also.

The fix is to ensure that the column is reset when incrementing the line count.

---

This is also being fixed in HTMLBars in https://github.com/tildeio/htmlbars/pull/450

### DIFF
--- a/packages/node_modules/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
+++ b/packages/node_modules/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
@@ -107,7 +107,9 @@ export default {
       changeLines = leadingNewlineDifference(content.original, content.value);
     }
 
-    this.tokenizer.line = this.tokenizer.line + changeLines;
+    this.tokenizer.line = content.loc.start.line + changeLines;
+    this.tokenizer.column = content.loc.start.column;
+
     this.tokenizer.tokenizePart(content.value);
     this.tokenizer.flushData();
   },
--- a/packages/node_modules/glimmer-syntax/tests/loc-node-test.ts
+++ b/packages/node_modules/glimmer-syntax/tests/loc-node-test.ts
@@ -85,3 +85,54 @@ test("html elements", function() {
   locEqual(div, 4, 6, 6, 12, 'div element');
   locEqual(hr, 5, 8, 5, 14, 'hr element');
 });
+
+test("html elements with nested blocks", function() {
+  var ast = parse(`
+    <div>
+      {{#if isSingleError}}
+        Single error here!
+      {{else if errors}}
+        Multiple errors here!
+      {{else}}
+        No errors found!
+      {{/if}} <p>Hi there!</p>
+    </div>
+  `);
+
+  let [,div] = ast.body;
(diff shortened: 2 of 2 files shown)


stefanpenner:
LGTM

packages/node_modules/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -107,7 +107,11 @@ export default {
       changeLines = leadingNewlineDifference(content.original, content.value);
     }
 
-    this.tokenizer.line = this.tokenizer.line + changeLines;
+    if (changeLines) {
```
wycats: Can this be moved into the previous conditional? It seems like `changeLines` is only set to a truthy value in there so it looks like the `content.rightStripped` check is enough to know whether to apply this change and reset `.column` to 0.

It weirdly appears that changeLines was created before this commit but not used :eyes:

wycats: Other than that, seems good. Feel free to merge.

wycats: Actually on further reflection I see that the variable was used in the deleted line. TLDR move the line and the column reset into the original conditional.

rwjblue: AFAIK, `leadingNewlineDifference` is not guaranteed to be > 0 (unless you know for certain that `content.rightStripped` guarantees that a newline was stripped?).  That is why I added a second conditional...

rwjblue: Based on https://github.com/wycats/handlebars.js/blob/9365b8290070f34bf797c836aed4335ce6a4094f/lib/handlebars/compiler/whitespace-control.js#L178-L194, it looks like `rightStripped` only indicates that some whitespace was stripped, and does not guarantee that we have `changeLines` of more than 1.

rwjblue:
OK, after digging in more I realized that basically any item that is after a mustache/block has incorrect location.  Instead of trying to manually track the location of the HTML parts vs the Handlebars parts, I decided to utilize the `ContentStatement`'s location information from the Handlebars AST to reset the tokenizers cursor position.

This change fixes the following additional issues that I identified while digging into the `{{else if foo}}` issue:

* `{{#foo-bar}}<div>Foo</div>{{/foo-bar}}`  -- elements inside blocks had incorrect start/end
* `{{foo-bar}} <p>Hi!</p>` -- elements after a mustache had incorrect start/end

PR #92, opened 2016-03-10
Ensure column is reset when incrementing right stripped lines.

rwjblue:
https://github.com/tildeio/glimmer/pull/91 initially intended to fix this, but while adding more examples to the test scenarios I zeroed in on another bug (that was ultimately wider spread than this one).

When a closing block tag is followed by a newline, that newline is stripped. We were properly handling the line being stripped, but not resetting the column when we incremented the line.

This ensures that the column is reset to 0 if the lines are incremented.


rwjblue:
Had to rebase for the great node_modules -> packages migration.  Should be good to go when green...

PR #96, opened 2016-03-13
Fix some of the demos

chadhietala:


stefanpenner:
nice, i wonder if we should add the demo's to the test suite?

chancancode:
> nice, i wonder if we should add the demo's to the test suite

+10. We always wanted to make `glimmer-demos` a regular package so it would get all the usual treatment, and add some smoke tests for each

PR #99, opened 2016-03-20
Normalize text values before inserting into DOM

wycats:


rwjblue:
Is the same for attributes already supported?

chancancode:
Good call :ok_hand: we should matrix test the basic curlies in different positions in ember :smile:

Issue #101, opened 2016-03-20
Fix {{{foo}}} where foo is an empty string


chancancode:
https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/dom.ts#L89

If bar === “” on initial render (or later become empty), it seems like we won't be able to update it after. Should probably put an empty comment in there.

PR #103, opened 2016-03-20
Unsafe exprs w/ siblings in IE, many unsafe nodes

mixonic:
Addresses: https://github.com/tildeio/glimmer/pull/69#discussion_r56769464

### DIFF
--- a/packages/glimmer-runtime/tests/updating-test.ts
+++ b/packages/glimmer-runtime/tests/updating-test.ts
@@ -13,8 +13,8 @@ const XHTML_NAMESPACE = 'http://www.w3.org/1999/xhtml';
  * prefix is incorrectly stripped off.
  */
 const serializesNSAttributesCorrectly = (function() {
-  let div = document.createElement('div');
-  let span = document.createElement('span');
+  let div = <HTMLElement> document.createElement('div');
+  let span = <HTMLElement> document.createElement('span');
   span.setAttributeNS('http://www.w3.org/XML/1998/namespace', 'xml:lang', 'en-uk');
   div.appendChild(span);
   return div.innerHTML === '<span xml:lang="en-uk"></span>';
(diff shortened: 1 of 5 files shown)


packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -134,6 +134,31 @@ test("updating a single trusting curly", () => {
...
+  rerender();
+
+  equal(getDiv().firstChild.textContent, 'hello ');
+  equalTokens(getDiv().childNodes[1].textContent, 'another ');
+  equal(getDiv().lastChild.textContent, 'world');
+
+  rerender({value});
+
+  equalTokens(root, '<div>hello <b>brave new </b>world</div>', 'rerender');
+});
```
chancancode: This is a good test (and we should keep it). However, I don't think it would have caught the specific bug you fixed: the bug requires `nextSibling` to be `null`, which `{{{value}}}` has to be the last child of the parent, something like `<div>hello {{{value}}}</div>`.

Maybe we should just test all three sibling cases (via duplication is fine, but if you want to abstract it :+1:):

* `<div>{{{value}}} world</div>` (tests `prev = null`)
* `<div>hello {{{value}}} world</div>` (the one you have here)
* `<div>hello {{{value}}}</div>` (tests `nextSibling = null`)

mixonic:
@chancancode @wycats this seems to catch a real issue in IE10 and IE11. The inserted HTML text node joins with the previous one after insertion, mucking up the node tracking.

packages/glimmer-runtime/tests/updating-test.ts
mixonic: It does catch the error in `insertAdjacentHTML`. On initial render there is no `nextSibling`, so the bounds returned are incorrect when stored. That there is a node after didn't matter for the bug I fixed in #69. I'll add a second test anyway.

chancancode:
@mixonic we might need to do another DOM patch where we detect rendering multiple adjacent `TextNode`s using `insertAdjacentHTML` causes them to be merged? In that case we might have to always put an empty comment before/after non-empty `{{{ }}}`? (We need to do a single empty comment for *empty* `{{{ }}}` in all browsers, but that's a separate issue in #101)

packages/glimmer-runtime/tests/updating-test.ts
chancancode: I see :fearful:

mixonic:
yup, that solution seems correct and not even too painful. I can take a stab in a bit.

mixonic:
Now fixes two bugs:

* IE, Edge, and Firefox do not return two text nodes when you `insertAdjacentHTML` with a sibling text node. Add a fix file for these. This fix touches many browsers, so it may be worth coming up with alternatives or filing bugs upstream to see if this is really thought of as a bug and will eventually be fixed (in Firefox at least). An alternative fix might include adding `<!---->` comments at the start and end of every insertion. This is for both safe and unsafe bindings.
* When there were many nodes being appended into a binding via the common wrapper-then-move pattern, the `nextSibling` was read *after* the node had been moved to a target binding. This meant subsequent siblings were abandoned on the throwaway wrapper node. This consolidates the moving login into a helper and adds tests that hit SVG fix and merging text node fix.

mixonic:
@chancancode this should now be implemented with a comment spacer instead of with a temporary wrapper and node moving. 👍

PR #105, opened 2016-03-20
Apply content normalization rules for concat

chancancode:
Addresses @rwjblue's comment: https://github.com/tildeio/glimmer/compare/normalize-concat?expand=1


rwjblue:
Thank you

PR #111, opened 2016-03-23
Move demos to packages, fix errors and lint.

krisselden:
Move demos code into packages so it stops breaking.

Ripples doesn't work because of each cleanup, Visualizer required hacking around private access to work.


chancancode:
:metal:

PR #112, opened 2016-03-25
Implement has-block helper

mmun:
Implements a has-block helper and tests it in subexpression, append and dynamic property position. I've always rewritten the tests to use `has-block` instead of `hasBlock`. We can deal with the camel-cased version in Ember's own tests.

A couple other things:
- I plan to refactor the template-compiler a bit so we don't have the `if (isHasBlock(action)) {` checks in 3 places.
- Attributes/concat don't work correctly because of a separate bug related to https://github.com/emberjs/ember.js/issues/12506

Actually both of these issues will likely be resolved together. The issue is that `foo="{{bar}}"` current assumes that this is a `get` for `bar`, but in fact it could also be a helper call for a helper named `bar`.

### DIFF
--- a/packages/glimmer-runtime/lib/syntax/core.ts
+++ b/packages/glimmer-runtime/lib/syntax/core.ts
@@ -48,6 +48,8 @@ import {
   CompiledSelfRef
 } from '../compiled/expressions/ref';
 
+import CompiledHasBlock from '../compiled/expressions/has-block';
+
 import CompiledHelper from '../compiled/expressions/helper';
 
 import CompiledConcat from '../compiled/expressions/concat';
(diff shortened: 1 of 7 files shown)


packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -924,6 +926,38 @@ export class Helper extends ExpressionSyntax<Opaque> {
...
+  name: InternedString;
+
+  constructor({ name }: { name: InternedString }) {
+    super();
+    this.name = name;
+  }
+
+  prettyPrint() {
+    return new PrettyPrint('expr', 'has-block', [this.name]);
+  }
```
wycats: I don't think we need `prettyPrint` anymore (but maybe it's needed to avoid a dangling type error). We do, however, need a `toJSON`.

packages/glimmer-runtime/tests/ember-component-test.ts
```diff
@@ -578,37 +547,78 @@ testComponent('parameterized hasBlock (inverse) when inverse supplied', {
...
+
+testComponent('parameterized has-block (content, default) when block not supplied', {
+  kind: 'curly',
+  layout: '{{has-block}}',
+  expected: 'false'
+});
+
+testComponent('parameterized has-block (attr, inverse) when inverse supplied', {
+  kind: 'curly',
+  layout: '<button class={{has-block "inverse"}}></button>',
```
wycats: Can you use some other test than `class={{unquoted}}`? The class prop/attr is the most entangled in likely changes.

Maybe better would be: `<button data-has-block="{{has-block "inverse"}}"></button>

mmun: It doesn't work at this this time. See https://github.com/tildeio/glimmer/pull/112#issue-143411395.

packages/glimmer-runtime/lib/syntax/core.ts
mmun: :+1: I will add CompiledHasBlock#toJSON.

packages/glimmer-runtime/tests/ember-component-test.ts
mmun: I'll switch the test off class to data- though. Good idea.

PR #114, opened 2016-03-28
Implement has-block-params helper

mmun:
Implements the has-block-helper with a bunch of tests. Many are skipped. I have a separate PR coming up that fixes the semantics of mustaches in attributes (and hence all of skipped tests in has-block/has-block-params) and hopefully can reduce some of code duplication that is present between the mustache/attributeMustache/SubExpression handlers in the template-compiler.


mmun:
ping @wycats

PR #115, opened 2016-03-28
Add missing dependencies to package.json

Dhaulagiri:
Add specific references to these packages in package.json so that `npm install` works in npm 2:
* broccoli-stew
* exists-sync


rwjblue:
THANK YOU!!

lisafeather:
Woohoo!

PR #124, opened 2016-04-03
Use Array.prototype.every in CompositeTag example

jgwhite:
The current example uses a short-circuiting return within an each which would do the right thing in Ruby but not, I think, in JavaScript.


chancancode:
Nice! I caught the bug the other day and was planning to replace it with a loop. But this is even better!

PR #127, opened 2016-04-05
BUGFIX Allow for slashed components

chadhietala:
This allows components to be declared as `{{foo-bar/biz-bar}}` in Glimmer. It also adds guard rails for the Glimmer subset of HandleBars.

### DIFF
--- a/packages/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
+++ b/packages/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
@@ -128,6 +128,26 @@ export default {
   },
 
   PathExpression: function(path) {
+    let { original, loc } = path;
+
+    if (original[0] === '@') {
+      original = original.slice(1);
+    }
+
+    /**
+     * This detects if a [HandleBars context switch](http://handlebarsjs.com/#paths) has been declared.
+     * In Glimmer context switching in general is prohibited.
+     */
+    if (path.depth > 0) {
+      throw new Error(`Handlebars context switching (using "../" in paths) is not supported by Glimmer. You should remove the context switching of "${path.original}" on line ${loc.start.line}.`);
+    }
+
+    if (original.slice(0, 2) === './') {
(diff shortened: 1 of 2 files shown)


chadhietala:
Results in this error:

```
Error: Compile Error: fizz-bar.baz-bar is not a helper
    at Helper.compile (http://localhost:7357/amd/glimmer-runtime.amd.js:6274:23)
    at Append.compile (http://localhost:7357/amd/glimmer-runtime.amd.js:5450:109)
    at EntryPointCompiler.compileStatement (http://localhost:7357/amd/glimmer-runtime.amd.js:3863:43)
    at EntryPointCompiler.compile (http://localhost:7357/amd/glimmer-runtime.amd.js:3892:22)
    at EntryPoint.compile (http://localhost:7357/amd/glimmer-runtime.amd.js:867:84)
    at Template.render (http://localhost:7357/amd/glimmer-runtime.amd.js:6754:37)
    at MyRootView.appendTo (http://localhost:7357/amd/glimmer-tests.amd.js:2654:42)
    at appendViewFor (http://localhost:7357/amd/glimmer-tests.amd.js:2712:14)
    at Object.<anonymous> (http://localhost:7357/amd/glimmer-tests.amd.js:3352:9)
    at runTest (http://localhost:7357/tests/qunit.js:903:28)
```

packages/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -128,6 +128,18 @@ export default {
...
 
   PathExpression: function(path) {
+    let { original } = path;
+
+    if (original[0] === '@') {
+      original = original.slice(1);
+    }
+
+    if (path.depth > 0) {
+      throw new Error(`Context switch detected. Glimmer does not support context switching, you should remove the context switching on ${path.original}`);
```
krisselden: I'm not sure about this error message, also, we should probably assert `{{./foo}}` which is essentially `{{this.foo}}` I'm not sure what we are doing with that.

PR #131, opened 2016-04-07
Fix regression with {{this.value}}

krisselden:
Adds test for {{this.value}} updating.

Moves tests compile errors into a different module and adds test for mixed slash and dot.


krisselden:
@wycats this ready for review

PR #133, opened 2016-04-15
Implement htmlSafe support

zackthehuman:
This PR implements support for `htmlSafe` during append. It treats any value that isn't const as *possibly* a SafeString, and then later checks if the value really is one.

It's possible that a value that previously was a SafeString is no longer one, and vice-versa. To account for this, it is treated as a range of nodes with a `Bounds`. In the case that it is not a SafeString and currently is text, it simple updates the `nodeValue` -- the same way as `UpdateAppendOpcode` does. Otherwise, it works basically like `UpdateTrustingAppendOpcode`.

I still need to do some work around detecting SafeString using its symbol. For now I think this is far enough along to get feedback.

### DIFF
--- a/packages/glimmer-runtime/lib/builder.ts
+++ b/packages/glimmer-runtime/lib/builder.ts
@@ -279,6 +279,14 @@ export class ElementStack {
     return fragmentBounds;
   }
 
+  insertTextBefore(nextSibling: Node, text: string): Fragment {
+    let concreteBounds = this.dom.insertTextBefore(this.element, this.nextSibling, text);
+    let fragmentBounds = new Fragment(concreteBounds);
+    this.blockStack.current.newBounds(fragmentBounds);
+
+    return fragmentBounds;
+  }
+
   // setAttribute(name: InternedString, value: any) {
   //   this.dom.setAttribute(this.element, name, value);
   // }
--- a/packages/glimmer-runtime/lib/compiled/opcodes/content.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/content.ts
@@ -2,7 +2,7 @@ import { Opcode, OpcodeJSON, UpdatingOpcode } from '../../opcodes';
 import { VM, UpdatingVM } from '../../vm';
 import { ReferenceCache, isModified, isConst, map } from 'glimmer-reference';
 import { Opaque, dict } from 'glimmer-util';
-import { clear } from '../../bounds';
+import { Bounds, clear } from '../../bounds';
 import { Fragment } from '../../builder';
 
 export function normalizeTextValue(value: Opaque): string {
@@ -11,6 +11,18 @@ export function normalizeTextValue(value: Opaque): string {
   } else {
     return String(value);
   }
+}
+
+export function normalizeTextOrTrustedValue(value: Opaque): Opaque {
+  if (isSafeString(value)) {
+    return value;
+  } else {
+    return normalizeTextValue(value);
+  }
+}
+
+export function isSafeString(value: Opaque): boolean {
+  return value && typeof value['toHTML'] === 'function';
 }
 
 abstract class UpdatingContentOpcode extends UpdatingOpcode {
@@ -27,12 +39,20 @@ export class AppendOpcode extends Opcode {
   evaluate(vm: VM) {
     let reference = vm.frame.getOperand();
 
-    let mapped = map(reference, normalizeTextValue);
+    let mapped = map(reference, normalizeTextOrTrustedValue);
     let cache = new ReferenceCache(mapped);
-    let node = vm.stack().appendText(cache.peek());
+    let value = cache.peek();
 
-    if (!isConst(reference)) {
-      vm.updateWith(new UpdateAppendOpcode(cache, node));
+    if (isSafeString(value)) {
+      let bounds = vm.stack().insertHTMLBefore(null, value);
+      vm.updateWith(new UpdateCautiousAppendOpcode(cache, bounds, null));
+    } else {
+      if (isConst(reference)) {
+        vm.stack().appendText(cache.peek());
+      } else {
+        let bounds = vm.stack().insertTextBefore(null, value);
+        vm.updateWith(new UpdateCautiousAppendOpcode(cache, bounds, bounds.firstNode() as Text));
+      }
     }
   }
 
@@ -42,33 +62,6 @@ export class AppendOpcode extends Opcode {
       type: this.type,
       args: ["$OPERAND"]
     };
-  }
-}
-
-export class UpdateAppendOpcode extends UpdatingContentOpcode {
-  type = 'update-append';
-  private cache: ReferenceCache<string>;
-  private textNode: Text;
-
-  constructor(cache: ReferenceCache<string>, textNode: Text) {
-    super();
-    this.cache = cache;
-    this.textNode = textNode;
-  }
-
-  evaluate() {
-    let value = this.cache.revalidate();
-    if (isModified(value)) this.textNode.nodeValue = value;
-  }
-
-  toJSON(): OpcodeJSON {
-    let { _guid: guid, type } = this;
-
-    let details = dict<string>();
-
-    details["lastValue"] = JSON.stringify(this.cache.peek());
-
-    return { guid, type, details };
   }
 }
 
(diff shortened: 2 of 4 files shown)


zackthehuman:
I realize I acidentally removed a branch from the original code, which is the cause of most of the errors.

zackthehuman:
Now with a test!

krisselden:
Looks good, nice work.

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -13,6 +13,21 @@ export function normalizeTextValue(value: Opaque): string {
...
+export function normalizeTextOrTrustedValue(value: Opaque): Opaque {
+  if (isSafeString(value)) {
+    return value;
+  } else {
+    return normalizeTextValue(value);
+  }
+}
+
+export function isSafeString(value: Opaque): boolean {
+  // FIXME: This feels not good enough to detect SafeStrings. What is the value
```
rwjblue: The implementation here (checking `toHTML`) is correct. We do not have the option to change what it means to be a `SafeString` at this point.  Changing would break a number of things elsewhere.

Can you remove this comment?

rwjblue: The implementation here (checking `toHTML`) is correct. We do not have the option to change what it means to be a `SafeString` at this point.  Changing would break a number of things elsewhere.

Can you remove the inline code comment here suggesting it is incorrect?

packages/glimmer-runtime/lib/builder.ts
```diff
@@ -279,6 +279,14 @@ export class ElementStack {
     return fragmentBounds;
   }
 
+  insertTextBefore(nextSibling: Node, text: string): Fragment {
```
chancancode: this can probably just return a `Text` and let upstream deal with the wrapping

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -13,6 +13,18 @@ export function normalizeTextValue(value: Opaque): string {
   }
 }
 
+export function normalizeTextOrTrustedValue(value: Opaque): Opaque {
```
chancancode: should return `string | SafeString`

chancancode: (or `Insertion`)

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -27,12 +39,20 @@ export class AppendOpcode extends Opcode {
...
+    let mapped = map(reference, normalizeTextOrTrustedValue);
     let cache = new ReferenceCache(mapped);
-    let node = vm.stack().appendText(cache.peek());
-
-    if (!isConst(reference)) {
-      vm.updateWith(new UpdateAppendOpcode(cache, node));
+    let value = cache.peek();
+
+    if (isSafeString(value)) {
+      let bounds = vm.stack().insertHTMLBefore(null, value);
```
chancancode: we should handle `isConst` for this branch also (which argues for better factoring of the code) – it could happen if you have a helper that produces a SafeString and we noticed that the inputs to the helper is static

packages/glimmer-runtime/lib/builder.ts
chancancode: nvm, I see why this is needed now. we need to insert a stable bounds (`Fragment`) here because the siblings etc might want to reference that

chancancode: which probably means we need a better name for than `insertTextBefore` at this level – the point is we are inserting some "dynamically bounded" content that happens to be text at the moment, but that might change later

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -45,33 +65,6 @@ export class AppendOpcode extends Opcode {
...
-
-    let details = dict<string>();
-
-    details["lastValue"] = JSON.stringify(this.cache.peek());
-
-    return { guid, type, details };
-  }
-}
-
 export class TrustingAppendOpcode extends Opcode {
```
chancancode: I think `TrustingAppendOpcode` and the corresponding updating opcode also needs to handle `SafeString`s. The difference is you will *always* insert things as HTML.. i.e. `insertHTMLBefore( isSafeString(value) ? valu.toHTML() : value )`

chancancode: Ultimately, that's what the `Insertion` type is for. We eventually need to handle `Node`s too, but that's a different project

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
zackthehuman: Is that something that `isConst` can determine? What if the helper is stateful and the return value changes the markup in the returned safeString? It seems like the code path would be the same.

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
zackthehuman: I will add tests cases for this path as well.

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
chancancode: isConst is the way we propagate const-ness – specifically each producer of a reference can decide to produce the a (subclas of) ConstRefernce to participate in the optimization. You are right that "class-based helpers" in Ember would not be able to participate in this optimization (i.e. we will not produce a const reference in that case).

chancancode: Ember doesn't actually implement that optimization for simple helpers yet, but it will be pretty simple to do that. Here are some examples of returning const references: https://github.com/emberjs/ember.js/pull/13332/files

* In `{{component "foo" ...}}` we can notice that foo is const (`isConst(args.positional.at(0)`), therefore this component helper cannot change to a different component type, which allows us to optimize out some updating stuff
* In `{{if @cond "foo" bar}}`, this has to be dynamic, because `@cond` and `bar` could change. However, if we are able to know that `@cond` is hard-coded to `true` from the invocation template, then we know we will not ever go into the `bar` branch, so the whole thing is actually `isConst` (always the static string "foo")

packages/glimmer-runtime/lib/dom.ts
```diff
@@ -115,6 +115,12 @@ class DOMHelper {
     return new ConcreteBounds(parent, first, last);
   }
 
+  insertTextBefore(parent: Element, nextSibling: Node, text: string): Bounds {
```
chancancode: this should return `Text` and let upstream handle the wrapping

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -127,3 +120,50 @@ export class UpdateTrustingAppendOpcode extends UpdatingContentOpcode {
...
+        this.textNode = null;
+        let nextSibling = clear(this.bounds);
+        this.bounds.update(vm.dom.insertHTMLBefore(parent, nextSibling, value));
+      } else {
+        if(this.textNode) {
+          this.textNode.nodeValue = value;
+        } else {
+          let nextSibling = clear(this.bounds);
+          this.bounds.update(vm.dom.insertTextBefore(parent, nextSibling, value));
+          this.textNode = this.bounds.firstNode() as Text;
```
chancancode: `vm.dom.insertTextBefore` should return `Text`:

```
let textNode = this.textNode = vm.dom.insertTextBefore(parent, nextSibling, value);
this.bounds.update(new SingleNodeBounds(textNode));
```

packages/glimmer-runtime/lib/builder.ts
chancancode: `appendInsertion`?

chancancode: I'm a bit sad that we now need so much machinery around `{{foo}}` when 90%+ of the time it's going to be just simple strings, but I don't have any ideas to improve it at the moment :/

chancancode: (to be clear, this is great work – I don't mean I don't like the implementation, I just mean I'm sad that implementing this feature forces us to pollute the common path with a lot of stuff that handles the edge cases, and would like to eventually figure out how to not do that)

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
zackthehuman: I avoided doing this because nowhere else are bounds constructed inside of `content.ts` -- they seem to all be constructed at a different layer (inside of the `dom` object). I can move it, it just seemed inconsistent.

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
zackthehuman: I still think there is an issue with const detection in this case. Just because the input to a helper is const doesn't mean the ouput is -- even if it's a simple helper. The only way to guarantee that would be to assume that all simple helpers are pure and I don't think that's a safe assumption.

What if I implement a helper like:

```function random([min, max]) {
    return Math.floor(Math.random() * (max - min + 1) + min);
}```

```<input value={{random 1 10}}/>```

The inputs are const but the output isn't. Will there be a way for developers to specify their constness?

packages/glimmer-runtime/lib/builder.ts
zackthehuman: I'm not sure `appendInsertion` makes sense because it implies that we're inserting at the end always which is not the case.

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
chancancode: Simple helpers are required to be a pure function of its inputs - when do you expect that helper to recompute? (Class based helpers on the other hand are explicitly designed for that scenario)

zackthehuman: Good point. I'll see what I can do.

zackthehuman:
I addressed the comments from here and submitted a new PR: https://github.com/tildeio/glimmer/pull/139

PR #135, opened 2016-04-15
Add loc to text, comment, and attribute nodes.

rwjblue:
@fivetanley + @rwjblue - Pairing === Love.


rwjblue:
Since this is mostly additive, I'm going to land (and add to htmlbars).  I'm happy to iterate on anything that y'all find/want to change.

PR #136, opened 2016-04-16
WIP refineExpression

chadhietala:
This is incomplete but I wanted to PR it to get some initial feedback on it. I've noticed that there are cases where it appears that we go from wire format -> compiled expressions. Maybe thats ok, but we want to refine expressions in some cases downstream. For instance:

```handlebars
{{foo-bar class=someTruth}} {{!-- given someTruth is `true` --}}
```
results in

```html
<div class="some-truth ember-view"></div>
```

So the idea is that `refineExpression` would return more specific `ExpressionSyntax`. However this brings up some interesting questions like do we want to expose `CompiledExpression` to the environment. If you implement `ExpressionSyntax`, `compile` [normally returns](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/syntax/core.ts#L183-L191) `CompiledExpression`. Or is a subclass of `Ref`good enough?


packages/glimmer-runtime/lib/environment.ts
```diff
@@ -121,10 +122,19 @@ export abstract class Environment {
...
+  expression(key: string, expression: ExpressionSyntax<Opaque>): ExpressionSyntax<Opaque> {
+    return this.refineExpression(parseExpression(key, expression)) || expression;
+  }
+
   statement(statement: StatementSyntax): StatementSyntax {
     return this.refineStatement(parseStatement(statement)) || statement;
   }
 
+  protected refineExpression(expression: ParsedExpression): ExpressionSyntax<Opaque> {
+    console.log(expression);
```
chadhietala: You are passed something like 
screen shot 2016-04-16 at 10 15 32 am

PR #139, opened 2016-04-19
Support for htmlSafe/SafeString

### LINKED ISSUE #101: Fix {{{foo}}} where foo is an empty string

zackthehuman:
This creates a new type of Append opcode that handles the value when it is a SafeString and also when it is a normal string. For SafeString it enters into "html" mode, and if the input changes into a normal string it switches into "text" mode. It can switch freely between the two modes and handles updates accordingly.

- The opcode also supports const expressions.
- New tests have been added which test the switching behavior of the opcode.
- Checked the code TypeScript errors (via `tslint` and `vscode-build`) and corrected them.

This change appears to fix #101 so I added a test that checks that case specifically.

### DIFF
--- a/packages/glimmer-runtime/lib/environment.ts
+++ b/packages/glimmer-runtime/lib/environment.ts
@@ -200,21 +200,30 @@ export abstract class Environment {
 
 export default Environment;
 
-// TS does not allow us to use computed properties for this, so inlining for now
-// import { TRUSTED_STRING } from './symbols';
+export interface SafeString {
+  toHTML(): string;
+}
 
-interface SafeString {
-  "trusted string [id=7d10c13d-cdf5-45f4-8859-b09ce16517c2]": boolean; // true
-  string: string;
+export function isSafeString(value: Opaque): value is SafeString {
+  return value && typeof value['toHTML'] === 'function';
+}
+
+export function isNode(value: Opaque): value is Node {
+  return value !== null && typeof value === 'object' && typeof value['nodeType'] === 'number';
+}
+
+export function isString(value: Opaque): value is string {
+  return typeof value === 'string';
 }
 
 export type Insertion = string | SafeString | Node;
+export type TrustedInsertion = string | Node;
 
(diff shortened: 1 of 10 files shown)


krisselden:
rebase and squash

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -233,27 +233,6 @@ export class Append extends StatementSyntax {
   }
 }
 
-class HelperInvocationReference implements Reference<Insertion> {
```
chancancode: :+1:

packages/glimmer-runtime/lib/environment.ts
```diff
@@ -200,21 +200,30 @@ export abstract class Environment {
...
 }
 
 export type Insertion = string | SafeString | Node;
+export type TrustedInsertion = string | Node;
 
 type PositionalArguments = Opaque[];
 type KeywordArguments = Dict<Opaque>;
 
 export interface Helper {
-  (args: EvaluatedArgs): PathReference<Opaque>;
```
chancancode: is this change intentional?

zackthehuman: Yes, otherwise there doesn't seem to be a way to return a `ConstReference` from a `Helper` since they aren't `PathReference`s. Is there a better way to resolve?

chancancode: @zackthehuman @krisselden has filled me in on the details 👍 looking into it!

Is 4148235 the latest commit you have? I already rebased on top of it, so I think we are good to merge after resolving this point!

zackthehuman: Yes that is the most up-to-date commit.

PR #140, opened 2016-04-19
Ensure `undefined` doesn't get pushed into classes

chadhietala:


mmun:
Is there a test covering this? I'm surprised by this behaviour...

chancancode:
@mmun the test that @chadhietala modified has a concat in there. I **think** having a trailing space in that case is arguably the correct behavior since that's technically what you asked for. But maybe htmlbars does this normalization today?

The use case in ember is sometimes we have one of the class references return an empty string. But I just realized we already have a way to handle this (returning null) - so maybe we should just make sure the ember references return that. Or we can change the check to be `!value` which also allows `undefined`, which I think might be what ember is returning.

chadhietala:
So I believe this change makes it so HTMLBars and Glimmer behave the same in regards to whitespace and attribute removal.

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -547,6 +547,28 @@ test("class attribute w/ concat follow the normal dirtying rules", function() {
   equalTokens(root, "<div class='hello world'>hello</div>");
 });
 
+test("attribute is removed if the binding becomes falsy", function() {
```
rwjblue: I think this only applies to `class` right?  If that is true, can you update this to say "class attribute is removed if...."?

chadhietala: 👍 Also added an assertion to make sure there is coverage around non `class` attributes.

rwjblue:
🍧

PR #141, opened 2016-04-19
Support meta for templates

asakusuma:
Specifically support adding a moduleName to templates, which is turn passed back into [refineStatement](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/environment.js#L50). Something like this is needed to support local lookup.

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/blocks.ts
+++ b/packages/glimmer-runtime/lib/compiled/blocks.ts
@@ -1,4 +1,4 @@
-import { InternedString } from 'glimmer-util';
+import { InternedString, StatementMeta } from 'glimmer-util';
 import { OpSeq } from '../opcodes';
 import { Program } from '../syntax';
 import { Environment } from '../environment';
@@ -13,6 +13,7 @@ export interface BlockOptions {
   children: InlineBlock[];
   program: Program;
   symbolTable: SymbolTable;
+  meta: StatementMeta;
 }
 
 export class CompiledBlock {
@@ -29,12 +30,14 @@ export abstract class Block {
   public children: InlineBlock[];
   public program: Program;
   public symbolTable: SymbolTable;
+  public meta: StatementMeta = null;
   protected compiled: CompiledBlock = null;
 
   constructor(options: BlockOptions) {
     this.symbolTable = options.symbolTable || null;
     this.children = options.children;
     this.program = options.program;
+    this.meta = options.meta;
   }
 }
 
--- a/packages/glimmer-runtime/lib/scanner.ts
+++ b/packages/glimmer-runtime/lib/scanner.ts
@@ -2,7 +2,7 @@ import { Program, Statement as StatementSyntax } from './syntax';
 import buildStatement from './syntax/statements';
 import { TopLevelTemplate, EntryPoint, InlineBlock, PartialBlock, Layout } from './compiled/blocks';
 import Environment from './environment';
-import { EMPTY_SLICE, LinkedList, Stack } from 'glimmer-util';
+import { EMPTY_SLICE, LinkedList, Stack, StatementMeta } from 'glimmer-util';
 import { SerializedTemplate, SerializedBlock, Statement as SerializedStatement } from 'glimmer-wire-format';
 import SymbolTable from './symbol-table';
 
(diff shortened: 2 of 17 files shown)


krisselden:
@asakusuma it needs a test that the meta for the a template and a block is correct with a component layout inbetween

asakusuma:
@wycats has suggested that the `SymbolTable` be used. I tried passing the layout symbolTable into `refineStatement`, but `parent` and `named` are both empty at the time:
empty symboltable

asakusuma:
Ok I've got it working where the meta is added to the `SymbolTable`, which is in turn passed into `refineStatement`.

@krisselden what exactly do you mean by "with a component layout inbetween"?

krisselden:
if I have a template that is `<div>outside{{#my-component}}inside{{/my-component}}</div>` and my-component's template is `my-component{{yield}}`  then inside my component its module id should be my-component, but the outside and inside should be that template.

asakusuma:
Test added

asakusuma:
Build is now passing with the added test

packages/glimmer-runtime/lib/scanner.ts
```diff
@@ -88,7 +89,8 @@ export class BlockScanner {
   }
 
   endBlock(): InlineBlock {
-    let { children, program } = this.stack.pop();
+    let swag = this.stack.pop();
+    let { children, program } = swag;
```
chadhietala: Hmm not sure about `swag` 😄 . I think this can go back to what it was.

packages/glimmer-test-helpers/lib/helpers.ts
```diff
@@ -41,11 +42,13 @@ interface CompileOptions {
  * @return {Template} A function for rendering the template
  */
 export function compile(string: string, options: CompileOptions): Template {
+  //console.log(options)
```
chadhietala: Remove the logs

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -520,8 +525,12 @@ export class TestEnvironment extends Environment {
     return this.components[<string>name[0]];
   }
 
-  compile(template: string) {
-    return rawCompile(template, { env: this });
+  compile(template: string, options?: CompileOptions) {
+    let rawOptions = { env: this, moduleName: null };
+    if (options) {
```
chadhietala: Seems like there is a world where these guards are not needed?

packages/glimmer-runtime/lib/scanner.ts
asakusuma: Man I was really hoping to slip this by you guys

packages/glimmer-test-helpers/lib/environment.ts
asakusuma: Since this compile function is specific to the test environment, we need to do this guarding in order to provide a default moduleName. moduleName is not relevant to a lot of tests.

asakusuma: I have removed some optional meta properties and guards from the non-test code.

packages/glimmer-runtime/tests/ember-component-test.ts
```diff
@@ -2042,6 +2049,23 @@ QUnit.test('Curly component hooks (attrs as self props)', function() {
...
+  env.statement = function(statement: StatementSyntax, symbolTable: SymbolTable) {
+    if (statement instanceof Text) {
+      equal(statement.content, symbolTable.getModuleName(), 'moduleName is correct');
+    }
+    return original.call(env, statement, symbolTable);
+  }
+
+  env.registerEmberishGlimmerComponent('my-component', inspectHooks(MyComponent), 'my-component{{yield}}');
+
+  appendViewFor('<div>-top-level{{#my-component}}-top-level{{/my-component}}</div>');
```
krisselden: the point of asking you to do this was to check the getModuleName() symbol in the top level, layout and block context.

krisselden: nevermind, I just realized your test is doing this.

packages/glimmer-runtime/lib/compiled/blocks.ts
```diff
@@ -29,12 +30,14 @@ export abstract class Block {
...
   public program: Program;
   public symbolTable: SymbolTable;
+  public meta: StatementMeta = null;
   protected compiled: CompiledBlock = null;
 
   constructor(options: BlockOptions) {
     this.symbolTable = options.symbolTable || null;
     this.children = options.children;
     this.program = options.program;
+    this.meta = options.meta || {};
```
krisselden: Looks like meta is guarded in an if during setMeta and this is inconsistent with the other constructor that doesn't || {}

asakusuma: Good catch, fixed

asakusuma: Also, `BlockOptions.meta` is not optional, so all I had to do was just remove the ` || {}`

asakusuma:
Closed in favor of https://github.com/tildeio/glimmer/pull/201 and https://github.com/tildeio/glimmer/pull/205

PR #142, opened 2016-04-20
#139 + Upsert refactor

chancancode:
Merged in the latest (4148235) changes in #139 with the upsert refactor: 

Move all the upset code into objects in a new file: each object only need to know how to update within the same type (e.g. html -> html, Node -> Node). If they fail to update because the type has changed, the parent can simply clear the rendered content and make a new upsert object of the appropriate type.

Also fixed CompiledHelpers claiming to correctly return a `PathReference<Opaque>`.

/cc @zackthehuman @krisselden

### DIFF
--- a/packages/glimmer-runtime/lib/environment.ts
+++ b/packages/glimmer-runtime/lib/environment.ts
@@ -200,21 +200,30 @@ export abstract class Environment {
 
 export default Environment;
 
-// TS does not allow us to use computed properties for this, so inlining for now
-// import { TRUSTED_STRING } from './symbols';
+export interface SafeString {
+  toHTML(): string;
+}
 
-interface SafeString {
-  "trusted string [id=7d10c13d-cdf5-45f4-8859-b09ce16517c2]": boolean; // true
-  string: string;
+export function isSafeString(value: Opaque): value is SafeString {
+  return value && typeof value['toHTML'] === 'function';
+}
+
+export function isNode(value: Opaque): value is Node {
+  return value !== null && typeof value === 'object' && typeof value['nodeType'] === 'number';
+}
+
+export function isString(value: Opaque): value is string {
+  return typeof value === 'string';
 }
 
 export type Insertion = string | SafeString | Node;
+export type TrustedInsertion = string | Node;
 
 type PositionalArguments = Opaque[];
 type KeywordArguments = Dict<Opaque>;
 
 export interface Helper {
-  (args: EvaluatedArgs): PathReference<Opaque>;
+  (args: EvaluatedArgs): Reference<Opaque>;
 }
 
 export interface ParsedStatement {
--- a/packages/glimmer-runtime/lib/environment.ts
+++ b/packages/glimmer-runtime/lib/environment.ts
@@ -200,25 +200,6 @@ export abstract class Environment {
 
 export default Environment;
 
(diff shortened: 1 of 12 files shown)


packages/glimmer-runtime/lib/environment.ts
```diff
@@ -200,21 +200,31 @@ export abstract class Environment {
...
-export type Insertion = string | SafeString | Node;
+export type Insertion = CautiousInsertion | TrustingInsertion;
+export type CautiousInsertion = string | SafeString | Node;
+export type TrustingInsertion = string | Node;
 
 type PositionalArguments = Opaque[];
 type KeywordArguments = Dict<Opaque>;
 
 export interface Helper {
-  (args: EvaluatedArgs): PathReference<Opaque>;
```
chancancode: @zackthehuman is this change intentional? I believe we actually want helpers to return `PathReference`s (see https://github.com/emberjs/ember.js/pull/13173#discussion_r57362608)

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -394,7 +394,7 @@ export class SimplePathReference<T> implements PathReference<T> {
 
 type UserHelper = (args: any[], named: Dict<any>) => any;
 
-class HelperReference implements PathReference<Opaque> {
```
chancancode: See https://github.com/tildeio/glimmer/pull/142/files#r60489824

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -409,10 +409,6 @@ class HelperReference implements PathReference<Opaque> {
 
     return helper(positional.value(), named.value());
   }
-
-  get(prop: InternedString): SimplePathReference<Opaque> {
-    return new SimplePathReference(this, prop);
-  }
```
chancancode: See https://github.com/tildeio/glimmer/pull/142/files#r60489824

packages/glimmer-runtime/lib/environment.ts
zackthehuman: Yes, otherwise there doesn't seem to be a way to return a ConstReference from a Helper since they aren't PathReferences. Is there a better way to resolve?

zackthehuman: We wanted a helper that returned a `ConstReference` so we could write a test for const `SafeStrings` and `Node`s.

chancancode: Yep, @krisselden told me about it. There is a few subclasses of `ConstReference` that implements `PathReference`, I'll fix it!

chancancode: I was surprised that this change didn't cause any type errors, and I just figured out why. In https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/expressions/helper.ts#L22, we used `Function.prototype.call` which is typed to be `call(thisArg: any, ...args: any[]): any`, so the return value became `any` and the type system happily let it go through, even though the method is supposed to return a `PathReference<Opaque>` 😞

Fixing that as well

chancancode: @zackthehuman @krisselden fixed in 8001311 👍

PR #146, opened 2016-04-23
Basic Modifier support

zackthehuman:
This adds both a `Modifier` syntax and a `ModifierOpcode`. The opcode basically does nothing, but it's nice because it shows up in the visualizer.

This work is needed to start working on more ambitious things like `{{action}}`s and the like.

- [x] Figure out how this composes with Glimmer components.
- [x] Update `Environment` with `hasModifier` and `lookupModifier`.
- [x] Check if modifier exists when compiling `Modifier` syntax. Throw if there isn't one.
- [x] Create the `ModifierManager` interface.
- [x] Create `UpdateModifierOpcode` and emit it after installing inside of `ModifierOpcode`.
- [x] Handle created `Modifier`'s destructor.
- [x] Implement `UpdateModifierOpcode` to actually perform updating.
- [x] Write a test that tests the ins and outs of this feature.

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
@@ -6,13 +6,16 @@ import {
   Reference,
   ReferenceCache,
   RevisionTag,
+  Revision,
   combineTagged,
   isConst as isConstReference,
   isModified
 } from 'glimmer-reference';
+import { ModifierManager } from '../../modifier/interfaces';
 import { DOMHelper } from '../../dom';
 import { NULL_REFERENCE } from '../../references';
 import { ValueReference } from '../../compiled/expressions/value';
+import { CompiledArgs, EvaluatedArgs } from '../../compiled/expressions/args';
 
 abstract class DOMUpdatingOpcode extends UpdatingOpcode {
   public type: string;
@@ -218,6 +221,90 @@ export class StaticAttrOpcode extends Opcode {
     details["value"] = JSON.stringify(value.value());
 
     return { guid, type, details };
+  }
+}
+
+export class ModifierOpcode extends Opcode {
+  public type = "modifier";
+  public name: InternedString;
+  public args: CompiledArgs;
+  private manager: ModifierManager<Opaque>;
+
+  constructor({ name, manager, args }: { name: InternedString, manager: ModifierManager<Opaque>, args: CompiledArgs }) {
+    super();
+    this.name = name;
+    this.manager = manager;
+    this.args = args;
+  }
+
+  evaluate(vm: VM) {
+    let { manager } = this;
+    let stack = vm.stack();
+    let { element, dom } = stack;
+    let args = this.args.evaluate(vm);
+
+    let modifier = manager.install(element, args, dom);
+    let destructor = manager.getDestructor(modifier);
+
+    if (destructor) {
+      vm.newDestroyable(destructor);
+    }
+
+    vm.updateWith(new UpdateModifierOpcode({
+      manager,
+      modifier,
+      element,
+      args
+    }));
+  }
+
+  toJSON(): OpcodeJSON {
+    let { _guid: guid, type, name, args } = this;
+
+    let details = dict<string>();
+
+    details["type"] = JSON.stringify(type);
+    details["name"] = JSON.stringify(name);
+    details["args"] = JSON.stringify(args);
+
+    return { guid, type, details };
+  }
+}
+
+export class UpdateModifierOpcode extends DOMUpdatingOpcode {
+  public type = "update-modifier";
+
+  private element: Element;
+  public args: EvaluatedArgs;
+  private manager: ModifierManager<Opaque>;
+  private modifier: Opaque;
+  private tag: RevisionTag;
+  private lastUpdated: Revision;
+
+  constructor({ manager, modifier, element, args }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element, args: EvaluatedArgs }) {
+    super();
+    this.modifier = modifier;
+    this.manager = manager;
+    this.element = element;
+    this.args = args;
+    this.lastUpdated = args.tag.value();
+  }
+
+  evaluate(vm: UpdatingVM) {
+    let { manager, modifier, element, args, lastUpdated } = this;
+
+    if (!args.tag.validate(lastUpdated)) {
+      manager.update(modifier, element, args, vm.dom);
+      this.lastUpdated = args.tag.value();
+    }
+  }
+
+  toJSON(): OpcodeJSON {
+    return {
+      guid: this._guid,
+      type: this.type,
+      args: [JSON.stringify(this.args)]
+    };
   }
 }
 
--- a/packages/glimmer-runtime/lib/modifier/interfaces.ts
+++ b/packages/glimmer-runtime/lib/modifier/interfaces.ts
@@ -1,0 +1,19 @@
+import { EvaluatedArgs } from '../compiled/expressions/args';
+import { DOMHelper } from '../dom';
+import { Destroyable } from 'glimmer-util';
+
+export interface ModifierManager<T> {
+  // At initial render, the modifier gets a chance to install itself on the
+  // element it is managing. It can also return a bucket of state that
+  // it could use at update time. From the perspective of Glimmer, this
+  // is an opaque token.
+  install(element: Element, args: EvaluatedArgs, dom: DOMHelper): T;
+
+  // When the args have changed, the modifier's `update` hook is called
+  // with its state bucket as well as the updated args.
+  update(modifier: T, element: Element, args: EvaluatedArgs, dom: DOMHelper);
+
+  // Convert the opaque token into an object that implements Destroyable.
+  // If it returns null, the modifier will not be destroyed.
+  getDestructor(modifier: T): Destroyable;
+}
--- a/packages/glimmer-runtime/lib/syntax/core.ts
+++ b/packages/glimmer-runtime/lib/syntax/core.ts
@@ -31,6 +31,10 @@ import {
   CloseComponentOpcode
 } from '../compiled/opcodes/component';
 
+import {
+  ModifierOpcode
+} from '../compiled/opcodes/dom';
+
 import buildExpression from './expressions';
 
 import {
@@ -220,15 +224,21 @@ export class Append extends StatementSyntax {
   }
 }
 
-/*
-export class Modifier implements StatementSyntax {
+export const MODIFIER_SYNTAX = "c0420397-8ff1-4241-882b-4b7a107c9632";
+
+export class Modifier extends StatementSyntax {
+  "c0420397-8ff1-4241-882b-4b7a107c9632" = true;
+
+  public type: string = "modifier";
+  public path: InternedString[];
+  public args: Args;
+
   static fromSpec(node) {
     let [, path, params, hash] = node;
 
     return new Modifier({
       path,
-      params: Params.fromSpec(params),
-      hash: Hash.fromSpec(hash)
+      args: Args.fromSpec(params, hash)
     });
   }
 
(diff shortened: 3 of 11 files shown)


packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -221,6 +221,34 @@ export class StaticAttrOpcode extends Opcode {
   }
 }
 
+export class ModifierOpcode extends Opcode {
```
chadhietala: Is this meant to be an abstract class?

zackthehuman: It makes sense for it to be abstract once we start implementing actions (i.e.: soon). Otherwise we can't create instances and see them in the visualizer.

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -220,15 +224,17 @@ export class Append extends StatementSyntax {
   }
 }
 
-/*
-export class Modifier implements StatementSyntax {
+export class Modifier extends StatementSyntax {
+  public type: string = "modifier";
+  public path;
```
chancancode: missing type (`string` I think)

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -241,16 +247,19 @@ export class Modifier implements StatementSyntax {
...
 
-  evaluate(stack) {
-    return stack.createMorph(Modifier);
+  compile(compiler: CompileInto, env: Environment) {
+    compiler.append(new ModifierOpcode({ name: intern(this.path) }));
+  }
+
+  prettyPrint(): PrettyPrint {
+    return new PrettyPrint('modifier', this.path, [this.args.prettyPrint()]);
   }
```
chancancode: `prettyPrint` is legacy, 🔥. It was added for the old version of the visualizer but is now completely replaced by `toJSON`. If something else is giving you a `TypeError` for not having it, you can probably do a sweep in the codebase to remove all of them

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -241,16 +247,19 @@ export class Modifier implements StatementSyntax {
...
     this.path = options.path;
-    this.params = options.params;
-    this.hash = options.hash;
+    this.args = options.args;
   }
 
-  evaluate(stack) {
-    return stack.createMorph(Modifier);
+  compile(compiler: CompileInto, env: Environment) {
+    compiler.append(new ModifierOpcode({ name: intern(this.path) }));
```
chancancode: Probably need to take the `args` as well

chancancode:
This is looking great!

Some next steps:

1. Figure out how this composes with (Glimmer) components. For `<foo-bar {{mod}} />`, the `Modifier` opcode needs to be added after the inner `OpenPrimitiveElement` opcode. It also needs to "shadow" appropriately for recursive invocations. For now, it is probably sufficient to disallow modifiers on components.

2. Add two methods on the `Environment`:

    ```typescript
    abstract hasModifier(modifierName: InternedString[]): boolean;
    abstract lookupModifier(modifierName: InternedString[]): ModifierManager<Opaque>;
    ```
    
    For this step, you can just `ModifierManager` to be something like `type ModifierManager<T> = Opaque;`.
    
    When compiling the `Modifier` syntax, check if the modifier exists and pass it to the `ModifierOpcode` or raise an error otherwise. You can probably steal the code from the `Helper` syntax.

3. Actually make the modifier interface. We suspect it would be quite similar to `ComponentManager`, something like:

    ```typescript
    export interface ModifierManager<T> {
      // At initial render, the modifier gets a chance to install itself on the
      // element it is managing. It can also return a bucket of state that
      // it could use at update time. From the perspective of Glimmer, this
      // is an opaque token.
      install(element: Element, args: EvaluatedArgs): T;

      // When the args have changed, the modifier's `update` hook is called
      // with its state bucket as well as the updated args.
      update(modifier: T, element: Element, args: EvaluatedArgs);

      // Convert the opaque token into an object that implements Destroyable.
      // If it returns null, the modifier will not be destroyed.
      getDestructor(modifier: T): Destroyable;
    }
    ```

    The `ModifierOpcode` will be given an `ModifierManager<Opaque>` (as per the previous step) – the generic `T` here is only used by the host to thread through type information it needs, but Glimmer itself never cares what's in it (hence an "opaque" state bucket).

    It looks up the current element from the element stack, and call `install` on it. It should also emit an `UpdateModifierOpcode` with the args, the manager and the state bucket. For now it doesn't have to do anything. 

    It also needs to add the modifier as a destroyable on the block if `getDestructor` returns a non-null value. The destroyable will be invoked automatically when the surrounding block is torn down (e.g. and `{{#if}}` switched from default to inverse). You shouldn't have to do anything special here as long as you are calling `vm.newDestroyable` correctly.

    All of these should be quite similar to the `OpenComponentOpcode`.

4. Implement `UpdateModifierOpcode`. This should be quite similar to `UpdateComponentOpcode`. Basically, check if the args has changed, and if so, call `manager.update(modifier, args);`.

5. Write a test in the `TestEnvironment` for this feature. You can write a simple modifier that installs (and updates) a special property on the element, or add them to a registry, etc.

packages/glimmer-runtime/lib/syntax/core.ts
zackthehuman: Nothing in `core.ts` has a `toJSON` method. There are many `prettyPrint`s, though. I'll look in to it.

chancancode: That's because we stopped needing to inspect the syntax objects. We just do the wire format and opcodes now.

chancancode: Might not be a bad idea to eventually show them on the visualizer too

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -221,6 +223,85 @@ export class StaticAttrOpcode extends Opcode {
...
+    this.manager = manager;
+    this.element = element;
+    this.args = args;
+  }
+
+  evaluate(vm: UpdatingVM) {
+    // Need to:
+    // 1) Check if args has changed (need a cache?)
+    let { manager, modifier, element, args } = this;
+    manager.update(modifier, element, args);
```
zackthehuman: Not sure how to invalidate/check if the args are dirty here. Need input.

chancancode: You can skip that for now (just always call update - as long as your tests aren't asserting how many times that get called it shouldn't block you). I can show you the code later 👍

zackthehuman: I updated this based on our conversation and it updates things now by keeping track of the revision tag of the args.

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -221,6 +224,91 @@ export class StaticAttrOpcode extends Opcode {
...
+  private tag: RevisionTag;
+  private lastUpdated: Revision;
+
+  constructor({ manager, modifier, element, args }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element, args: EvaluatedArgs }) {
+    super();
+    this.modifier = modifier;
+    this.manager = manager;
+    this.element = element;
+    this.args = args;
+    this.tag = args.tag;
```
chancancode: Unused

packages/glimmer-runtime/lib/modifier/interfaces.ts
```diff
@@ -0,0 +1,18 @@
...
+  install(element: Element, args: EvaluatedArgs): T;
+
+  // When the args have changed, the modifier's `update` hook is called
+  // with its state bucket as well as the updated args.
+  update(modifier: T, element: Element, args: EvaluatedArgs);
+
+  // Convert the opaque token into an object that implements Destroyable.
+  // If it returns null, the modifier will not be destroyed.
+  getDestructor(modifier: T): Destroyable;
+}
```
chancancode: New line

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -790,3 +790,20 @@ test("Namespaced attribute with unquoted expression throws", function() {
...
+  compilesTo(
+    "<div {{foo}}></div>",
+    "<div></div>"
+  );
+
+  let div = root.firstChild;
+
+  equal(manager.installedElements.length, 1);
+  equal(div, manager.installedElements[0]);
+});
```
chancancode: New line

chancancode: Actually, can ✂️ this test. It was an early bootstrapping thing. Now all the new tests are INUR in the other file. So we don't need to add more tests here.

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -1793,3 +1793,68 @@ test("HTML namespace is continued to child templates", function() {
...
   equal(getSvg().firstChild.namespaceURI, SVG_NAMESPACE);
 });
+
+QUnit.module("Updating Element Modifiers", {
+  setup: commonSetup
+});
+
+test("Updating a element modifier", assert => {
+  let manager = new TestModifierManager();
+  env.registerInternalModifier('foo', manager);
```
chancancode: Don't need to call it "internal". The diff in helper is internal helpers work with references and the user helpers are just functions. It's not a glimmer distinction. Just how the TestEnvironment is set up. For modifiers we don't have that difference

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -1793,3 +1793,68 @@ test("HTML namespace is continued to child templates", function() {
...
+  equal(valueNode, manager.installedElements[0]);
+  equal(manager.updatedElements.length, 0);
+
+  rerender();
+
+  equalTokens(root, '<div></div>', "no change");
+  equal(manager.installedElements.length, 1);
+  equal(valueNode, manager.installedElements[0]);
+  equal(manager.updatedElements.length, 0);
+});
```
chancancode: New line

chancancode:
Looks great to me! Can you add a test to make sure they are torn down correctly? Also, would prefer that At least one of the tests modifiers does something observable - like installing and  a custom property or attribute on the element. I think that might suggest that we should be passing the DOM helper to the modifiers.

packages/glimmer-runtime/tests/updating-test.ts
chadhietala: @zackthehuman if you use VS Code this formatting happens for free 😄

zackthehuman: @chadhietala I have `"files.trimTrailingWhitespace": true` set in my settings so I think it negates any setting otherwise. It's weird because I'm used to having a newline at the end since that's what my Sublime does. This is annoying.

chancancode: I don't think it do it automatically. That setting you mentioned does a different thing (trim the extra spaces at the end of each line, which is good).

You want this: https://github.com/SamVerschueren/vscode-final-newline

zackthehuman:
I added a test that tracks the destruction of the modifier. It turns out that having the `DOMHelper` around is pretty handy so I also pass it it during `install`.

packages/glimmer-runtime/lib/modifier/interfaces.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface ModifierManager<T> {
+  // At initial render, the modifier gets a chance to install itself on the
+  // element it is managing. It can also return a bucket of state that
+  // it could use at update time. From the perspective of Glimmer, this
+  // is an opaque token.
+  install(element: Element, args: EvaluatedArgs, dom: DOMHelper): T;
+
+  // When the args have changed, the modifier's `update` hook is called
+  // with its state bucket as well as the updated args.
+  update(modifier: T, element: Element, args: EvaluatedArgs);
```
chancancode: we should probably pass the DOMHelper here too (you might have to reapply an attribute, etc)

zackthehuman: Done. Added checks in the test for updates.

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -544,8 +563,13 @@ export class OpenElement extends StatementSyntax {
     let current = scanner.next();
     let args = dict<ExpressionSyntax<Opaque>>();
     let attrs: InternedString[] = [];
+    let isModifier = !!current[MODIFIER_SYNTAX];
```
chancancode: I don't think we can cache this – if we have `<foo-bar bat="baz" {{modifier}} />` the first thing we see would be an attribute syntax (bat="baz"), followed by a modifier. In this case, because we never updated `isModifier` inside the loop, it will always be false. (Probably don't need the `!!` also)

zackthehuman: Doh. Fixed.

PR #148, opened 2016-04-24
Fix helpers in attrs & some refactors

mmun:
Previously we tried to "unwrap" MustacheStatements in attributes values like `<div foo={{bar}}>` to a `PathExpression`. This has the unexpected consequence of not allow you to invoke `bar` if it was a helper that took no arguments.

Related ember issue: https://github.com/emberjs/ember.js/issues/12506


rwjblue:
Thanks for cleaning this up @mmun!

PR #150, opened 2016-04-27
Set privilege for falsy class names

chadhietala:


chadhietala:
See https://github.com/emberjs/ember.js/pull/13401#discussion-diff-61283668 for the use case.

PR #151, opened 2016-04-27
Organize opcodes into a builder structure

wycats:
This is a simple re-org with tests still passing, but it should allow
more experimentation with the process of building opcodes, including
simplifying what an opcode is and allowing abstractions that build
several opcodes at once.


mmun:
I like it :+1:

PR #156, opened 2016-04-29
Export UNDEFINED_REFERENCE from glimmer-runtime

zackthehuman:
Exporting the `UNDEFINED_REFERENCE` so that it can be used inside of Ember.

### DIFF
--- a/packages/glimmer-runtime/index.ts
+++ b/packages/glimmer-runtime/index.ts
@@ -13,7 +13,7 @@ export { default as Template } from './lib/template';
 
 export { default as SymbolTable } from './lib/symbol-table';
 
-export { ConditionalReference } from './lib/references';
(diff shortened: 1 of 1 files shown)


chancancode:
Can you export `NULL_REFERENCE` as well? If it's not already defined in the references package, you can move it.

chancancode:
Also, I just realized this: in https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/references.ts#L8, it probably makes more sense to return an `UNDEFINED_REFERENCE` now that we have one.

zackthehuman:
@chancancode Sure, will send another PR :)

zackthehuman:
PR for NULL_REFERENCE: https://github.com/tildeio/glimmer/pull/157

PR #160, opened 2016-05-02
Pass DynamicScope to modifier install hook

zackthehuman:
It turns out that having the DyntamicScope is pretty handy for real-world modifiers. This updates the interface so that the scope is passed when a modifier is installed.


chancancode:
If we want that we should also pass it to the update hook for parity

zackthehuman:
Updated. Passing DynamiScope along to `update` as well.

PR #162, opened 2016-05-03
Add getAttribute method to DOMHelper

zackthehuman:
Adds a method to `getAttribute` as a compliment to `setAttribute`. This helper is needed for element actions since they'll need to get/set the `data-ember-action` attribute.

### DIFF
--- a/packages/glimmer-runtime/lib/dom.ts
+++ b/packages/glimmer-runtime/lib/dom.ts
@@ -53,6 +53,10 @@ class DOMHelper {
     this.uselessElement = this.document.createElement('div');
   }
 
+  getAttribute(element: Element, name: string) {
+    return element.getAttribute(name);
+  }
+
   setAttribute(element: Element, name: string, value: string) {
     element.setAttribute(name, value);
   }


chadhietala:
LGTM

wycats:
Why is reading the `data-ember-action` attribute needed for `{{action}}`

krisselden:
@wycats it is to coalesce multiple action modifiers on one element via the first one that sets the attribute

https://github.com/emberjs/ember.js/blob/783a656e105a7121fc0dc38374931880d14b6bc3/packages/ember-routing-htmlbars/lib/keywords/element-action.js#L47

krisselden:
which is kind of weird because in that case more than one cleanup for the same id would happen in the old code https://github.com/emberjs/ember.js/blob/783a656e105a7121fc0dc38374931880d14b6bc3/packages/ember-routing-htmlbars/lib/keywords/element-action.js#L60

zackthehuman:
Yeah, for actions in `ember-glimmer` I will unregister using the actionID as well as the actionState object, so it will only remove the actionState it's supposed to, one at a time.

Regardless, it needs to check for and grab the `data-ember-action` attribute in order to facilitate the binding as Kris mentioned.

PR #166, opened 2016-05-06
Allow iterators to return current position (index) during {{#each}}

Joelkang:
In the `opcodes/lists.ts` positional params only takes PathReferences. Not entirely sure if I should convert it to take any kind of reference, or if my casting to <any> is acceptable.

### DIFF
--- a/packages/glimmer-reference/lib/iterable.ts
+++ b/packages/glimmer-reference/lib/iterable.ts
@@ -160,9 +160,9 @@ export class ReferenceIterator {
 }
 
 export interface IteratorSynchronizerDelegate {
-  retain(key: InternedString, item: PathReference<any>, memo: PathReference<any>);
-  insert(key: InternedString, item: PathReference<any>, memo: PathReference<any>, before: InternedString);
-  move(key: InternedString, item: PathReference<any>, memo: PathReference<any>, before: InternedString);
+  retain(key: InternedString, item: PathReference<Opaque>, memo: PathReference<Opaque>);
+  insert(key: InternedString, item: PathReference<Opaque>, memo: PathReference<Opaque>, before: InternedString);
+  move(key: InternedString, item: PathReference<Opaque>, memo: PathReference<Opaque>, before: InternedString);
   delete(key: InternedString);
   done();
 }
--- a/packages/glimmer-reference/lib/iterable.ts
+++ b/packages/glimmer-reference/lib/iterable.ts
@@ -9,7 +9,7 @@ export interface IterationItem<T> {
 
 export interface AbstractIterator<T, U extends IterationItem<T>> {
   isEmpty(): boolean;
-  getPosition(): number;
+  getPosition(): number | FIXME<'user str to InternedString'>;
   next(): U;
 }
 
@@ -140,7 +140,7 @@ export class ReferenceIterator {
     this.artifacts = artifacts;
   }
 
-  getPosition(): ConstReference<number> {
+  getPosition(): ConstReference< number | FIXME<'user str to InternedString'>> {
     return new ConstReference(this.iterator.getPosition());
   }
 
--- a/packages/glimmer-reference/lib/iterable.ts
+++ b/packages/glimmer-reference/lib/iterable.ts
@@ -1,5 +1,6 @@
 import { FIXME, LinkedList, ListNode, InternedString, Opaque, dict } from 'glimmer-util';
 import { VersionedPathReference as PathReference } from './validators';
+import { ConstReference } from './const';
 
 export interface IterationItem<T> {
   key: string;
@@ -8,6 +9,7 @@ export interface IterationItem<T> {
 
 export interface AbstractIterator<T, U extends IterationItem<T>> {
   isEmpty(): boolean;
+  getPosition(): number;
   next(): U;
 }
 
@@ -136,6 +138,10 @@ export class ReferenceIterator {
   constructor(iterable: OpaqueIterable) {
     let artifacts = new IterationArtifacts(iterable);
     this.artifacts = artifacts;
+  }
+
+  getPosition(): ConstReference<number> {
+    return new ConstReference(this.iterator.getPosition());
   }
 
   next(): IterationItem<PathReference<Opaque>> {
--- a/packages/glimmer-reference/lib/iterable.ts
+++ b/packages/glimmer-reference/lib/iterable.ts
@@ -1,6 +1,6 @@
 import { FIXME, LinkedList, ListNode, InternedString, Opaque, dict } from 'glimmer-util';
 import { VersionedPathReference as PathReference } from './validators';
-import { ConstReference } from './const';
+import { UpdatableReference } from 'glimmer-object-reference';
 
 export interface IterationItem<T> {
   key: string;
@@ -31,16 +31,19 @@ class ListItem extends ListNode<PathReference<Opaque>> implements IterationItem<
   public retained: boolean = false;
   public seen: boolean = false;
   private iterable: OpaqueIterable;
+  public position: UpdatableReference<number | FIXME<'user string to InternedString'>>;
 
-  constructor(iterable: OpaqueIterable, result: OpaqueIterationItem) {
+  constructor(iterable: OpaqueIterable, result: OpaqueIterationItem, position: number | FIXME<'user string to InternedString'>) {
     super(iterable.referenceFor(result));
     this.key = result.key as FIXME<'user string to InternedString'>;
     this.iterable = iterable;
+    this.position = new UpdatableReference(position);
   }
 
-  update(item: OpaqueIterationItem) {
+  update(item: OpaqueIterationItem, position: number | FIXME<'user string to InternedString'>) {
     this.retained = true;
     this.iterable.updateReference(this.value, item);
+    this.position.update(position);
   }
 
   shouldRemove(): boolean {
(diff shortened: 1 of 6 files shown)


chadhietala:
This looks like it will fix `{{#each}}` as well.

Joelkang:
Yea, not 100% sure where I should add tests for this tbh; could use some direction there

chadhietala:
@Joelkang Are you implementing `each-in` in glimmer or Ember? This might be a question for @wycats. You can probably add the tests here https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/tests/updating-test.ts or a better idea may be to make a new file in the runtime tests for this feature.

Joelkang:
I'm implementing it in ember but this is a prerequisite for that. I
actually already have an ember prototype but let me verify with test before
I throw up a PR. Will write tests for this in the meanwhile; thanks for the
pointer

On Friday, May 6, 2016, Chad Hietala <notifications@github.com> wrote:

> @Joelkang <https://github.com/Joelkang> Are you implementing each-in in
> glimmer or Ember? This might be a question for @wycats
> <https://github.com/wycats>. You can probably add the tests here
> https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/tests/updating-test.ts
> or a better idea may be to make a new file in the runtime tests for this
> feature.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/tildeio/glimmer/pull/166#issuecomment-217585051>
>


-- 
Cheers,
Joel

packages/glimmer-reference/lib/iterable.ts
```diff
@@ -138,6 +140,10 @@ export class ReferenceIterator {
     this.artifacts = artifacts;
   }
 
+  getPosition(): ConstReference< number | FIXME<'user str to InternedString'>> {
```
krisselden: should be a ValueReference

krisselden:
Looks like this just failed because of the typescript version being unpinned.  I submitted a fix https://github.com/tildeio/glimmer/pull/169

packages/glimmer-reference/lib/iterable.ts
```diff
@@ -8,6 +9,7 @@ export interface IterationItem<T> {
 
 export interface AbstractIterator<T, U extends IterationItem<T>> {
   isEmpty(): boolean;
+  getPosition(): number | FIXME<'user str to InternedString'>;
```
krisselden: Should this be Opaque?

packages/glimmer-reference/lib/iterable.ts
```diff
@@ -86,18 +91,18 @@ export class IterationArtifacts {
...
     list.append(node);
     return node;
   }
 
-  insertBefore(item: OpaqueIterationItem, reference: ListItem): ListItem {
+  insertBefore(item: OpaqueIterationItem, reference: ListItem, position: number | FIXME<'user string to InternedString'>): ListItem {
     let { map, list, iterable } = this;
 
-    let node = map[item.key] = new ListItem(iterable, item);
+    let node = map[item.key] = new ListItem(iterable, item, position);
```
krisselden: why are we accepting a ListItem and a position, and then creating a new ListItem?

packages/glimmer-reference/lib/iterable.ts
```diff
@@ -138,22 +143,27 @@ export class ReferenceIterator {
...
 
     let iterator = (this.iterator = this.iterator || artifacts.iterate());
 
     let item = iterator.next();
+    let position = iterator.getPosition();
 
     if (!item) return null;
 
-    return artifacts.append(item);
+    return artifacts.append(item, position);
```
krisselden: I'm not sure why position is sometimes part of item and sometimes not.  Why isn't it like value where the list item is updated?

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -96,25 +97,77 @@ class ArrayIterator implements OpaqueIterator {
...
+
+  constructor(kvPairs: KVPair[]) {
+    this.array = kvPairs;
+  }
+
+  isEmpty(): boolean {
+    return this.array.length === 0;
+  }
+
+  getPosition(): string {
```
krisselden: this feels a little weird.

chancancode:
Great work!

Some suggestions:

As you have identified, the second arg is used for more than just "position"s. I think we should just call it something more generic. I think it would be a good idea to make that a generic "memo" slot that the iterator can use to yield additional information to the template.

The public contract (that Ember uses) becomes:

```typescript
export interface IterationItem<T, U> {
  key: string;
  value: T;
  memo: U;
 }

export interface AbstractIterator<T, U, V extends IterationItem<T, U>> {
  isEmpty(): boolean;
  next(): V;
}

export interface AbstractIterable<T, U, ItemType extends IterationItem<T, U>, ValueReferenceType extends PathReference<T>, MemoReferenceType extends PathReference<U>> {
  iterate(): AbstractIterator<T, ItemType>;

  valueReferenceFor(item: ItemType): ValueReferenceType;
  updateValueReference(reference: ValueReferenceType, item: ItemType);

  memoReferenceFor(item: ItemType): MemoReferenceType;
  updateMemoReference(reference: MemoReferenceType, item: ItemType);
}
```

This allows Ember's iterator to yield whatever it needs for the second arg (and use whatever reference types it sees fit). As far as Glimmer is concerned, it's just another opaque value we pass along to the template.

Once you have that, the rest should just fall into place with relatively little changes.

packages/glimmer-reference/lib/iterable.ts
Joelkang: I think the reason why did this was that it didn't make sense to me that the position was part of the iterationItem interface since it was meta-item. Because of this I wound up passing everything across the stack. Given both your and Godfrey's comments, I'll make it internal to IterationItem so there's no more verbose passing around

Joelkang:
@chancancode  @krisselden updated per your comments

packages/glimmer-reference/lib/iterable.ts
```diff
@@ -152,9 +160,9 @@ export class ReferenceIterator {
 }
 
 export interface IteratorSynchronizerDelegate {
-  retain(key: InternedString, item: PathReference<any>);
-  insert(key: InternedString, item: PathReference<any>, before: InternedString);
-  move(key: InternedString, item: PathReference<any>, before: InternedString);
+  retain(key: InternedString, item: PathReference<any>, memo: PathReference<any>);
```
krisselden: why are these any instead of Opaque?

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -80,14 +80,14 @@ import {
   UpdatableReference
 } from "glimmer-object-reference";
 
-type KeyFor = (item: Opaque, index: number) => string;
+type KeyFor = (item: Opaque, index: number | string) => string;
```
krisselden: why is this changed? seems like memo leaking into KeyFor

packages/glimmer-reference/lib/iterable.ts
chancancode: Good question. In general the answer is probably "this code predates the introduction of `Opaque`", so it might be time to fix it!

packages/glimmer-test-helpers/lib/environment.ts
chancancode: Since this is only used by the tests, I'm not too sad about it. We can probably make it `type keyFor<T> = (item: Opaque, memo: T) => string;`. Then `ArrayIterator` can have a `KeyFor<number>`, and `ObjectKeysIterator` can have a `KeyFor<string>`.

packages/glimmer-runtime/lib/vm/update.ts
```diff
@@ -1,7 +1,8 @@
 import { Scope, DynamicScope, Environment } from '../environment';
 import { Bounds, clear, move as moveBounds } from '../bounds';
 import { ElementStack, Tracker } from '../builder';
-import { LOGGER, Destroyable, Stack, LinkedList, InternedString, Dict, dict } from 'glimmer-util';
+import { LOGGER, Destroyable, Stack, FIXME, LinkedList, InternedString, Dict, dict } from 'glimmer-util';
```
chancancode: Some unused stuff here (`FIXME` is what my 👀 could spot, but tslint would tell you for sure)

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -146,18 +182,30 @@ class Iterable implements AbstractIterable<Opaque, IterationItem<Opaque>, Updata
       return array.length > 0 ? new ArrayIterator(array, keyFor) : EMPTY_ITERATOR;
     } else if (iterable === undefined || iterable === null) {
       return EMPTY_ITERATOR;
-    } else {
+    } else if (typeof iterable === 'object') {
+       let keys = Object.keys(iterable);
+       let kvPairs = keys.map(key => <KVPair>{ key, value: iterable[key] });
```
chancancode: Usually you wouldn't need to do an explicit cast like this. If it couldn't figure it out, something is probably wrong.

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -96,17 +96,53 @@ class ArrayIterator implements OpaqueIterator {
...
 
-    return { key, value };
+    return { key, value, memo };
+  }
+}
+
+type KVPair = {
+  key: string;
+  value: Opaque;
+};
```
chancancode: You could write

```typescript
interface KVPair {
  key: string;
  value: Opaque;
}
```

(Or the iterator could just take an array of keys and an array of values, since it's all internal it doesn't really matter)

chancancode:
This is awesome. Very happy with the way it is factored! 💯 

(Also, thank you for taking the time to write those tests, really appreciate it!)

packages/glimmer-runtime/lib/vm/update.ts
krisselden: tslint in now part of CI so that should fail the build.

PR #168, opened 2016-05-09
Support fragments

asakusuma:
go tagless

- [X] Support compile-time tagless components
- [x] Support dynamic tagless components

### DIFF
--- a/packages/glimmer-runtime/lib/compiler.ts
+++ b/packages/glimmer-runtime/lib/compiler.ts
@@ -19,7 +19,9 @@ import {
   ExitOpcode,
   LabelOpcode,
   PutArgsOpcode,
-  PutValueOpcode
+  PutValueOpcode,
+  JumpUnlessOpcode,
+  TestOpcode
 } from './compiled/opcodes/vm';
 
 import * as Syntax from './syntax/core';
@@ -251,12 +253,18 @@ class WrappedBuilder {
       list.append(BindBlocksOpcode.create(layout));
     }
 
+    let tagExpr;
+
     if (this.tag.isDynamic) {
-      let tag = makeFunctionExpression(this.tag.dynamicTagName);
-      list.append(new PutValueOpcode({ expression: tag.compile(list, env) }));
+      let BODY = new LabelOpcode({ label: 'BODY' });
+      tagExpr = makeFunctionExpression(this.tag.dynamicTagName).compile(list, env);
+      list.append(new PutValueOpcode({ expression: tagExpr }));
+      list.append(new TestOpcode());
+      list.append(new JumpUnlessOpcode({ target: BODY }));
       list.append(new OpenDynamicPrimitiveElementOpcode());
       list.append(new DidCreateElementOpcode());
       this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
+      list.append(BODY);
     } else if(this.tag.isStatic) {
       let tag = this.tag.staticTagName;
       list.append(new OpenPrimitiveElementOpcode({ tag }));
@@ -266,7 +274,14 @@ class WrappedBuilder {
 
     layout.program.forEachNode(statement => compileStatement(env, statement, list));
 
-    if (this.tag.isDynamic || this.tag.isStatic) {
+    if (this.tag.isDynamic) {
+      let END = new LabelOpcode({ label: 'END' });
+      list.append(new PutValueOpcode({ expression: tagExpr }));
+      list.append(new TestOpcode());
+      list.append(new JumpUnlessOpcode({ target: END }));
+      list.append(new CloseElementOpcode());
+      list.append(END);
+    } else if(this.tag.isStatic) {
       list.append(new CloseElementOpcode());
     }
 
--- a/packages/glimmer-runtime/lib/compiler.ts
+++ b/packages/glimmer-runtime/lib/compiler.ts
@@ -255,18 +255,20 @@ class WrappedBuilder {
       let tag = makeFunctionExpression(this.tag.dynamicTagName);
       list.append(new PutValueOpcode({ expression: tag.compile(list, env) }));
       list.append(new OpenDynamicPrimitiveElementOpcode());
-    } else {
+      list.append(new DidCreateElementOpcode());
+      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
+    } else if(this.tag.isStatic) {
       let tag = this.tag.staticTagName;
       list.append(new OpenPrimitiveElementOpcode({ tag }));
+      list.append(new DidCreateElementOpcode());
+      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
     }
-
-    list.append(new DidCreateElementOpcode());
-
-    this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
 
     layout.program.forEachNode(statement => compileStatement(env, statement, list));
 
-    list.append(new CloseElementOpcode());
+    if (this.tag.isDynamic || this.tag.isStatic) {
+      list.append(new CloseElementOpcode());
+    }
 
     return new CompiledBlock(list, layout.symbolTable.size);
   }
@@ -326,11 +328,12 @@ function isOpenElement(syntax: StatementSyntax): syntax is OpenElement {
 
 class ComponentTagBuilder implements Component.ComponentTagBuilder {
   public isDynamic = null;
+  public isStatic = null;
   public staticTagName: InternedString = null;
   public dynamicTagName: FunctionExpression<string> = null;
 
   static(tagName: InternedString) {
-    this.isDynamic = false;
+    this.isStatic = true;
     this.staticTagName = tagName;
   }
 
--- a/packages/glimmer-runtime/tests/ember-component-test.ts
+++ b/packages/glimmer-runtime/tests/ember-component-test.ts
@@ -889,6 +889,21 @@ QUnit.test('dynamic tagName', assert => {
   assertEmberishElement('aside', {}, `Hello. It's me.`);
 });
 
+QUnit.test('dynamic tagless component', assert => {
+  class FooBar extends EmberishCurlyComponent {
+    tagName = '';
+  }
+
+  env.registerEmberishCurlyComponent('foo-bar', FooBar, `Michael Jordan says "Go Tagless"`);
+
+  appendViewFor(`{{foo-bar}}`);
+  assertAppended('Michael Jordan says "Go Tagless"');
+
+  rerender();
+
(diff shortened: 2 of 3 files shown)


asakusuma:
Travis is failing because of typescript failures. Tests are functionally passing.

packages/glimmer-runtime/lib/builder.ts
```diff
@@ -329,7 +329,13 @@ class BlockTracker implements Tracker {
   }
 
   parentElement() {
-    return this.parent;
+    let parent = this.parent;
+    let firstNode = this.firstNode();
+    if (firstNode && parent && parent.children && parent.children.length === 0) {
```
krisselden: check for fragment should be `parent.nodeType === Node.DOCUMENT_FRAGMENT_NODE` and we should update this.parent to be the element but this really should be done when it is appended

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -356,7 +356,7 @@ export class NamespacedAttribute extends ElementPatchOperation<string> {
   private namespace: InternedString;
 
   static install(dom: DOMHelper, element: Element, namespace: InternedString, name: InternedString, value: string) {
-    if (value !== null) {
+    if (value !== null && element.tagName) {
```
krisselden: check for element should be `element.nodeType === Node.ELEMENT_NODE` and the typing should be `element: Element | DocumentFragment`

packages/glimmer-runtime/lib/dom.ts
```diff
@@ -153,6 +153,9 @@ class DOMHelper {
   }
 
   insertBefore(element: Element, node: Node, reference: Node) {
+    if (reference && element.children.length === 0) {
```
krisselden: same comment as above regarding type and nodeType check.

krisselden: This shouldn't be changed to accepting DocumentFragment, it should be resolved in the caller.

asakusuma:
I talked to Kris directly, I'm going to take another approach and try to use a `Bounds` to represent the fragment instead of a `DocumentFragment`

chancancode:
@asakusuma I think there is a slight confusion here :smile:

What we call "fragments" in Ember actually has nothing to do with `DocumentFragment`s in the DOM API. "Fragments" in Ember just means `Ember.Component` without an wrapper element. Certain APIs are disabled (`this.element`, `this.$`, attribute bindings, etc), but that's about it.

I don't think we actually need to implement this feature in Glimmer using `DocumentFragment`s. Actually, most of the Glimmer system already doesn't care about whether a component has an element or not. 

The main place you need to change is the component builder API. Specifically, the [`WrappedBuilder`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L229-L273) is [what Ember actually uses](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/components/curly-component.js#L175-L178) today.

[How it works today](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L256-L257) is that we put the value of the [component's `tagName` property](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/components/curly-component.js#L153-L161) [on a register](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/vm.ts#L74-L76), and then we [read it off from the register and call `openElement` with that value](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts#L69-L70).

If we know that a component is a fragment at compile time, this will be pretty easy. Specifically, if you look at the `compile` method on the `WrappedBuilder`, all we need to do is:

1. Don't emit the [`OpenPrimitiveElementOpcode`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L260)
2. Don't emit the [`DidCreateElementOpcode`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L263)
3. Don't compile/emit the [attributes statements/opcodes](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L265)
4. Don't emit the [`CloseElementOpcode`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiler.ts#L269)

If you did that, the method would probably look like this:

```typescript
class WrappedBuilder {
  compile(): CompiledBlock {
    let { env, layout } = this;

    let list = new CompileIntoList(env, layout.symbolTable);

    if (layout.hasNamedParameters()) {
      list.append(BindNamedArgsOpcode.create(layout));
    }

    if (layout.hasYields()) {
      list.append(BindBlocksOpcode.create(layout));
    }

    if (this.tag.isDynamic) {
      let tag = makeFunctionExpression(this.tag.dynamicTagName);
      list.append(new PutValueOpcode({ expression: tag.compile(list, env) }));
      list.append(new OpenDynamicPrimitiveElementOpcode());
      list.append(new DidCreateElementOpcode());
      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
    } else if (this.tag.isStatic) {
      let tag = this.tag.staticTagName;
      list.append(new OpenPrimitiveElementOpcode({ tag }));
      list.append(new DidCreateElementOpcode());
      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
    } else if {
      // no tag, no attributes
    }

    layout.program.forEachNode(statement => compileStatement(env, statement, list));

    if (this.tag.isDynamic || this.tag.isStatic) {
      list.append(new CloseElementOpcode());
    }

    return new CompiledBlock(list, layout.symbolTable.size);
  }
}
```

I think you should probably start with adding support for that (statically known "fragment components"), and add a test for that. It will be a good exercise in figuring out how the machinery works.

However, this is not immediately useful for Ember, because we never know ahead of time – there is no "syntax" for a fragment component. The only time we know a component is a fragment is after we have instantiated the JS instance and its `tagName` property is an empty string. i.e. all `Ember.Component` has a "dynamic" (reads: only known at runtime, as opposed to the opcode-compilation time) tag.

So the dynamic tag branch needs to support the fragment case too.

What we need to do is:

1. In `OpenDynamicPrimitiveElementOpcode`, we should check that if the `tagName` (the value on the operand register) is empty. If it is empty, do not call `createElement`. The operations on the element stack have well-defined semantics, and we should not leak this detail to that part of the system (i.e. make `createElement("")` create a `DocumentFragment`).

    Side note: I think `null` makes more sense for Glimmer than an empty string – let's make that the Glimmer protocol for "no tag plz" and change the ember side to normalize it.

2. We need a `DidCreateDynamicElementOpcode` that checks if the operand register is `null` and not call the hook on the component manager.

3. We need to jump pass the attribute opcodes we didn't create an element.

4. We need a `CloseDynamicElementOpcode` that checks if the operand register is `null` and do nothing

So the code would probably look like this:

```typescript
class WrappedBuilder {
  compile(): CompiledBlock {
    let { env, layout } = this;

    let list = new CompileIntoList(env, layout.symbolTable);

    if (layout.hasNamedParameters()) {
      list.append(BindNamedArgsOpcode.create(layout));
    }

    if (layout.hasYields()) {
      list.append(BindBlocksOpcode.create(layout));
    }

    let tagExpr;

    if (this.tag.isDynamic) {
      let BODY = new LabelOpcode({ label: "BODY" });
      tagExpr = makeFunctionExpression(this.tag.dynamicTagName).compile(list, env);
      list.append(new PutValueOpcode({ expression: tagExpr }));
      list.append(new OpenDynamicPrimitiveElementOpcode());
      list.append(new DidCreateDynamicElementOpcode());
      list.append(new JumpIfNullOpcode({ target: BODY })); // <- make this, see https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/vm.ts#L348-L368
      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
      list.append(BODY);
    } else if (this.tag.isStatic) {
      let tag = this.tag.staticTagName;
      list.append(new OpenPrimitiveElementOpcode({ tag }));
      list.append(new DidCreateElementOpcode());
      this.attrs['buffer'].forEach(statement => compileStatement(env, statement, list));
    } else if {
      // no tag, no attributes
    }

    layout.program.forEachNode(statement => compileStatement(env, statement, list));

    if (this.tag.isDynamic) {
      list.append(new PutValueOpcode({ expression: tagExpr });
      list.append(new CloseDynamicElementOpcode());
    } else if (this.tag.isStatic) {
      list.append(new CloseElementOpcode());
    }

    return new CompiledBlock(list, layout.symbolTable.size);
  }
}
```

asakusuma:
@chancancode thanks, this is super helpful. This makes a lot more sense now. I'll give this a shot today.

chancancode:
@asakusuma I realized something stupid. If we are adding a jump opcode anyway, we don't have to change `OpenDynamicPrimitiveElementOpcode` and `DidCreateDynamicElementOpcode` and `CloseElementOpcode` to do anything special. We can just jump pass them :smile:

```
      PUT-VALUE(tagNameExpr)
      JUMP-IF-NULL($operand, target: BODY)
      OPEN-DYNAMIC-ELEMENT($operand)
      DID-CREATE-ELEMENT
      ...attr statements...
BODY: NOOP
      ...body statments...
      PUT-VALUE(tagNameExpr)
      JUMP-IF-NULL($operand, target: END)
      CLOSE-ELEMENT
END:  NOOP
```

asakusuma:
@chancancode ready for another review. I didn't really do any factoring beyond what you prescribed though.

packages/glimmer-runtime/lib/compiler.ts
```diff
@@ -251,22 +253,37 @@ class WrappedBuilder {
...
 
+    let tagExpr;
+
     if (this.tag.isDynamic) {
-      let tag = makeFunctionExpression(this.tag.dynamicTagName);
-      list.append(new PutValueOpcode({ expression: tag.compile(list, env) }));
+      let BODY = new LabelOpcode({ label: 'BODY' });
+      tagExpr = makeFunctionExpression(this.tag.dynamicTagName).compile(list, env);
+      list.append(new PutValueOpcode({ expression: tagExpr }));
+      list.append(new TestOpcode());
```
chancancode: :+1:

chancancode:
👏 @asakusuma can you add the [comments](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/syntax/builtins/if.ts#L40-L50) at the beginning of the `compile` method? (I guess we have to document three possible paths now)

asakusuma:
@chancancode will do

PR #172, opened 2016-05-14
Implement partials

asakusuma:
Big thanks to @mmun for getting this started walking me through this.

cc @chadhietala

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/blocks.ts
+++ b/packages/glimmer-runtime/lib/compiled/blocks.ts
@@ -63,6 +63,16 @@ export class InlineBlock extends Block {
   }
 }
 
+export class PartialBlock extends InlineBlock {
+  initBlocks(blocks = this['children'], parentTable = this['symbolTable']): this {
+    blocks.forEach(block => {
+      let table = SymbolTable.initForBlock({ parent: parentTable, block });
+      this.initBlocks(block['children'], table);
+    });
+    return this;
+  }
+}
+
 export abstract class TopLevelTemplate extends Block {
   initBlocks(blocks = this['children'], parentTable = this['symbolTable']): this {
     blocks.forEach(block => {
--- a/packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
@@ -1,12 +1,21 @@
 import { AbstractOpcode, Opcode, OpcodeJSON, UpdatingOpcode } from '../../opcodes';
 import { CompiledExpression } from '../expressions';
-import { CompiledArgs } from '../expressions/args';
+import { CompiledArgs, EvaluatedArgs } from '../expressions/args';
 import { VM, UpdatingVM, BindDynamicScopeCallback } from '../../vm';
-import { Layout, InlineBlock } from '../blocks';
+import { Layout, InlineBlock, PartialBlock } from '../blocks';
 import { turbocharge } from '../../utils';
 import { NULL_REFERENCE } from '../../references';
-import { ListSlice, Opaque, Slice, Dict, dict, assign } from 'glimmer-util';
+import SymbolTable from '../../symbol-table';
+import { PathReference } from 'glimmer-reference';
+import { ValueReference } from '../expressions/value';
+import { ListSlice, Opaque, Slice, Dict, dict, assign, InternedString } from 'glimmer-util';
 import { CONSTANT_TAG, ReferenceCache, Revision, RevisionTag, isConst, isModified } from 'glimmer-reference';
+import {
+  CompileInto,
+  SymbolLookup,
+  Statement as StatementSyntax
+} from '../../syntax';
+import Scanner from '../../scanner';
 
 export class PushChildScopeOpcode extends Opcode {
   public type = "push-child-scope";
@@ -306,6 +315,70 @@ export class EvaluateOpcode extends Opcode {
   }
 }
 
+export class EvaluatePartialOpcode extends Opcode {
+  public type = "evaluate-partial";
+  public symbolTable: SymbolTable;
+  public name: CompiledExpression<any>;
+  private cache = dict<PartialBlock>();
+
+  constructor({ name, symbolTable }: { symbolTable: SymbolTable, name: CompiledExpression<any> }) {
+    super();
+    this.name = name;
+    this.symbolTable = symbolTable;
+  }
+
+  evaluate(vm: VM) {
+    let reference: PathReference<any> = this.name.evaluate(vm);
+    let name: InternedString = reference.value();
+
+    let block = this.cache[name];
+    if (!block) {
+      let { template } = vm.env.lookupPartial([name]);
+      let scanner = new Scanner(template, vm.env);
+      block = scanner.scanPartial(this.symbolTable);
+    }
+
+    vm.invokeBlock(block, EvaluatedArgs.empty());
+
+    if (!isConst(reference)) {
+      let cache = new ReferenceCache(reference);
+      vm.updateWith(new Assert(cache));
+    }
+  }
+
+  toJSON(): OpcodeJSON {
+    return {
+      guid: this._guid,
+      type: this.type,
+      args: [this.name.toJSON()]
+    };
+  }
+}
+
+export class NameToPartialOpcode extends Opcode {
+  public type = "name-to-partial";
+
+  evaluate(vm: VM) {
+    let reference = vm.frame.getOperand();
+    let name: InternedString = reference.value();
+    let partial = name && vm.env.hasPartial([name]) ? vm.env.lookupPartial([name]) : false;
+    vm.frame.setOperand(new ValueReference(partial));
+
+    if (!isConst(reference)) {
+      let cache = new ReferenceCache(reference);
+      vm.updateWith(new Assert(cache));
+    }
+  }
+
+  toJSON(): OpcodeJSON {
+    return {
+      guid: this._guid,
+      type: this.type,
+      args: ["$OPERAND"]
+    };
+  }
+}
+
 export class TestOpcode extends Opcode {
   public type = "test";
 
@@ -477,4 +550,3 @@ export class DidModifyOpcode extends UpdatingOpcode {
     this.target.didModify();
   }
 }
-
--- a/packages/glimmer-runtime/lib/partial.ts
+++ b/packages/glimmer-runtime/lib/partial.ts
@@ -1,0 +1,12 @@
+import { SerializedTemplate } from 'glimmer-wire-format';
+
+export class PartialDefinition {
+  name: string;
+  template: SerializedTemplate;
+
+  constructor(name: string, template: SerializedTemplate) {
+    this.name = name;
+    this.template = template;
+  }
+
+}
--- a/packages/glimmer-runtime/lib/scanner.ts
+++ b/packages/glimmer-runtime/lib/scanner.ts
@@ -1,9 +1,10 @@
 import { Program, Statement as StatementSyntax } from './syntax';
 import buildStatement from './syntax/statements';
-import { TopLevelTemplate, EntryPoint, InlineBlock, Layout } from './compiled/blocks';
+import { TopLevelTemplate, EntryPoint, InlineBlock, PartialBlock, Layout } from './compiled/blocks';
 import Environment from './environment';
 import { EMPTY_SLICE, LinkedList, Stack } from 'glimmer-util';
 import { SerializedTemplate, SerializedBlock, Statement as SerializedStatement } from 'glimmer-wire-format';
+import SymbolTable from './symbol-table';
 
 export default class Scanner {
   private spec: SerializedTemplate;
@@ -24,6 +25,13 @@ export default class Scanner {
     return this.scanTop<Layout>(({ program, children }) => {
       let { named, yields } = this.spec;
       return Layout.create({ children, program, named, yields, symbolTable: null });
+    });
+  }
+
+  scanPartial(symbolTable: SymbolTable): PartialBlock {
+    return this.scanTop<PartialBlock>(({ program, children }) => {
+      let { locals } = this.spec;
+      return new PartialBlock({ children, program, locals, symbolTable });
     });
   }
 
--- a/packages/glimmer-runtime/lib/syntax/builtins/partial.ts
+++ b/packages/glimmer-runtime/lib/syntax/builtins/partial.ts
@@ -1,0 +1,79 @@
+import { assert } from "glimmer-util";
+
+import {
+  CompileInto,
+  SymbolLookup,
+  Statement as StatementSyntax
+} from '../../syntax';
+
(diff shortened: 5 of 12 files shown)


packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+        name
+      }));
+    }
+  }
+
+  toJSON(): OpcodeJSON {
+    return {
+      guid: this._guid,
+      type: this.type,
+      args: []
```
asakusuma: What is supposed to go here?

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+    this.compiler = compiler;
+    this.nameReference = nameReference;
+    this.name = name;
+  }
+
+  evaluate(vm: UpdatingVM) {
+    let name = this.nameReference.value();
+
+    if (name !== this.name) {
+      vm.throw();
```
asakusuma: I feel like this is not the right way to force a re-render when the partial name changes

mmun:
Can your rebase my commit out? 💃

asakusuma:
@mmun you got it

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
chadhietala: I think this is fine. At this point the `this.name` is basically a CONST.

packages/glimmer-runtime/lib/partial.ts
```diff
@@ -0,0 +1,12 @@
+import { SerializedTemplate } from 'glimmer-wire-format';
+
+export class PartialDefinition {
+  name: string;
+  serializedTemplate: SerializedTemplate;
```
chancancode: You can just call this `template` since we have types

packages/glimmer-runtime/lib/scanner.ts
```diff
@@ -27,6 +28,13 @@ export default class Scanner {
     });
   }
 
+  scanInlineBlock(symbolTable: SymbolTable): InlineBlock {
```
chancancode: This should probably be called `scanPartial`

packages/glimmer-runtime/lib/syntax/builtins/partial.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  prettyPrint() {
+    return `partial ${this.args.prettyPrint()}`;
+  }
+
+  compile(compiler: CompileInto & SymbolLookup, env: Environment) {
+
+    /*
+    //        Enter(BEGIN, END)
```
chancancode: This needs to be updated (and doesn't need the block comment syntax)

packages/glimmer-runtime/lib/syntax/builtins/partial.ts
```diff
@@ -0,0 +1,79 @@
...
+    //        Exit
+    */
+
+    assert(this.args.positional.values.length > 0, `Partial found with no arguments. You must specify a template.`);
+    assert(this.args.positional.values.length < 2, `Partial found with more than one argument. You can only specify a single template.`);
+
+    let compiledPartialNameExpression = this.args.positional.values[0].compile(compiler, env);
+
+    let BEGIN = new LabelOpcode({ label: "BEGIN" });
+    let ELSE = new LabelOpcode({ label: "ELSE" });
```
chancancode: This isn't used

packages/glimmer-runtime/lib/partial.ts
mmun: I like this reasoning :+1:

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -681,6 +708,10 @@ export class TestEnvironment extends Environment {
     return modifier;
   }
 
+  compileSpec(template: string) {
+    return compileRealSpec(template, { env: this });
```
chancancode: This can be inlined into `registerPartial` (it doesn't need to be its own method)

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+    return {
+      guid: this._guid,
+      type: this.type,
+      args: [this.name.toJSON()]
+    };
+  }
+}
+
+
+export class EvaluateUpdatingPartialOpcode extends UpdatingOpcode {
```
chancancode: We don't need a specialized opcode here. We already have [`Assert`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/vm.ts#L392-L429) that does exactly this. You just give it a [`ReferenceCache`](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-reference/lib/validators.ts#L275-L324) and it will take care of the rest.

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+    vm.invokeBlock(block, vm.frame.getArgs());
+
+    if (!isConst(reference)) {
+      vm.pushDynamicScope();
+
+      vm.updateWith(new EvaluateUpdatingPartialOpcode({
+        compiler: this.compiler,
+        nameReference: reference,
+        name
+      }));
```
chancancode: We can just use `Assert` here. See the comment below and you can copy the code from https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/compiled/opcodes/component.ts#L84-L86

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+  }
+
+  evaluate(vm: VM) {
+    let reference: PathReference<any> = this.name.evaluate(vm);
+    let name: InternedString = reference.value();
+
+    let { serializedTemplate } = vm.env.lookupPartial([name]);
+    let scanner = new Scanner(serializedTemplate, vm.env);
+    let block = scanner.scanInlineBlock(this.compiler.getSymbolTable());
+    vm.invokeBlock(block, vm.frame.getArgs());
```
chancancode: I don't think a partial can have args?

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
 
+export class EvaluatePartialOpcode extends Opcode {
+  public type = "evaluate-partial";
+  public compiler: CompileInto & SymbolLookup;
+  public name: CompiledExpression<any>;
+
+  constructor({ name, compiler }: { compiler: CompileInto & SymbolLookup, name: CompiledExpression<any> }) {
+    super();
+    this.name = name;
+    this.compiler = compiler;
```
chancancode: Can you pass the symbol table instead of the compiler?

mmun: The symbol table is a private field of the compiler.

mmun: It seems like you be calling `this.compiler.compileInlineBlock()` instead of doing the scanning business in `evaluate`

chancancode: You already added `this.compiler.getSymbolTable()`, and the only place `this.compiler` is used in this class is to call that method, so it seems good that you just pass the symbol table from the caller instead of doing it in here.

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
mmun: Confirm they can not.

chancancode: You can pass `EvaluatedArgs.empty()`

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
mmun: I didn't add it :P I don't think exposing the symbol table is good...

packages/glimmer-test-helpers/lib/environment.ts
mmun: The "spec" name just needs to 🔥 🔥

packages/glimmer-runtime/lib/compiled/blocks.ts
```diff
@@ -61,6 +61,14 @@ export class InlineBlock extends Block {
     let ops = new InlineBlockCompiler(this, env).compile();
     return this.compiled = new CompiledBlock(ops, this.symbolTable.size);
   }
+
+  initBlocks(blocks = this['children'], parentTable = this['symbolTable']): this {
```
wycats: I don't think you should overload `InlineBlock`; instead, make a new top level `PartialBlock` that does the set of things you're trying to do here.

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+  }
+
+  evaluate(vm: VM) {
+    let reference: PathReference<any> = this.name.evaluate(vm);
```
chancancode: Don't need the type annotation here. TypeScript can infer it.

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+    let name: InternedString = reference.value();
```
chancancode: ditto

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
 }
 
+export class EvaluatePartialOpcode extends Opcode {
+  public type = "evaluate-partial";
+  public compiler: CompileInto & SymbolLookup;
+  public name: CompiledExpression<any>;
+
+  constructor({ name, compiler }: { compiler: CompileInto & SymbolLookup, name: CompiledExpression<any> }) {
+    super();
+    this.name = name;
```
chancancode: You shouldn't need to take the name – per your compilation, the name should already be evaluated and sitting on the args register at this point

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
chancancode: Actually. Nvm. What you have discovered is a legitimate TypeError. See my comment on the Syntax compilation side

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -306,6 +313,77 @@ export class EvaluateOpcode extends Opcode {
...
+    this.compiler = compiler;
+  }
+
+  evaluate(vm: VM) {
+    let reference: PathReference<any> = this.name.evaluate(vm);
+    let name: InternedString = reference.value();
+
+    let { serializedTemplate } = vm.env.lookupPartial([name]);
+    let scanner = new Scanner(serializedTemplate, vm.env);
+    let block = scanner.scanInlineBlock(this.compiler.getSymbolTable());
```
chancancode: It's pretty common for a particular partial callsite to flip-flop between two (or more) partials:

```hbs
{{#each model as |post|}}
  {{partial post.type}}
{{/each}}
```

Here, the partial name might go from "featured", "post", "story", "post", "quote", ... The compilation is expensive, so we shouldn't do it more than once per unique template; we should add a cache here:

```js
export class EvaluatePartialOpcode extends Opcode {
  private cache = new dict<Partial>();

  evaluate(vm: VM) {
    ...
    
    if (!this.cache[name]) {
      // do compilation
    }

    ...
}
```

This is similar to a [polymorphic inline cache](https://en.wikipedia.org/wiki/Inline_caching#Polymorphic_inline_caching).

packages/glimmer-runtime/lib/syntax/builtins/partial.ts
```diff
@@ -0,0 +1,79 @@
...
+    let END = new LabelOpcode({ label: "END" });
+
+    compiler.append(new EnterOpcode({ begin: BEGIN, end: END }));
+    compiler.append(BEGIN);
+    compiler.append(new PutArgsOpcode({ args: this.args.compile(compiler, env) }));
+    compiler.append(new TestOpcode());
```
chancancode: There is a missing step here. I assume this is trying to test that if the partial name is falsy, don't render anything. However, the env can fail to lookup the partial for other reasons. So I think `PutArgs` is fine, but you should insert another opcode here that converts the name partial into a `PartialDefinition` before running the `TestOpcode`?

packages/glimmer-runtime/lib/compiled/opcodes/vm.ts
mmun: This is a cool idea!

packages/glimmer-runtime/lib/scanner.ts
mmun: There's nothing really Partial specific? I feel like a proper factoring of the scanner would have a "scan this block relative to this symbol table". But whatever... call it `scanPartial` ;P

asakusuma:
Added some more tests to cover some of the cases mentioned in the feedback

packages/glimmer-runtime/tests/ember-component-test.ts
```diff
@@ -948,8 +948,8 @@ module("Components - generic - attrs");
 module("Components - integration - scope");
 
 testComponent('correct scope - conflicting local names', {
-  layout: stripTight`{{#with @a as |item|}}{{@a}}: {{item}}, {{#with @b as |item|}}
-                     {{@b}}: {{item}}, {{#with @c as |item|}}{{@c}}: {{item}}{{/with}}{{/with}}{{/with}}`,
+  layout: stripTight`{{#if @a as |item|}}{{@a}}: {{item}}, {{#if @b as |item|}}
+                     {{@b}}: {{item}}, {{#if @c as |item|}}{{@c}}: {{item}}{{/if}}{{/if}}{{/if}}`,
   invokeAs: { props: { a: '"A"', b: '"B"', c: '"C"' } },
```
mmun: this should be reverted

asakusuma: Done

PR #176, opened 2016-05-20
Ensure partial re-renders are stable

asakusuma:
Addresses https://github.com/tildeio/glimmer/issues/174

Next, I'll refactor so that you get the stable DOM test for free with `rerender()`, which will address https://github.com/tildeio/glimmer/issues/175

Tested the changes in Ember and confirmed that they make the partial tests pass.

cc @mmun

### DIFF
--- a/packages/glimmer-runtime/tests/partial-test.ts
+++ b/packages/glimmer-runtime/tests/partial-test.ts
@@ -1,5 +1,13 @@
 import { Template, RenderResult } from "glimmer-runtime";
-import { BasicComponent, TestEnvironment, TestDynamicScope, equalTokens, strip } from "glimmer-test-helpers";
+import {
+  BasicComponent,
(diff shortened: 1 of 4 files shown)


packages/glimmer-runtime/tests/partial-test.ts
```diff
@@ -189,7 +204,7 @@ QUnit.test('dynamic partial with local reference', assert => {
 
   equalTokens(root, `You smaht. You loyal. `);
   rerender({ name: 'test', qualities: ['smaht', 'loyal'] });
-  equalTokens(root, `You smaht. You loyal. `);
+  equalTokens(root, `You smaht. You loyal. `, true);
```
mmun: probably intended this to be on the previous line?

mmun:
Woohoo!

PR #177, opened 2016-05-23
Eat falsy component definitions and invalidate if their ref changes

zackthehuman:
If the `definition` of a component resolves to `undefined`/`null` or another falsy value it should be a signal that there is no component to render. In that case, we can just wait for the value to change since there may be a component definition in the future.

This change makes it possible to implement the behavior tested in https://github.com/emberjs/ember.js/blob/b418b95ea3d31e3a3656906e6ad784422f648c31/packages/ember-glimmer/tests/integration/components/dynamic-components-test.js#L489 where there is a definition initially, but then the value changes to `undefined` and there is no longer a definition. The only requirement here is that the thing providing the value for the definition returns a falsy value. Currently in Ember it will just `throw` in this case, but returning `null` makes this work.

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/component.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/component.ts
@@ -59,33 +59,39 @@ export class OpenDynamicComponentOpcode extends Opcode {
       definition = cache.peek();
     }
 
-    let manager = definition.manager;
-    let component = manager.create(definition, args, dynamicScope);
-    let selfRef = manager.getSelf(component);
-    let destructor = manager.getDestructor(component);
+    if(definition) {
+      let manager = definition.manager;
+      let component = manager.create(definition, args, dynamicScope);
+      let selfRef = manager.getSelf(component);
+      let destructor = manager.getDestructor(component);
 
-    let callerScope = vm.scope();
+      let callerScope = vm.scope();
 
-    // pass through the list of outer attributes to shadow from the
-    // invocation site, as well as the component definition as internal
-    // arguments.
-    args = args.withInternal();
-    args.internal['shadow'] = shadow;
-    args.internal['definition'] = definition;
-    args.internal['component'] = component;
+      // pass through the list of outer attributes to shadow from the
+      // invocation site, as well as the component definition as internal
+      // arguments.
+      args = args.withInternal();
+      args.internal['shadow'] = shadow;
+      args.internal['definition'] = definition;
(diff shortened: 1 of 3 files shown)


packages/glimmer-runtime/lib/compiled/opcodes/component.ts
```diff
@@ -59,33 +59,37 @@ export class OpenDynamicComponentOpcode extends Opcode {
...
+      vm.invokeLayout({ templates, args, shadow, layout, callerScope });
+      vm.env.didCreate(component, manager);
+
+      if (!isConst(definitionRef)) {
+        vm.updateWith(new Assert(cache));
+      }
+
+      vm.updateWith(new UpdateComponentOpcode({ name: definition.name, component, manager, args, dynamicScope }));
+    } else {
       vm.updateWith(new Assert(cache));
```
zackthehuman: I'm a little worried that I should be doing some kind of cleanup here since a dynamicScope is created but I never register a destructor. How does that new scope get cleaned up (does it even have to be?)

chadhietala:
LGTM

krisselden:
Can you enable that test? Or add a regression test?

zackthehuman:
@krisselden Can't enable the test yet because I need to make a change in Ember to return `null` when you lookup components that there is no definition for. I have the change locally but need to get this in first so I can bump glimmer and commit as one unit.

zackthehuman:
I will look in to adding a test for this in Glimmer.

zackthehuman:
@krisselden Added a test for this specific use case where a component ref goes rogue.

packages/glimmer-runtime/tests/ember-component-test.ts
```diff
@@ -871,6 +871,59 @@ testComponent('parameterized has-block-params (concatted attr, default) when blo
   expected: '<button data-has-block-params="is-false"></button>'
 });
 
+module("Components - curlies - dynamic component");
+
+QUnit.test('missing definition', assert => {
```
krisselden: Can you also test it being initially false and becoming a component?  Also, the test implies that any lookup failures should be silent and just render blank.  Also, why is it rendering a comment?

zackthehuman: Added tests for both cases. I'm not sure why it's rendering a comment. I can dig in and see. It makes sense that it puts _something_ in the DOM to keep track of where it goes, but I'm not sure why it's a comment.

PR #185, opened 2016-06-13
[WIP] Dynamic component currying support

zackthehuman:
This adds support for dynamic components + currying of their args. Still a work in progress. Note: the actual currying and stuff doesn't happen here -- it would happen in Ember. This just adds machinery to support it without doing weird things.

The basic idea here is the separate the component definition from any args that need to be curried when instantiating it. This means that places that use to return a `ComponentDefinition` now need to return something like `{ definition: ComponentDefinition, args: EvaluatedArgs[] }`.

This applies only for dynamic components.

This also adds a new opcode for testing the property of a reference, called `TestPropOpcode`. This can be used to observe if/when the `definition` reference's value changes between truthy and falsy values (for cases where the component reference "disappears" and appears later as well as the converse).

The `TestPropOpcode` doesn't use Emberish truthiness -- it relies on Glimmerish truthiness.

### DIFF
--- a/packages/glimmer-runtime/lib/compiler.ts
+++ b/packages/glimmer-runtime/lib/compiler.ts
@@ -23,7 +23,8 @@ import {
   PutArgsOpcode,
   PutValueOpcode,
   JumpUnlessOpcode,
-  TestOpcode
+  TestOpcode,
+  TestPropOpcode
 } from './compiled/opcodes/vm';
 
 import * as Syntax from './syntax/core';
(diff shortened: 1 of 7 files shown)


packages/glimmer-runtime/lib/compiler.ts
```diff
@@ -426,7 +427,7 @@ class ComponentBuilder {
     compiler.append(BEGIN);
     compiler.append(new PutArgsOpcode({ args: definitionArgs }));
     compiler.append(new PutValueOpcode({ expression: definition }));
-    compiler.append(new TestOpcode());
+    compiler.append(new TestPropOpcode("definition" as FIXME<'intern'>));
```
chadhietala: I believe we need this in Ember for `disconnectOutlet` as well.

zackthehuman:
Closing since we solved this another way in Ember.

PR #187, opened 2016-06-14
Refactor Attributes And Component Arguments

chadhietala:
This removes the syntax around `DynamicProperty` and the notion that quotes have meaning in terms of how values are set. The logic around deciding to use `setAttribute` and property setting is pretty much the same as HTMLBars, however we have introduced the concept of attribute/property change lists.

### Change Lists
A change list is a 2 method POJO that knows how to do the actual setting and updating of an attribute or property on an element. In Glimmer we take a similar strategy to jQuery's `attr()`, in that we prefer `setAttribute` over property slotting. Because of this we have both Attribute and Property change lists.

By default Glimmer comes with a set of change lists that make sense for smoothing over browser inconsistencies and security. That being said we expose the `attributeFor` hook in the `Environment` where all attributes funnel through. This allows for triaging user-land cases. A change list is used by the `Attribute` class which is aware of when setting and updating needs to occur.

### Arguments and Attributes
Instead of having nuance in calling semantics, this PR introduces the notion of `Argument`s (denoted by `@`) for setting and reading. This looks like the following:

```handlebars
<foo-bar @bar=baz foo="foo" />
```

```handlebars
<div>{{@bar}}</div>
```

In above example `foo` attribute would be shadowed onto the top level element; the `div`, where as `@bar` is set and referenced by semantics that feel complimentary. This draws very clear lines in not only HTML attributes and arguments for the component, but also internal and external references.

Supersedes #197


chadhietala:
@krisselden @wycats can I get some eyes on the general approach before I dig into the IE issues.

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -351,12 +351,21 @@ abstract class ElementPatchOperation<V> implements ElementOperation {
   abstract toJSON(): Dict<string>;
 }
 
+export const STYLE_WARNING = `Binding style attributes may introduce cross-site scripting vulnerabilities; please ensure that values being bound are properly escaped.`
+
+function styleWarning(name) {
+  if (name === 'style') {
+    console.warn(STYLE_WARNING);
```
rwjblue: We will need to be able to provide the environment with the mechanism for warning I think...

chadhietala: Confirm. This isn't really acceptable.

packages/glimmer-runtime/lib/builder.ts
```diff
@@ -89,13 +91,19 @@ class GroupedElementOperations implements ElementOperations {
...
+    }
     if (this.element.tagName === 'INPUT' && name === 'value') {
       this.group.push(new EmptyStringResetingProperty(this.element, name, reference));
     }
     this.group.push(new NonNamespacedAttribute(this.element, name, reference));
   }
 
   addAttributeNS(namespace: InternedString, name: InternedString, reference: PathReference<string>) {
+    if (name === 'style') {
+      this.env.warn(`${name}-attribute`);
```
krisselden: If name === style why format it?

chadhietala: I want some type of API here that allows us to disambiguate on the calling side or I can just special case the method and call it a day.

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -1173,7 +1173,7 @@ test("non-standard namespaced attribute nodes w/ concat follow the normal dirtyi
   equalTokens(root, "<div epub:type='dedication backmatter'>hello</div>", "After reset");
 });
 
-test("property nodes follow the normal dirtying rules", function() {
+QUnit.skip("property nodes follow the normal dirtying rules", function() {
```
chancancode: feel free to ✂️

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -575,7 +555,7 @@ export class DynamicPropOpcode extends Opcode {
   evaluate(vm: VM) {
```
chancancode: This class needs to 🔥

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -164,13 +165,14 @@ export class CloseElementOpcode extends Opcode {
     let className = classList.toReference();
 
     if (isConstReference(className)) {
-      NonNamespacedAttribute.install(dom, element, 'class' as InternedString, className.value());
+      Attribute.setAttribute(dom, element, 'class' as InternedString, className.value());
```
chancancode: Shouldn't this also go through `attributeFor`?

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -790,3 +791,45 @@ test("Namespaced attribute with unquoted expression throws", function() {
...
+QUnit.module('Style attributes', {
+  setup() {
+    commonSetup();
+    orignalAttributeFor = env.attributeFor;
+    env.attributeFor = function(element, attr, reference) {
+      if (attr === 'style') {
+        return new ElementOperation(element, StyleAttribute, attr, reference as Reference<string>);
+      }
+      return orignalAttributeFor.call(env, element, attr, reference);
+    }
```
chancancode: Should subclass here. You can make `commonSetup` optionally take an environment instance (`commonSetup(env = new TestEnvironment()) { ... }`)

chadhietala: Done

packages/glimmer-runtime/lib/dom/props.ts
```diff
@@ -0,0 +1,77 @@
...
+// * strange spec outlier
+const ATTR_OVERRIDES = {
+
+  // phantomjs < 2.0 lets you set it as a prop but won't reflect it
+  // back to the attribute. button.getAttribute('type') === null
+  BUTTON: { type: true, form: true },
+
+  INPUT: {
+    // TODO: remove when IE8 is droped
+    // Some versions of IE (IE8) throw an exception when setting
```
mixonic: IE8 is unsupported. IE9 is the minimum version afaik (and is definitely the lowest tested version of IE)

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
chadhietala: This is gone now.

PR #188, opened 2016-06-15
Support for positional params in EmberishCurlyComponent for test environment

zackthehuman:
This is more or less prerequisite work for tests involving closure components since they'll need to curry named and positional args. It didn't occur to me that positional args weren't implemented in EmberishCurlyComponents.

### DIFF
--- a/packages/glimmer-test-helpers/lib/environment.ts
+++ b/packages/glimmer-test-helpers/lib/environment.ts
@@ -37,6 +37,8 @@ import {
 
   // Values
   EvaluatedArgs,
+  EvaluatedNamedArgs,
+  EvaluatedPositionalArgs,
 
   // Syntax Classes
   StatementSyntax,
@@ -83,7 +85,8 @@ import {
   OpaqueIterable,
   AbstractIterable,
   IterationItem,
-  isConst
+  isConst,
+  combine
 } from "glimmer-reference";
 
 import {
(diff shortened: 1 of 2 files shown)


packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -232,9 +235,13 @@ export class EmberishCurlyComponent extends GlimmerObject {
   public attrs: Attrs;
   public element: Element;
   public parentView: Component = null;
+  public args: ProcessedArgs;
+  public argsRevision: number;
 
   static create(args: { attrs: Attrs }): EmberishCurlyComponent {
-    return super.create(args) as EmberishCurlyComponent;
+    let result = super.create(args) as EmberishCurlyComponent;
+    result.argsRevision = result.args && result.args.tag && result.args.tag.value();
```
chancancode: Where is this used?

zackthehuman: Doh. I meant to use it inside of the component update but we don't really have to since we compute the value for each update (unlike in Ember where we only recompile if the tag revision changes). Will remove.

chancancode: We also have another problem, when we did the work to add tags, we didn't bother to implement it correctly in glimmer-objects since they are "only used in tests", so pretty much EVERYTHING in the test environment is volatile 😞 😞 😞 😞 😞

So even if you add this here chances are that it would probably just always invalidate 😐 This is actually quite problematic, because we use the same setup in benchmarks. It just so happens that most of the benchmarks we are doing wouldn't be able to take advantage of those optimizations anyway, but still...

zackthehuman: Yeah, about half way in to writing this code I realized that it's basically all recomputed all the time.

PR #189, opened 2016-06-15
Adds a CommonJS build

tomdale:
This PR adds a CommonJS build for consumption by Node tools (like, say, a Broccoli plugin for compiling templates).

There are a couple weird things here that could use explanation.

First, rather than building into `dist/cjs/` (like `htmlbars` does), this puts the CJS files inside `dist/node_modules/`. This is because inside Glimmer we require across packages via top-level names rather than relative paths. For example, you might see something like `import { preprocess } from "glimmer-syntax";` instead of `import { preprocess } from "../../glimmer-syntax";`.

Node doesn't handle this case if we put it in a `cjs` directory, because it only looks for non-relative packages inside directories called `node_modules`. We should either publish each Glimmer package as a separate npm package, or maybe just use relative module names.

The other unusual thing is that Glimmer relies on SimpleHTMLTokenizer, which ships as either ES6 or an AMD-ish file (which does indeed support CommonJS consumption as well). When compiled from ES6 to AMD-ish, though, everything gets concatenated into a single file. Glimmer imports just pieces of SHT, like `import EventedTokenizer from "simple-html-tokenizer/evented-tokenizer"`; when compiled to CJS by TypeScript, that becomes `require('simple-html-tokenizer/evented-tokenizer')`. In order to support these modules in Node, we need to maintain separate files rather than concatenating everything together.

To that end, this PR uses TypeScript as a poor man's ES6-to-CommonJS compiler, converting from ES6 modules to CommonJS modules and colocating them with Glimmer's CommonJS modules in `dist/node_modules`.

This PR also removes the duplicate `typescript` dev dependency, and adds `esnext:main` and `main` entries to `package.json`.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -1,4 +1,4 @@
-/* globals __dirname */
+/*jshint node:true*/
 
 var path = require('path');
 var existsSync = require('exists-sync');
@@ -7,10 +7,10 @@ var merge = require('broccoli-merge-trees');
 var typescript = require('broccoli-typescript-compiler');
 var transpileES6 = require('emberjs-build/lib/utils/transpile-es6');
 var handlebarsInlinedTrees = require('./build-support/handlebars-inliner');
-var getVersion = require('git-repo-version');
 var stew = require('broccoli-stew');
 var mv = stew.mv;
 var find = stew.find;
+var rename = stew.rename;
 
 function transpile(tree, label) {
   return transpileES6(tree, label, { sourceMaps: 'inline' });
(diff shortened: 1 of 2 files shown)


ember-cli-build.js
```diff
@@ -70,12 +73,42 @@ module.exports = function() {
...
+  /*
+   * ES5 Named AMD Build
+   */
   libTree = transpile(libTree, 'ES5 Lib Tree');
   var es5LibTree = mv(libTree, 'named-amd');
 
+  /*
+   * CommonJS Build
+   */
+  tsOptions.tsconfig.compilerOptions.module = "commonjs";
```
rwjblue: Are these objects being deeply cloned when we pass them to `typescript`?  I am concerned that sharing the object by ref here would lead to commonjs being used across the board...

tomdale:
@rwjblue It seemed to work correctly, but I agree that creating a new copy each time is more robust. Updated in most recent commit.

PR #190, opened 2016-06-17
Include TypeScript source in npm package

tomdale:
This is useful for TypeScript apps that want to consume Glimmer directly

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,9 @@
+bower_components
+/node_modules
+tmp
+npm-debug.log
+sauce-example.log
+
+testem.log
+
+.vscode
--- a/package.json
+++ b/package.json
@@ -24,9 +24,6 @@
   "author": "Tilde, Inc.",
   "license": "MIT",
   "readmeFilename": "README.md",
-  "files": [
-    "dist"
-  ],
   "dependencies": {
     "broccoli-concat": "^2.1.0",
     "broccoli-funnel": "^1.0.1",


rwjblue:
Actually, I believe that this is not quite correct for consuming in an application without that app having to manually build via `ember-cli-build.js` (which isn't included in in `files` either).  

There is a bit of work going on in the build to make a `handlebars` package (see [here](https://github.com/tildeio/glimmer/blob/master/ember-cli-build.js#L85) and [here](https://github.com/tildeio/glimmer/blob/master/build-support/handlebars-inliner.js)), and the rest of the app assumes it can import these from `'handlebars/*'` paths (i.e. [here](https://github.com/tildeio/glimmer/blob/078009646b6838474aa2268f332721570153db8e/packages/glimmer-syntax/lib/parser.ts#L1)).  We also assume that you can import directly from `simple-html-tokenizer` in a few places...

tomdale:
@rwjblue I see your point. For my particular use case, I'm actually requiring Glimmer's `ember-cli-build.js` file for the build (that's another file that should probably be in the `files` array), and only using the TypeScript files directly for the type annotations. This is actually modeled on how TypeScript itself is set up.

Still, it seems kind of silly to "build" the TypeScript stuff into `dist/`. Ideally the weirdness with `handlebars/*` and `simple-html-tokenizer` can be addressed more cleanly than by patching it in in a few different, non-standard ways. I think the best bet, ideally, is to consume everything as ES6 modules and use TypeScript's `paths` feature (only on TS nightly at the moment) to tell the compiler where to find all the dependencies.

tomdale:
@rwjblue Given https://github.com/emberjs/ember.js/pull/13615 and the fact that Ember relies on including the built-in Glimmer `ember-cli-build.js`, I think we should probably remove the `files` entry from `package.json` entirely. At the moment, Ember relies on being able to consume the entire package, and do its own build, from npm.

rwjblue:
Now this will make it exclude `dist/`, which makes CJS not work. Can you add an `.npmignore` that is roughly a copy of `.gitignore` but not excluding `dist/`?

rwjblue:
LGTM

asakusuma:
@tomdale I think you'll need to rebase

tomdale:
Looks like the build is failing, not sure why yet.

rwjblue:
@tomdale - @asakusuma fixed (by pinning typescript version) last night

rwjblue:
YOLO

PR #191, opened 2016-06-20
Properly handle updating value to null in MS browsers

asakusuma:
- [x] Ensure test fails before fix
- [x] Apply fix

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
@@ -478,6 +478,17 @@ export class Property extends ElementPatchOperation<Opaque> {
   }
 }
 
+export class EmptyStringResetingProperty extends Property {
+  protected update(dom: DOMHelper, element: Element, value: Opaque) {
+    // Needed for cases where removeAttribute() does not work
+    if (value === null) {
+      value = '';
+    }
+
+    super.update(dom, element, value);
+  }
+}
+
 function formatElement(element: Element): string {
   return JSON.stringify(`<${element.tagName.toLowerCase()} />`);
 }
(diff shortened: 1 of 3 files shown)


rwjblue:
I'm ok with this as a way to unblock many tests in Ember, but it does need to be revisited and thought through. 

I will try to schedule some time with @wycats / @chancancode / @krisselden / @mmun to make a "grand plan".

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -463,6 +463,13 @@ export class Property extends ElementPatchOperation<Opaque> {
 
   protected update(dom: DOMHelper, element: Element, value: Opaque) {
     let { name } = this;
+
+    // Needed for compatibility with Microsoft browsers
+    // Copied from https://github.com/tildeio/htmlbars/blob/ec541e79dbf0b03471070a0de3ac810f8fc924ed/packages/dom-helper/lib/main.js#L309-L319
+    if (value === null && (name === 'value' || name === 'type' || name === 'src')) {
```
rwjblue: This makes me somewhat sad.  Can you create a subclass of `Property` for this (and use it in `addAttribute` as you have)?

```js
class EmptyStringResetingProperty extend Property {
  protected update(dom: DOMHelper, element: Element, value: Opaque) {
    if (value === null) {
      value = '';
    }
    
    super.update(dom, value);
  }
}
```

asakusuma:
Can someone kick the build. I think saucelabs is just timing out.

rwjblue:
Restarted!

rwjblue:
This is good for now and unblocks many things in Ember itself, but I am still planning on following up with @wycats / @chancancode / @krisselden on a better final solution.

PR #193, opened 2016-06-21
Fix typescript errors

asakusuma:
Issue repro: https://github.com/tildeio/glimmer/pull/192

I'm guessing this is not the long term solution and we should figure out the underlying cause of how this issue got into the build, but this will at least unblock the build.

cc @rwjblue @krisselden @mmun


rwjblue:
Can we pin typescript to the last known working version instead? We can then unpin it when we have fixed this properly (or TS has reverted their bug 🙃)...

rwjblue:
Thank you!

PR #195, opened 2016-06-21
Fix how input handles initial null attributes

asakusuma:
Test and fix IE9 issue around ember input https://github.com/emberjs/ember.js/pull/13615

- [x] Ensure test fails without fix
- [x] Introduce fix that makes test pass


asakusuma:
@rwjblue I verified that this does in fact fix my ember glimmer PR https://travis-ci.org/emberjs/ember.js/jobs/139366684

PR #196, opened 2016-06-22
Fix package name in bower.json

tricknotes:
In favor of the name in [package.json](https://github.com/tildeio/glimmer/blob/d4d1e3a86/package.json#L2).


rwjblue:
👍

PR #197, opened 2016-06-22
Create sanitized attribute and property classes

zackthehuman:
This creates "sanitized" subclasses of `Property` and `NonNamespacedAttribute` which enforce sanitization of known dangerous tag/attribute combinations. This also makes https://github.com/emberjs/ember.js/blob/1be0354068d933065ac542f49d42d73409366a47/packages/ember-glimmer/tests/integration/components/attribute-bindings-test.js#L456 and https://github.com/emberjs/ember.js/blob/3cfc9614bbd92ee8e80b6c4894b024f2e14bffae/packages/ember-glimmer/tests/integration/helpers/unbound-test.js#L98 pass.

### DIFF
--- a/packages/glimmer-runtime/lib/builder.ts
+++ b/packages/glimmer-runtime/lib/builder.ts
@@ -94,12 +94,13 @@ class GroupedElementOperations implements ElementOperations {
     let { element } = this;
     let { tagName } = element;
 
-    if (SanitizedProperty.requiresSanitization(tagName, name)) {
+    if (SanitizedNonNamespacedAttribute.requiresSanitization(tagName, name)) {
       this.group.push(new SanitizedNonNamespacedAttribute(element, name, reference));
     } else if (tagName === 'INPUT' && name === 'value') {
       this.group.push(new EmptyStringResetingProperty(element, name, reference));
+    } else {
+      this.group.push(new NonNamespacedAttribute(element, name, reference));
     }
-    this.group.push(new NonNamespacedAttribute(element, name, reference));
   }
 
   addAttributeNS(namespace: InternedString, name: InternedString, reference: PathReference<string>) {
--- a/packages/glimmer-runtime/lib/builder.ts
+++ b/packages/glimmer-runtime/lib/builder.ts
@@ -15,7 +15,9 @@ import {
   EmptyStringResetingProperty,
   NamespacedAttribute,
   NonNamespacedAttribute,
-  Property
+  Property,
+  SanitizedProperty,
+  SanitizedNonNamespacedAttribute
 } from './compiled/opcodes/dom';
 
 interface FirstNode {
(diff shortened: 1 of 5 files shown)


packages/glimmer-runtime/lib/builder.ts
```diff
@@ -89,18 +91,30 @@ class GroupedElementOperations implements ElementOperations {
   }
 
   addAttribute(name: InternedString, reference: PathReference<string>) {
-    if (this.element.tagName === 'INPUT' && name === 'value') {
-      this.group.push(new EmptyStringResetingProperty(this.element, name, reference));
+    let { element } = this;
+    let { tagName } = element;
+
+    if (SanitizedProperty.requiresSanitization(tagName, name)) {
```
rwjblue: Should this use the method off of SanitizedNonNamespacedAttribute instead?

packages/glimmer-runtime/lib/sanitize-attribute-value.ts
```diff
@@ -0,0 +1,66 @@
+import { FIXME, InternedString, Opaque } from 'glimmer-util';
+import { SafeString, isSafeString } from './upsert';
+import { DOMHelper } from './dom';
+
+const badProtocols = {
```
rwjblue: Lets store these various constants as arrays, not pojos.

packages/glimmer-runtime/lib/builder.ts
zackthehuman: Yes, this is the result of copy pasta. Will fix.

packages/glimmer-runtime/lib/sanitize-attribute-value.ts
zackthehuman: And use `indexOf` instead of a lookup?

rwjblue: Yes

packages/glimmer-runtime/lib/builder.ts
```diff
@@ -89,18 +91,31 @@ class GroupedElementOperations implements ElementOperations {
...
+    let { element } = this;
+    let { tagName } = element;
+
+    if (SanitizedNonNamespacedAttribute.requiresSanitization(tagName, name)) {
+      this.group.push(new SanitizedNonNamespacedAttribute(element, name, reference));
+    } else if (tagName === 'INPUT' && name === 'value') {
+      this.group.push(new EmptyStringResetingProperty(element, name, reference));
+    } else {
+      this.group.push(new NonNamespacedAttribute(element, name, reference));
     }
```
zackthehuman: I changed the conditions here. This used to push two objects in the case of `input[value]`. Was this the correct behavior? For the sanitized case it should be one or the other, not both.

asakusuma: Interesting, I would have thought the tests would fail if you did this but apparently not. I had assumed you needed both. I added a specific test case around this though, so I'm not too worried about this change breaking things.

rwjblue: Good catch!

krisselden:
This was superseded

PR #201, opened 2016-06-27
Propagate moduleName to statement

asakusuma:
- Needed for local lookup
- Supports moduleName property for compile options
- Add moduleName to statement on runtime template re-inflatation
- Tested against ember.js local lookup tests

cc @krisselden @chadhietala @wycats 

Would like some validation that this is the right approach before I add tests to glimmer itself


asakusuma:
Added a test

PR #207, opened 2016-07-03
Pass block meta info into helper lookup

asakusuma:
Enables local lookup for helpers

cc @krisselden @rwjblue


chadhietala:
I think CI just flaked here.

asakusuma:
@chadhietala no typescript check failed. 1) `compiler` is defined as some other type that does not define `symbolTable` 2) even if `compiler` was a `Compiler`, `symbolTable` is private. Need to figure out a better way to expose the meta.

asakusuma:
Got things green

PR #213, opened 2016-07-12
Propagate curly trustworthiness

chadhietala:
We need to know the calling side of the ref e.g. either double or triple
curlies.

To keep the wire-format as small as possible we actually do want a `TrustingAttr` type. This type is just a  delegate to `DynamicAttr`, but slots the `isTrusting` field as "true". This is a tradeoff that I think makes sense as we pay the cost of an additional type, but we don't bloat the wire-format with bools for each attribute's trustworthiness. I suspect that cost of the additional type in Glimmer outweighs the amount of dynamic attributes used in a real application.

### DIFF
--- a/packages/glimmer-compiler/lib/javascript-compiler.ts
+++ b/packages/glimmer-compiler/lib/javascript-compiler.ts
@@ -143,9 +143,14 @@ export default class JavaScriptCompiler {
     this.push(['staticAttr', name, value, namespace]);
   }
 
-  dynamicAttr(name: string, namespace: string) {
+  dynamicAttr(name: str, namespace: str) {
     let value = this.popValue<Expression>();
     this.push(['dynamicAttr', name, value, namespace]);
+  }
+
+  trustingAttr(name: str, namespace: str) {
+    let value = this.popValue<Expression>();
+    this.push(['trustingAttr', name, value, namespace]);
   }
 
   staticArg(name: str) {
--- a/packages/glimmer-compiler/lib/template-compiler.ts
+++ b/packages/glimmer-compiler/lib/template-compiler.ts
@@ -10,6 +10,10 @@ export interface CompileOptions {
   moduleName?: string;
 }
 
+function isTrustedValue(value) {
+  return value.escaped !== undefined && !value.escaped;
+}
+
 export default class TemplateCompiler {
   static compile(options: CompileOptions, ast) {
     let templateVisitor = new TemplateVisitor();
@@ -99,8 +103,12 @@ export default class TemplateCompiler {
         this.opcode('dynamicArg', action, name);
       }
     } else {
+      let isTrusting = isTrustedValue(value);
+
       if (isStatic) {
         this.opcode('staticAttr', action, name, namespace);
+      } else if (isTrusting) {
+        this.opcode('trustingAttr', action, name, namespace);
       } else if (action.value.type === 'MustacheStatement') {
         this.opcode('dynamicAttr', action, name);
       } else {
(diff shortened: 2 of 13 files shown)


packages/glimmer-compiler/lib/javascript-compiler.ts
```diff
@@ -148,6 +148,11 @@ export default class JavaScriptCompiler {
     this.push(['dynamicAttr', name, value, namespace]);
   }
 
+  trustedAttr(name: string, namespace: string) {
```
chancancode: shouldn't this be a parameter to `dynamicAttr` (like it is for append)?

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -186,20 +186,22 @@ export class StaticAttrOpcode extends Opcode {
   public namespace: InternedString;
   public name: InternedString;
   public value: ValueReference<string>;
+  public isTrusted: boolean;
```
chancancode: This is not a thing for static right?

chancancode:
Pretty sure most of these names should be "trusting" now "trusted". In Glimmer "trusting" means it doesn't care/do extra checks, "trusted" means the content is known to be safe somehow. (There might be some historic inconsistency about that, feel free to change those places if you come across anything.)

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -10,6 +10,10 @@ export interface CompileOptions {
   moduleName?: string;
 }
 
+function isTrustingValue(value) {
```
chancancode: This should be "isTrusted" (meaning the value has been TSA pre-✔️ ed)

trusted is TSA pre-✔️ed travelers, trusting is a lousy security guard just let everyone through

krisselden: another way to think about it is operation vs value, you create a trusting operation like trusting append, and then there are trusted values that have been prechecked or asserted to have been

wycats: @krisselden exactly. That's why I used this naming. 

<table>
<tr>
  <th></th><th>Trusted Value</th><th>Untrusted Value</th>
</tr>
<tr>
  <th>Trusting Operation (triple curlies)</th><td>✔️</td><td>✔️</td>
</tr>
<tr>
  <th>Cautious Operation (double curlies)</th><td>✔️</td><td>✖️</td>
</tr>
</table>

Trusted Values:

* Literals in a template
* Escaped strings using `escapeExpression()`
* Strings marked as safe using `htmlSafe()`

Untrusted Values:

* JavaScript strings that were neither escaped nor marked as safe

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -386,17 +386,19 @@ export class DynamicAttrNSOpcode extends Opcode {
...
     this.name = name;
     this.namespace = namespace;
+    this.isTrusting = isTrusting;
   }
 
   evaluate(vm: VM) {
     let { name, namespace } = this;
     let reference = vm.frame.getOperand();
-    vm.stack().setAttributeNS(namespace, name, reference);
+    vm.stack().setAttributeNS(namespace, name, reference, this.isTrusting);
```
chancancode: nit: `let { name, namespace, isTrusting } = this;` for consistency, can you fix after merge?

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -418,16 +420,18 @@ export class DynamicAttrNSOpcode extends Opcode {
...
     super();
     this.name = name;
+    this.isTrusting = isTrusting;
   }
 
   evaluate(vm: VM) {
     let { name } = this;
     let reference = vm.frame.getOperand();
-    vm.stack().setAttribute(name, reference);
+    vm.stack().setAttribute(name, reference, this.isTrusting);
```
chancancode: ditto

PR #217, opened 2016-07-15
Let the host (e.g. Ember) handle layout caching

chancancode:
This removes the special CACHED_LAYOUT symbol and instead replaced it with a `layoutFor` hook on the component manager. The primary reason for this is to decouple layout caching from component definitions (the hosts can implement their own caching layer).

In addition to that, it also solves the problem of Ember’s “late-bound” layout feature, because the `layoutFor` hook is called *after* `create`, and we pass in the component instance as an argument to `layoutFor`. This allows you to make whatever decisions you need to construct/lookup the right layout.

Finally, this also allows the hosts to return optimized layouts based on instance-side information.


chancancode:
TODO: deal with visualizer and `layoutFor`

chancancode:
Also: bring back caching

PR #219, opened 2016-07-16
pin typescript compiler to ^1.8

fivetanley:
The "next" tag for typescript has shifted to '2.1.0-dev.20160716',
when it was previously a 1.x value. The `broccoli-typescript-compiler`
requires a peer dependency of `typescript@^1.6.2 || ^1.7.0 || ^1.8.0 ||
^1.9.0-dev`, so a `2.x` does not fall within this semver range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "handlebars": "^3.0.2",
     "qunit": "^0.7.2",
     "simple-html-tokenizer": "^0.2.5",
-    "typescript": "next"
+    "typescript": "^1.8.10"
   },
   "devDependencies": {
     "benchmark": "^1.0.0",


fivetanley:
We may want to use typescript 2.x here, I don't know the right answer.

rwjblue:
We actually require features in 2.x series.  @wycats /  @chancancode / @krisselden may have more info.

fivetanley:
Ok I will close this one and open one against broccoli-typescript-compiler to allow for 2.x then

rwjblue:
I believe the peer dep listed in broccoli-typescript-transpiler needs to be updated

fivetanley:
opened https://github.com/tildeio/broccoli-typescript-compiler/pull/20 instead

wycats:
For what it's worth, it's very important to me that we remain on the latest TypeScript to avoid it becoming a project simply to use new features.

rwjblue:
Hehe. It's a chore either way. This way our build breaks randomly for unrelated things as the Typescript project changes API's we are using (or starts doing more strenuous type checking where there was none before), the other way (locking to a specific version) lets us pick when to deal with potential breakage from newer typescript versions. 

I don't particularly care which we do, but it's a work either way. :sweat_smile:

PR #229, opened 2016-07-24
Cache compiled layouts

### LINKED ISSUE #228: Implement caching in TestEnvironment
Now that caching for layouts is done by consuming env we should implement this caching in the `TestEnvironment` so our benches and demos don't artificially regress.

chadhietala:
Fixes #228

### DIFF
--- a/packages/glimmer-test-helpers/lib/environment.ts
+++ b/packages/glimmer-test-helpers/lib/environment.ts
@@ -286,8 +286,12 @@ class BasicComponentManager implements ComponentManager<BasicComponent> {
     return new klass(args.named.value());
   }
 
-  layoutFor(definition: BasicComponentDefinition, component: BasicComponent, env: Environment): CompiledBlock {
-    return compileLayout(new BasicComponentLayoutCompiler(definition.layoutString), env);
+  layoutFor(definition: BasicComponentDefinition, component: BasicComponent, env: TestEnvironment): CompiledBlock {
+    if (env.compiledLayouts[definition.name]) {
+      return env.compiledLayouts[definition.name];
+    }
+
+    return env.compiledLayouts[definition.name] = compileLayout(new BasicComponentLayoutCompiler(definition.layoutString), env);
   }
 
   getSelf(component: BasicComponent): PathReference<Opaque> {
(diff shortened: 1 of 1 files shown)


stefanpenner:
test?

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -441,14 +452,18 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
     return component;
   }
 
-  layoutFor(definition: EmberishCurlyComponentDefinition, component: EmberishCurlyComponent, env: Environment): CompiledBlock {
+  layoutFor(definition: EmberishCurlyComponentDefinition, component: EmberishCurlyComponent, env: TestEnvironment): CompiledBlock {
     let layoutString = definition.layoutString;
 
     if (!layoutString && layoutString !== '') {
       layoutString = component['layout'];
     }
```
chancancode: This is the late-bound layout case – so we either cannot cache it (just do a early return here), or need to change the cache key to account for that.

If the goal of this is to not regress demos and benchmarks, it's probably fine to not cache late-bound layouts – I don't think that feature is used outside of tests.

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -441,14 +452,18 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
...
 
-    return compileLayout(new EmberishCurlyComponentLayoutCompiler(layoutString), env);
+    if (env.compiledLayouts[definition.name]) {
+      return env.compiledLayouts[definition.name];
+    }
```
chancancode: Save the result of the lookup in a local variable, and move this to the top (above the layoutString stuff)

chancancode:
Merged it because I need it for my visualizer work. I think @krisselden has some ideas re: testing

stefanpenner:
@chancancode merging without test is a setting a poor precedent, I would recommend keeping your visualizer code in a branch until this is complete next time.

I have open an issue, so the tests are not forgotten: https://github.com/tildeio/glimmer/issues/232

chadhietala:
I'm not sure if tests actually make sense here. `layoutFor` is an abstract method that a component manager must implement otherwise you will get a compile error. We specifically moved the caching into user-land/consuming environment so if we were to write tests we would be writing tests for test helpers that are specific the glimmer test harness / demos.

chadhietala:
Maybe there should be tests for the demos?

chancancode:
Yeah, I think it's somewhat orthogonal to this, but I always wanted some smoke tests (as in QUnit.test) for the demos so when a feature breaks the demos we will see it on CI.

As for testing this specifically, I agree that there aren't a lot of meaningful assertions you could write about this. The glimmer semantics is that this book will be called (which is for sure covered by the integration tests) once per invocation, the "interesting" thing here is happening inside the test environment. It's kind of a pretty meta thing to be testing. You can write a test to test that the compilation is happening the right amount of times, but it will be testing the test environment. Not completely unreasonable given the demos uses it, but I think overal the more important thing here is to have tests for the demos.

PR #230, opened 2016-07-25
Optimistic append

chancancode:
This enables a feature (closure components) in Ember where certain curlies can dynamically resolve into component definition objects and need to be rendered as component invocations.

The general strategy to handle that is to change all curlies into (at least semantically) the equivalent of:

```hbs
{{#if (isComponentDefinition foo)}}{{component foo}}{{else}}{{foo}}{{/if}}
```

However, doing all that extra work and bookkeeping for literally every single curly would be very slow. Therefore, we take a tiered approach to handle this.

First of all, in Ember the syntactic requirement for this feature is that there needs to be at least one dot in the path lookup, e.g. `... as |f|}}{{f.input}}...`, `{{foo.bar.baz}}` or `{{this.foo}}`. This allows us to emit the same append opcode (now called `OptimizedAppend`) for the majority of the append curlies. (By default, all curlies are optimized, but the consuming environment can override that in `refineStatement`.)

Even then, there are still a large amount of path lookups with a dot that will not ever resolve into a component, and doing the extra work for all those cases would be less than ideal. For that reason, even we need to emit `UnoptimizedAppend`, we still assume that the values are going to be simple values.

If we discover that the assumption is not true, i.e. we see that the value is indeed a component definition at either initial render time or at update time, we will then deopt the opcodes into the expanded form.

See the code comments for more details on how it works.

TODO:

- [x] Handle deopts in the visualizer

cc @wycats @krisselden @zackthehuman @chadhietala


chadhietala:
I feel like we should add some benchmarks around this to understand what it means to "deopt".

chancancode:
As seen in visualizer:

## Optimized (i.e. no dots in path)

```hbs
{{#with (if name (component name) "nope") as |foo|}}
  {{foo}}
{{/with}}
```

Rendered with a simple value (`{ name: null }`):



Rendered with a component definition (`{ name: "h-card" }`):



## Unoptimized (i.e. no dots in path)

```hbs
{{#with (hash bar=(if name (component name) "nope")) as |foo|}}
  {{foo.bar}}
{{/with}}
```

Rendered with a simple value (`{ name: null }`):



Eager deopt: rendered with a component definition (`{ name: "h-card" }`):



Lazy deopt: Rendered initially with a with a simple value (`{ name: null }`), then updated with a component definition (`{ name: "h-card" }`):

packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -74,7 +77,7 @@ export abstract class BasicOpcodeBuilder extends StatementCompilationBufferProxy
     super(inner);
   }
 
-  abstract compile<E>(expr: CompilesInto<E>): E;
+  abstract compile<E>(expr: Represents<E> | E): E;
```
Serabe: Don't know much about TypeScript, but given `Represents<E> = CompilesInto<E> | E`, is it necessary to add the `| E` here too?

chancancode: You are correct sir

packages/glimmer-demos/lib/visualizer.ts
```diff
@@ -360,9 +362,10 @@ function renderContent() {
 
   let env = new TestEnvironment();
 
-  env.registerHelper('foo', function(){
-    return 'FOOOOOOO';
+  env.registerHelper('foo-bar', function(){
```
krisselden: hyphens aren't required for lazy helper lookup

packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -263,8 +266,18 @@ export abstract class BasicOpcodeBuilder extends StatementCompilationBufferProxy
...
+    if (testFunc === 'const') {
+      this.append(new vm.TestOpcode(vm.ConstTest));
+    } else if (testFunc === 'simple') {
+      this.append(new vm.TestOpcode(vm.SimpleTest));
+    } else if (testFunc === 'environment') {
+      this.append(new vm.TestOpcode(vm.EnvironmentTest));
+    } else if (typeof testFunc === 'function') {
+      this.append(new vm.TestOpcode(testFunc));
+    } else {
+      throw 'unreachable';
```
krisselden: `throw new Error('unreachable')`

packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -281,8 +294,12 @@ export abstract class BasicOpcodeBuilder extends StatementCompilationBufferProxy
 }
 
 export default class OpcodeBuilder extends BasicOpcodeBuilder {
-  compile<E>(expr: CompilesInto<E>): E {
-    return expr.compile(this, this.env, this._block);
+  compile<E>(expr: Represents<E>): E {
+    if (typeof expr['compile'] === 'function') {
```
krisselden: can you `isCompilableExpression(expr: Represents<E>): expr is CompilesInto<E>`

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -89,11 +106,121 @@ export abstract class AppendOpcode<T extends Insertion> extends Opcode {
...
+    // Keep in mind that even if we *don't* reach here at initial render time,
+    // it is still possible (although quite rare) that the simple value we
+    // encounter during initial render could later change into a component
+    // definition object at update time. That is handled by the "lazy deopt"
+    // code on the update side (scroll down for the next big block of comment).
+
+    let buffer = new CompileIntoList(env, null);
+    let dsl = new OpcodeBuilderDSL(buffer, null, env);
+
+    dsl.block({ templates: null }, (dsl, BEGIN, END) => {
```
krisselden: can there be an empty const for `{ templates: null }`? it isn't mutable?

krisselden: Also, I thought we were going to move away from pojo args. is this the only shape the dsl sees?

packages/glimmer-runtime/lib/compiled/opcodes/content.ts
```diff
@@ -62,21 +71,29 @@ export function normalizeValue(value: Opaque): CautiousInsertion {
...
+    }
 
     let stack = vm.stack();
     let upsert = this.insert(stack.dom, stack, value);
     let bounds = new Fragment(upsert.bounds);
 
-    vm.stack().newBounds(bounds);
+    stack.newBounds(bounds);
 
     if (!isConst(reference)) {
```
krisselden: can we call isConst once?

packages/glimmer-demos/lib/visualizer.ts
chancancode: I just changed it because I keep doing `{{#with ... as |foo|}}...{{foo}}` and `foo` ended up winning as a helper (seems like a bug??), so I named it something less common

packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
chancancode: :+1:

packages/glimmer-runtime/lib/component/interfaces.ts
```diff
@@ -75,11 +75,20 @@ export interface ComponentAttrsBuilder {
   dynamic(name: string, value: FunctionExpression<string>);
 }
 
+
+const COMPONENT_DEFINITION_BRAND = 'COMPONENT DEFINITION [id=e59c754e-61eb-4392-8c4a-2c0ac72bfcd4]';
+
+export function isComponentDefinition(obj: any): obj is ComponentDefinition<Opaque> {
```
krisselden: please use `typeof obj === 'object' && obj && obj[COMPONENT_DEFINITION_BRAND]` so it doesn't compile a coercion to boolean for any value.

This is common source deopts if obj has a range of types and optimizes the coercion before seeing them all whereas typeof is generic and reduces the coercion to just checking for null.

krisselden:
LGTM

PR #231, opened 2016-07-25
Add prepareArgs to ComponentManager<T> interface

zackthehuman:
This adds the `prepareArgs` hook to the `ComponentManager<T>` interface as well as a default implementation for the test helpers. This is needed for closure component support.

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/component.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/component.ts
@@ -84,22 +84,23 @@ export class OpenComponentOpcode extends Opcode {
 
     let manager = definition.manager;
     let hasDefaultBlock = templates && !!templates.default; // TODO Cleanup?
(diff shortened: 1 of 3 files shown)


chancancode:
:+1:

chancancode:
Looks good to me if tests are green

chancancode:
This could use a test, but if you are going to add tests for closure components anyway that would cover it.

zackthehuman:
I'm not sure how I would test this directly in a meaningful way. If the managers in `ember-test-helpers` made a call to it in their `create` method and everything was green, would this be sufficient?

chancancode:
You could make a new kind of component manager (via refine statement or via a helper passed the component syntax or whatever) that, say, reverses or empties the args and test that it does that.

But since I assumed you are going to do some closure component helper tests here, that would likely cover it and be a more meaningful test. You don't have to do the complicated merging logic, you could go for something simplier like "last one wins".

PR #233, opened 2016-07-25
Actually run Glimmer test suite

tomdale:
#214 introduced an issue where all tests would appear to pass, but in reality no tests were being run.

That's because the new tslint broccoli tree was filtering out non-test files. This change merges the two trees instead of assuming that the tslinter passes through non-transformed files.


rwjblue:
👍 - Good catch!

PR #240, opened 2016-07-28
Bump loader.js

chadhietala:


ember-cli-build.js
```diff
@@ -271,8 +271,8 @@ module.exports = function() {
   ];
 
   if (hasBower) {
```
josemarluedke: Since you moved `loader.js` from bower, this check seems to not make much sense now.

PR #242, opened 2016-07-29
Prevent .tslint files from getting embedded into Ember.

rwjblue:


rwjblue:
Once I fixed the `tests/index.html` to load the TSLint files I realized that TSLint is not passing on the repo.  I will open an issue for that so someone can follow-up and re-enable tslint.

PR #244, opened 2016-07-30
Always compute before “stamping” the validation

chancancode:
Computing the reference value could potentially be side-effectful (from the perspective of the revision timeline), so you should always compute the value before getting the tag value.


chancancode:
cc @wycats @krisselden

PR #245, opened 2016-07-30
Fix undefined input values

chadhietala:


chancancode:
Oops, wrong browser profile, that was me merging 😄 

I was confused about this earlier, what is the difference between `InputValuePropertyChangeList` and `InputValueAttributeChangeList`? I mean I know what they do differently, but when are each of those used?

chancancode:
Hm I missed it earlier, can you test updating also?

chadhietala:
Ah yes `InputValueAttributeChangeList` can go away it's effectively dead code. Will follow up with another PR.

chancancode:
Basically following INUR

PR #249, opened 2016-08-02
Make AST traversal recursive

chancancode:
Previously, if you return a new AST node, that new node is not walked.

This is causing a bug in Ember where we transform:

```hbs
{{#each-in foo ...}}...{{/each-in}}
```

...into...

```hbs
{{#each (-each-in foo) ...}}...{{/each}}
```

In the case where there are nested `each-in`s, like this:

```hbs
{{#each-in foo as |fooKey fooValue|}}
  {{#each-in fooValue as |barKey barValue}}
    ...
  {{/each-in}}
{{/each-in}}
```

Because we walk from the outside in, and the outside `{{#each-in}}` returned a new node, the inner block was never walked, causing the inner `{{#each-in}}` to not be transformed.

This commit fixes the problem by recursively walking the returned nodes.

### DIFF
--- a/packages/glimmer-syntax/lib/traversal/traverse.ts
+++ b/packages/glimmer-syntax/lib/traversal/traverse.ts
@@ -13,6 +13,16 @@ function visitNode(visitor, node) {
     result = handler.enter.call(null, node);
   }
 
+  if (result !== undefined && result !== null) {
+    if (JSON.stringify(node) === JSON.stringify(result)) {
+      result = undefined;
+    } else if (Array.isArray(result)) {
+      return visitArray(visitor, result) || result;
+    } else {
+      return visitNode(visitor, result) || result;
+    }
+  }
+
   if (result === undefined) {
     let keys = visitorKeys[node.type];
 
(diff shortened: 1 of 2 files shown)


mmun:
This should be reverted.

It is intentional that nodes that you return are not traversed. It is a similar API to Babel.

Previously, if you wanted to run the same traversal on the new subtree you could just explicitly do it. After this change, it is difficult to _not_ run a subtraversal if you don't want it. This could lead to things like unexpected infinite loops, etc. I wouldn't be surprised if this broke code out there.

mmun:
As an example, it is idiomatic to simply return the node itself if you want to traversing into that node:

```
BlockStatement(node) { return node; }
```

We could also investigate changing the semantics to align with this PR and introducing a `return this.break();` or something to stop the traversal.

wycats:
> Previously, if you wanted to run the same traversal on the new subtree you could just explicitly do it

What's the idiomatic way to do that?

The thing that bothers me is that if you forget to say traverse into a block, you just... lose the contents of the block. It was largely coincidental that the existing AST transforms (which largely *don't* continue the traversal) don't have nested contents that also needs to be traversed.

mmun:
This is what I mean by explicit traversal of the subtree:

```js
let visitor = {
  MustacheStatement(node) {
    if (isNodeToBeTransformed(node)) {
      let transformedNode = transformNode(node);
      traverse(transformedNode, visitor);
      return transformedNode;
    }
  }
};

traverse(astRoot, visitor);
```

mmun:
I'm not opposed to the semantics in this PR if we think that it won't break apps, but I want to have _some way_ to bail out of traversing a subtree.

PR #252, opened 2016-08-03
Allow for null literals

chadhietala:


chancancode:
Does other kind of literals (true false strings numbers etc) have this problem? Also, is undefined actually a thing we support ?_?

packages/glimmer-wire-format/index.ts
```diff
@@ -78,7 +78,10 @@ export namespace Expressions {
   export const isHasBlockParams = is<HasBlockParams>('hasBlockParams');
 
   export function isValue(value: any): value is Value {
-    return value !== null && typeof value !== 'object';
+    if (value === null) {
+      return true;
```
rwjblue: This confuses me. If `value` is `null` the `isValue` method is supposed to return true? Seems to me like null/undefined are **not** values?

rwjblue:
Can we also have a test for when s bound value resolves to null/undefined (vs just using the null/undefined literal in the template).

chancancode:
@rwjblue I _think_ we already have tests for those, and they are pretty
unrelated to this at this level (syntax). But if we somehow don't have them
then yeah we should add some.

chancancode:
(Those tests would be in the updating tests file)

packages/glimmer-wire-format/index.ts
chadhietala: Changed this to `isPrimitiveValue`

packages/glimmer-runtime/lib/syntax/expressions.ts
```diff
@@ -24,11 +24,11 @@ const {
...
   isUnknown,
-  isValue
+  isPrimitiveValue
 } = SerializedExpressions;
 
 export default function(sexp: SerializedExpression): any {
-  if (isValue(sexp)) {
+  if (isPrimitiveValue(sexp)) {
     return ValueSyntax.fromSpec(sexp);
   } else {
```
chancancode: Can you change this to use the same `if (...) return` syntax/style?

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -349,6 +349,46 @@ test("The compiler can handle simple helpers", function() {
   compilesTo('<div>{{testing title}}</div>', '<div>hello</div>', { title: 'hello' });
 });
 
+test("GH#13999 The compiler can handle simple helpers with inline null parameter", function() {
+  env.registerHelper('say-hello', function() {
+    return 'hello';
```
chancancode: Can you assert that the argument is actually `null`?

Something like:

```
let value;

env.registerHelper('say-hello', function(params) {
  value = params[0];
  return hello;
});

strictEqual(value, ..., '...');
```

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -349,6 +349,46 @@ test("The compiler can handle simple helpers", function() {
...
+
+  compilesTo('<div>{{say-hello null}}</div>', '<div>hello</div>');
+});
+
+test("GH#13999 The compiler can handle simple helpers with inline undefined parameter", function() {
+  env.registerHelper('say-hello', function() {
+    return 'hello';
+  });
+
+  compilesTo('<div>{{say-hello undefined}}</div>', '<div>hello</div>');
```
chancancode: I suspect that this is not something we actually support. The wire format is converted into json, which doesn't support `undefined` as a value. `undefined` feels very JavaScript-centric, whereas Handlebars try to stay a bit more host-language-neutural in general (there are implementations in other languages, and having `undefined` vs `null` in those environments would feel quite out-of-place).

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -349,6 +349,46 @@ test("The compiler can handle simple helpers", function() {
...
+});
+
+test("GH#13999 The compiler can handle simple helpers with inline undefined parameter", function() {
```
chancancode: duplicated test?

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -349,6 +349,46 @@ test("The compiler can handle simple helpers", function() {
...
+});
+
+test("GH#13999 The compiler can handle components with undefined named arguments", function() {
+  env.registerHelper('say-hello', function() {
+    return 'hello';
+  });
+
+  compilesTo('<div>{{say-hello foo=undefined}}</div>', '<div>hello</div>');
+});
+
```
chancancode: I can think of at least a few more cases to test – `{{null}}` ("append"/content position), `<div class="foo {{null}} bar" />` (concat) and `{{#if null}}` ("block" arguments).

chancancode:
Looks good other than some minor things. This is pretty important, so we probably should test this in Ember side also (once this lands).

chancancode:
Does TypeScript allow you to explicitly add ` | null` to https://github.com/tildeio/glimmer/blob/master/packages/glimmer-wire-format/index.ts#L43 without strict null-checking mode?

chadhietala:
It is a tslint error to introduce nullable types in that union.

packages/glimmer-runtime/tests/initial-render-test.ts
chadhietala: Added these guys

chancancode:
Maybe we should either configure tslint to ignore that line, or at least add `/* | null */;` for documentation

chadhietala:
Just ignored the line from tslint.

chancancode:
http://tryhandlebarsjs.com seems to suggest that Handlebars do support and differentiate between `undefined` and `null` (which I think is pretty strange/inappropriate):

tryhandlebarsjs

@wycats (@mmun?) can you clarify what we need to support here?

chadhietala:
There is `UndefinedLiteral` and `NullLiteral`s

chancancode:
I think there are a few options 1) maybe the language doesn't care and this is just a host extension, since we work in JS we support and differentiate both for convenience 2) the language requires us to support and differentiate both 3) we don't have to support undefined. 

Either way, it seems like you have found a bug: accepting undefined syntax but turning it into null doesn't seem acceptable; if we have to support it at all we would need to round trip it in the wire format somehow.

(That being said, we could land this fix and the tests and mark the undefined tests as bugs)

chadhietala:
So from a back compat point of view `UndefinedLiteral`s resolve to `null` in HTMLBars, however from from an implementation point of view it is incorrect.

chadhietala:
I have created https://github.com/emberjs/ember.js/issues/14016.

PR #255, opened 2016-08-04
Preserve UndefinedLiteral

chadhietala:


chadhietala:
Eh I wondering if just a custom replacer and reviver would work.

chadhietala:
After some investigation the replacer/reviver would not work as well.

packages/glimmer-compiler/lib/javascript-compiler.ts
```diff
@@ -182,7 +182,13 @@ export default class JavaScriptCompiler {
   /// Expressions
 
   literal(value: any) {
-    this.pushValue(value);
+    if (value === 'undefined') {
+      this.pushValue(JSON.stringify("undefined"));
```
chancancode: Overloading strings wouldn't work – now if you have an incoming string `"undefined"` you would have a conflict. This also unnecessarily complicates decoding strings; I think we should just go with `{ "undefined": true }` or `{ "literal": "undefined" }`. This is rarely used (and if is it would be gzipped just fine), and we have bought a lot of headroom for wire format size, so I'm not too worried about it.

PR #258, opened 2016-08-10
Don't compile on save

asakusuma:
Let the build handle compilation


tomdale:
@asakusuma Isn't this off by default?

locks:
That's the default value that VS Code fills it with, so I'm assuming it is false by default. Please re-open if that's not so!

PR #260, opened 2016-08-11
Pass block meta into partial methods

asakusuma:
Needed for engines

cc @krisselden @chancancode


asakusuma:
Update to use `owner` from `refineStatement` to `PartialSyntax` to the partial opcodes to the partial lookup methods

PR #265, opened 2016-08-14
Update testem config for sauce labs runs.

rwjblue:
* Adds Microsoft Edge
* Updates Chrome to use `latest` and `latest-1` (was using 39 and 38)
* Updates Firefox to use `latest` and `latest-1` (was using 34 and 33)
* Updates Safari to use 9 and 8 (was 8 and 7)
* Wraps `<url>` in quotes to avoid issues with quoting discovered in Ember.

### DIFF
--- a/testem-sauce.json
+++ b/testem-sauce.json
@@ -7,39 +7,43 @@
   "launchers":
     {
       "SL_Chrome_Current": {
-        "command": "ember sauce:launch -b chrome -v 39 --no-ct -u <url>",
+        "command": "ember sauce:launch -p 'Windows 10' -b chrome -v latest --no-ct -u '<url>'",
         "protocol": "tap"
       },
       "SL_Chrome_Last": {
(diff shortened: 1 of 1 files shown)


rwjblue:
Grumble grumble.  Looks like `latest` isn't working.

johanneswuerbach:
I asked sauce support and that is what they said:

> The problem here is the platform you’re (not) specifying. The idea behind the latest version capability is for platforms and browsers where there are multiple browser versions supported on a particular platform, you can get the latest (or latest-n) – that’s Chrome and Firefox really.

> For MicrosoftEdge, “latest” works if you specify “Windows 10” as your platform, but we only support one version of Edge anyway.

> Similarly with Safari, “latest” works if you specify a platform, such as “OS X 10.11” – but again, we only support one version of Safari per OS X version anyway.

Basically specifying platform should help.

rwjblue:
Awesome, I will update. 

Thank you @johanneswuerbach!

PR #266, opened 2016-08-14
Use `dom.setAttribute` for non-dynamic (static) attributes.

rwjblue:
This is a failing test for https://github.com/emberjs/ember.js/issues/13992.

Discussed with @wycats and will be implementing a separate
`addDynamicAttribute` (which will be what `addAttribute` will become)
and an `addStaticAttribute` to `ElementStack`.


rwjblue:
@wycats - You mentioned to ping you when I had this up for review.

PR #267, opened 2016-08-16
Failing tests for `<select multiple>`.

rwjblue:
This is a follow-up to https://github.com/tildeio/glimmer/pull/266 and adds tests for dynamic selected attributes (as opposed to only static).

Related to https://github.com/emberjs/ember.js/issues/13992.

### DIFF
--- a/packages/glimmer-runtime/tests/initial-render-test.ts
+++ b/packages/glimmer-runtime/tests/initial-render-test.ts
@@ -319,6 +319,82 @@ test("Mountain range of nesting", function() {
              'FOO<span>BAR<a>BAZ<em>BOOBREW</em>BAT</a></span><span><span>FLUTE</span></span>ARGH', context);
 });
 
+test("Static <option selected> is preserved properly", function() {
+  let template = compile(`
+    <select>
+      <option>1</option>
+      <option selected>2</option>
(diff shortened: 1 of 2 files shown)


rwjblue:
Updated with a fix that uses `setAttribute` for initial render, and sets via a property for updates.

rwjblue:
While researching the failure, we discovered a somewhat fundamental issue with the way that we are doing our DOM building.  

Currently, attributes are attached to a given element when we call in the `close-element` opcode.  This is problematic in the `<select multiple>` case because the `multiple` attribute **must** be present when `<option>`'s are added or the `selectedness` for all but the last `<option` that has a `selected` attribute will be changed to `false`.  [Relevant section](https://html.spec.whatwg.org/multipage/forms.html#the-select-element) from the spec.

[screenshot](https://html.spec.whatwg.org/multipage/forms.html#the-select-element)

@chancancode and I started spiking on adding a `flush-attributes` opcode that would be ran before the first opcode that is not a `static-attr` or `dynamic-attr`.  This will allow us to flush all attributes to the element before any child contents are added.

rwjblue:
OK, after more discussion with @chancancode we decided to leave this PR as a failing test PR and do the actual implementation separately.

PR #273, opened 2016-08-17
Tree construction

chadhietala:
This splits DOMTreeConstruction and DOMChanges into 2 separate classes, the idea being that you implement just a Node DOMTreeConstruction class for SSR.


packages/glimmer-runtime/lib/dom/helper.ts
```diff
@@ -165,16 +196,55 @@ class DOMHelper {
...
 }
 
+export function insertHTMLBefore(this: void, _useless: Simple.HTMLElement, _parent: Simple.Element, _nextSibling: Simple.Node, html: string): Bounds { // tslint:disable-line
+  // TypeScript vendored an old version of the DOM spec where `insertAdjacentHTML`
+  // only exists on `HTMLElement` but not on `Element`. We actually work with the
+  // newer version of the DOM API here (and monkey-patch this method in `./compat`
+  // when we detect older browsers). This is a hack to work around this limitation.
+  let parent = _parent as HTMLElement;
+  let useless = _useless as HTMLElement;
+  let nextSibling = _nextSibling as Node;
```
chadhietala: @wycats Does this casting seem right? I believe this pass is only going to be taken in the browser so it seems fine to upcast these here.

packages/glimmer-runtime/lib/dom/helper.ts
```diff
@@ -104,39 +167,7 @@ class DOMHelper {
...
-      // from merging adjacent text nodes. See ./compat/text-node-merging-fix.ts
-      parent.insertBefore(this.uselessElement, nextSibling);
-      this.uselessElement.insertAdjacentHTML('beforeBegin', html);
-      last = this.uselessElement.previousSibling;
-      parent.removeChild(this.uselessElement);
-    }
-
-    let first = prev ? prev.nextSibling : parent.firstChild;
-    return new ConcreteBounds(parent, first, last);
+    return insertHTMLBefore(this.uselessElement, _parent, nextSibling, html);
```
chadhietala: At one time this was `abstract` however it seems like there needs to be a concrete implementation, because we need to apply HTML fixes for tree construction that delegate back to `super` in certain cases.

PR #283, opened 2016-08-21
Cache property normalization

chadhietala:


packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -358,29 +358,40 @@ export class DynamicAttribute implements Attribute  {
...
     this.namespace = namespace;
+    this.normalizedKey = `${element.tagName}.${this.namespace ? this.namespace + '.' : ''}${this.name}.${element.namespaceURI}`;
   }
 
   patch(env: Environment) {
     let { element, cache } = this;
+    let normalized = env.normalizedProperties[this.normalizedKey];
+    let name = this.name;
+    if (typeof normalized === 'object') {
+      name = env.normalizedProperties[this.normalizedKey].normalized;
```
krisselden: Please use an api for setting/getting caches

chadhietala:
@krisselden Is there a specific type of cache that should be used?

chadhietala:
Also we will need to change Ember since the return value from `attributeFor` is different.

chancancode:
Looking good. I have a few thoughts:

1. We need to split up the caching part of the hook (which the consuming env wouldn't need to override normally) and the guts of hook (which the consuming env would likely need to override). Otherwise (as you can see in your `TestEnvironment` code), the consuming env would always have to duplicate the keying and caching logic before calling super. I suggest something like `attributeFor` (cached) vs `lookupAttribute` (uncached user-hook). I think the cache should be `private`.

2. `attributeFor` returns `{ type, name, manager }`, but I don't think `type` is used anywhere. (That's the point of the manager – its job is to encapsulate the difference.)

3. Following the same logic as (2), I think we should just encapsulate the `name` difference in the manager as well.

I'll elaborate on (3). The original intent of making the managers static is to minimize allocation. However, if we need to allocate a `{ name, manager }` POJO per `(ns, tagName, attrName)` combo, that seems like no better than just allocating one `new SpecialSauceManager(normalizedName)` per key. @krisselden can correct me if I am wrong, but I don't think that's any more expensive than making the POJO.

With that setup, the manger can store off the `normalizedName` and encapsulate the difference. Just as you won't have to pass `type` ('attr' vs 'prop') to the managers in `setAttribute` and `updateAttribute` (see point 2), you can remove the `name` argument from those hooks and no one has to worry about destructuring, etc.

How dose that sound?

chancancode:
@chadhietala re: caching, you can start with this:

```typescript
class Cache<T> {
  private store = new EmptyObject<T>();

  get(key: string): T {
    return this.store[key];
  }

  set(key: string, value: T): T {
    return this.store[key] = value;
  }

  fetch(key: string, func: (key) => T): T {
    return this.get(key) || this.set(key, func(key));
  }
}
```

(Note that this doesn't handle caching `undefined` as a value, but seems fine for now)

I think the idea is like `Dict`, where we make an explicit type to encode the intent, which allows us the experiment with different implementation strategy later (say, use an array to boot, then upgrade to an empty object, etc).

packages/glimmer-runtime/lib/dom/attribute-managers.ts
```diff
@@ -0,0 +1,151 @@
...
+
+   if (type === 'attr') {
+     return element.getAttribute(name);
+   } else {
+     return element[name];
+   }
+};
+
+export class AttributeManager {
+  constructor(public name: string) {}
```
chancancode: Shouldn't this be private (protected?)

packages/glimmer-runtime/tests/initial-render-test.ts
```diff
@@ -985,23 +985,23 @@ test("Case-sensitive tag has capitalization preserved", function() {
...
-  setAttribute(dom, element, attr, value) {
+class StyleAttributeManager extends AttributeManager {
+  setAttribute(dom, element, value) {
     warnings++;
-    AttributeChangeList.setAttribute(dom, element, attr, value);
-  },
+    super.setAttribute(dom, element, value);
+  }
   updateAttribute() {}
 };
```
chancancode: It doesn't really matter here, but in Ember this should be non-allocating, since it have the same `name` for all instances. (You could just make const STYLE_ATTRIBUTE_MANAGER = new Style... and keep returning the same instance for all tagNames)

packages/glimmer-runtime/lib/environment.ts
```diff
@@ -117,9 +119,15 @@ export class Scope {
   }
 }
 
+export interface NormalizedAttributeManager {
+  name: string;
+  attributeManager: AttributeManager;
+}
```
chancancode: Dead code

chadhietala: 🔪 💀 🔪

tomdale:
Closing due to how old this is.

PR #285, opened 2016-08-21
Component bounds

chancancode:


packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -507,6 +519,10 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
     }
   }
 
+  didRenderLayout(component: EmberishCurlyComponent, bounds: Bounds) {
+    component.bounds = bounds;
```
krisselden: Let's not add a conflict with user land props

krisselden: Also shaping issues with this, I'm assuming this is for getClientRect(view) so we just need to loosly  associate a bounds

packages/glimmer-runtime/lib/compiled/opcodes/component.ts
```diff
@@ -5,116 +5,86 @@ import { VM, UpdatingVM } from '../../vm';
...
-  constructor({ args, definition }: PutComponentDefinitionOptions) {
+  constructor(private definition: ComponentDefinition<Component>) {
     super();
-    this.args = args;
-    this.definition = definition;
   }
 
   evaluate(vm: VM) {
-    let args = this.args.evaluate(vm).withInternal();
-    args.internal["definition"] = this.definition;
```
krisselden: Yay!

krisselden:
Looks awesome

packages/glimmer-test-helpers/lib/environment.ts
locks: can you ELI5? trying to dig into this sort of stuff

rwjblue: The changes we are commenting on are in the `glimmer-test-helpers` package and are intended to roughly emulate what Ember's environment will do.

> Let's not add a conflict with user land props

@krisselden is basically saying that we should not stomp on properties that may exist already on the `component` passed in here.  We could use a private symbol, an underscored property, or even put information on a weakmap/meta object instead.

> Also shaping issues with this

Changing an objects shape (adding new methods or properties) after its constructor has completed (aka after `init`) makes the specific "class" have many possible shapes. Since the shape of the object is taken into consideration by V8 when optimizing this changing of shape can trigger deoptimizations when previously optimized code now sees a different shape.

chancancode: Yeah, this only need to _loosely_ emulate what Ember would do, and I think this is close enough. The ember version does use a symbol and pre-shape it in the constructor: https://github.com/emberjs/ember.js/commit/376a88df885fd4e88f39542988cdddd6021aded1

PR #289, opened 2016-08-23
Use SymbolTable throughout runtime.

rwjblue:
The SymbolTable conceptually represents the static semtantics of a particular piece of Handlebars code. Previously, we passed around blocks, block metas, symbol tables, etc. interchangably, on an as-needed basis.

This commit attempts to reliably pass around SymbolTable whenever static information for a block is needed.

Note that BlockMeta, which is static information supplied by a host environment compiler, is part of this static information, and is now accessed using symbolTable.getMeta().

Also note that there are a number of places that were cheating in various ways to avoid appearing to need the static information, and that is cleaned up in this commit.

### DIFF
--- a/packages/glimmer-compiler/tests/compile-options-test.ts
+++ b/packages/glimmer-compiler/tests/compile-options-test.ts
@@ -15,5 +15,5 @@ QUnit.test('moduleName option is passed into meta', function() {
     env,
     moduleName
   });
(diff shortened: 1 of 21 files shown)


rwjblue:
WHOO! Looks like this is finally ready to review.  Passing changeset for Ember can be reviewed here: https://github.com/emberjs/ember.js/compare/update-glimmer-threading-symbol-table.

PR #290, opened 2016-08-24
Require blockMeta for env lookups

asakusuma:
cc @rwjblue 

Edit: Should simplify code in https://github.com/emberjs/ember.js/pull/14112


rwjblue:
We did https://github.com/tildeio/glimmer/pull/289 instead.

asakusuma:
Yowza that's a big PR

PR #294, opened 2016-08-29
Dynamically-scoped variable accessors

ef4:
This adds the expressison syntax `{{-get-dynamic-var "yourVariableName"}}` and the statement syntax `{{#-with-dynamic-var "yourVariableName" someValue}}...{{/with-dynamic-var}}`.

It's a blocker for a glimmer2-compatible liquid-fire, because we need to read and write the `outletState` variable.

These are both dash-prefixed to mark them private, but https://github.com/emberjs/rfcs/pull/95 discusses making them public.

This is work-in-progress until @chancancode reviews and we simplify.

### DIFF
--- a/packages/glimmer-runtime/lib/syntax/builtins/with-dynamic-var.ts
+++ b/packages/glimmer-runtime/lib/syntax/builtins/with-dynamic-var.ts
@@ -1,0 +1,44 @@
+import {
+  Statement as StatementSyntax
+} from '../../syntax';
+
+import OpcodeBuilderDSL from '../../compiled/opcodes/builder';
+import * as Syntax from '../core';
+import Environment from '../../environment';
+import { default as VM } from '../../vm/append';
+import { DynamicScope } from '../../environment';
+import { EvaluatedArgs } from '../../compiled/expressions/args';
+
+export default class WithDynamicVarSyntax extends StatementSyntax {
+  type = "with-dynamic-var-statement";
+
+  public args: Syntax.Args;
+  public templates: Syntax.Templates;
+  public isStatic = false;
+
+  constructor({ args, templates }: { args: Syntax.Args, templates: Syntax.Templates }) {
+    super();
+    this.args = args;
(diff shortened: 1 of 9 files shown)


ef4:
Tests broke when I rebased.  640c4e9d060efaf137db0d432437dfec3d5c3012 is the culprit. Investigating.

ef4:
Ah, my tests were using `with-keywords`, which was removed.

`with-dynamic-var` itself is probably the appropriate replacement.

rwjblue:
The current build failures look like typescript validation issues. You should be able to replicate with `npm run vscode-build`.

ef4:
What is the appropriate type for this to return? `ValueReference<any>` is no good because `UNDEFINED_REFERENCE` is not of that type.

```ts
  evaluate(vm: VM): ValueReference<any> {
    let scope = vm.dynamicScope();
    if (scope.hasOwnProperty(this.varName)) {
      return scope[this.varName];
    } else {
      return UNDEFINED_REFERENCE;
    }
  }
```

ef4:
Looks like maybe `PathReference<Opaque>` is right, based solely on pattern-matching other places in the codebase.

packages/glimmer-runtime/lib/syntax/builtins/with-dynamic-var.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  compile(dsl: OpcodeBuilderDSL, env: Environment) {
+    let callback = (_vm: VM, _scope: DynamicScope) => {
+      let vm = _vm as any;
+      let scope = _scope as any;
+
+      let args: EvaluatedArgs = vm.frame.getArgs();
+
+      scope[<any>args.positional.values[0].value()] = args.positional.values[1];
```
ef4: Is this case to `<any>` appropriate, or is there a better way to do this?

chancancode:
Closed in #297

PR #297, opened 2016-08-31
Dynamic vars

chancancode:
See #294 for more details


chancancode:
@ef4 we changed the setter API to use named arguments, which solves 1) setting multiple vars at once, and 2) the name being const

mmun:
@chancancode I agree, I was going to suggest that too. We used a similar API in https://github.com/thefrontside/ember-let.

PR #298, opened 2016-09-02
Add meta to compiled template

asakusuma:
Enables looking up the template meta without having to evaluate the actual template. This is a backwards incompatible change that will require changes in places like [here](https://github.com/emberjs/ember.js/blob/01110cb48f224790e717a15a0e5525e793ec5e23/packages/ember-glimmer/lib/template.js#L38-L49).

cc @krisselden @rwjblue

### DIFF
--- a/packages/glimmer-compiler/lib/compiler.ts
+++ b/packages/glimmer-compiler/lib/compiler.ts
@@ -1,7 +1,13 @@
 import { preprocess } from "glimmer-syntax";
+import { BlockMeta } from "glimmer-wire-format";
 import TemplateCompiler from "./template-compiler";
 
-export type TemplateSpec = string;
+export type TemplateProgram = string;
+
+export type TemplateSpec = {
+  meta: BlockMeta;
+  template: TemplateProgram;
+}
 
 /*
  * Compile a string into a template spec string. The template spec is a string
@@ -19,7 +25,11 @@ export type TemplateSpec = string;
  * @return {TemplateSpec} A template spec string
  */
 export function compileSpec(string: string, options: any): TemplateSpec {
+  let meta: BlockMeta = typeof options === 'object' ? options.meta : {};
   let ast = preprocess(string, options);
   let program = TemplateCompiler.compile(options, ast);
-  return JSON.stringify(program);
+  return {
+    meta,
+    template: JSON.stringify(program)
+  };
 }
--- a/packages/glimmer-compiler/tests/compile-options-test.ts
+++ b/packages/glimmer-compiler/tests/compile-options-test.ts
@@ -1,5 +1,6 @@
 import { compile } from "glimmer-test-helpers";
 import { TestEnvironment } from "glimmer-test-helpers";
+import { compileSpec } from "glimmer-compiler";
 
 let env: TestEnvironment;
 
@@ -13,7 +14,27 @@ QUnit.test('moduleName option is passed into meta', function() {
   let moduleName = 'It ain\'t hard to tell';
   let template = compile('Hi, {{name}}!', {
     env,
-    moduleName
+    meta: {
+      moduleName
+    }
   });
   equal(template.raw.symbolTable.getMeta().moduleName, moduleName, 'Template has the moduleName');
+});
+
+QUnit.module('compileSpec', {
+  setup() {
+    env = new TestEnvironment();
+  }
+});
+
+QUnit.test('returned meta is correct', function() {
+  let wire = compileSpec('Hi, {{name}}!', {
+    meta: {
+      moduleName: 'my/module-name',
+      metaIsOpaque: 'yes'
+    }
+  });
+
+  equal(wire.meta.moduleName, 'my/module-name', 'Template has correct meta');
+  equal((<any>wire.meta).metaIsOpaque, 'yes', 'Template has correct meta');
 });
--- a/packages/glimmer-test-helpers/lib/helpers.ts
+++ b/packages/glimmer-test-helpers/lib/helpers.ts
@@ -76,7 +76,7 @@ export function compileRealSpec(string: string, options: TestCompileOptions): Se
 }
 
 export function template(templateSpec: TemplateSpec): SerializedTemplate {
-  return JSON.parse(templateSpec);
+  return JSON.parse(templateSpec.template);
 }
 
 export function equalInnerHTML(fragment, html, msg?) {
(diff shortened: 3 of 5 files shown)


packages/glimmer-compiler/lib/compiler.ts
```diff
@@ -19,7 +20,14 @@ export type TemplateSpec = string;
  * @return {TemplateSpec} A template spec string
  */
 export function compileSpec(string: string, options: any): TemplateSpec {
+  let meta: BlockMeta = {};
+  if (typeof options === 'object' && options.meta.moduleName && options.meta.moduleName) {
```
krisselden: it should treat meta as opaque, also, I think options is just meta, cc @rwjblue ?

krisselden: turns out meta isn't a key on options, options is passed as `{moduleName}`

rwjblue: We can absorb this in Ember though.  Can we do something like:

```js
let meta: BlockMeta = Object.assign({}, options.meta);
```

krisselden:
Also it would be cleaner to not have it in both places, and to have defined an interface for the result that is { BlockMeta and SerializedTemplate } that `TemplateCompiler.compile` returns

asakusuma:
@krisselden have what in both places?

I agree on the interface I will add this

krisselden:
the meta is also going into the SerializedTemplate as well?

asakusuma:
Oh true I did not think about that

asakusuma:
Updated based on comments.

@krisselden any ideas on what to do about duplicate meta? I mean they are needed in both places so not sure there's much we can do here.

krisselden:
can it not be passed into to Template?

krisselden:
https://github.com/tildeio/glimmer/blob/078009646b6838474aa2268f332721570153db8e/packages/glimmer-runtime/lib/template.ts#L23

asakusuma:
@krisselden well, as is, the meta is being passed to [here](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-compiler/lib/javascript-compiler.ts#L73) via the [`TemplateCompiler.complile` call](https://github.com/tildeio/glimmer/pull/298/files#diff-b1fa6c1a8a598bfb7a357a1329071ab9L23)

rwjblue:
I do not fully understand the motivation here.  Can you explain the usecase / problem that this solves?

packages/glimmer-compiler/lib/compiler.ts
```diff
@@ -19,7 +25,15 @@ export type TemplateSpec = string;
  * @return {TemplateSpec} A template spec string
  */
 export function compileSpec(string: string, options: any): TemplateSpec {
+  let meta: BlockMeta = {};
+  if (typeof options === 'object') {
+    meta = Object.assign(meta, options.meta);
```
rwjblue: I don't think we can use `Object.assign` here, sadly.  It is not supported on IE9 - IE11, or on Node 0.10 - 0.12.

wycats:
> I do not fully understand the motivation here. Can you explain the usecase / problem that this solves?

The basic idea is to make it possible to get the metadata for a template without parsing it.

rwjblue:
The implementation in Ember eagerly parses the JSON string though, which parsing step are you referring to?

packages/glimmer-compiler/tests/compile-options-test.ts
```diff
@@ -13,7 +14,27 @@ QUnit.test('moduleName option is passed into meta', function() {
...
+QUnit.test('returned meta is correct', function() {
+  let wire = compileSpec('Hi, {{name}}!', {
+    meta: {
+      moduleName: 'my/module-name',
+      metaIsOpaque: 'yes'
+    }
+  });
+
+  equal(JSON.parse(wire).meta.moduleName, 'my/module-name', 'Template has correct meta');
+  equal(JSON.parse(wire).meta.  metaIsOpaque, 'yes', 'Template has correct meta');
```
rwjblue: Bizarre spacing here

rwjblue:
> The implementation in Ember eagerly parses the JSON string though

I guess we are double stringifying to avoid that?  

Also, doesn't this bloat templates by storing the meta twice (plus the extra encoding required by `JSON.stringify(JSON.stringify(program))`)?

rwjblue:
Maybe it be better to just return an object of two string properties, and pass the meta back in to avoid it being duplicated?

```
export default Ember.HTMLBars.template({
  meta: "some JSON.strinfify'ed thing",
  template: "some JSON.strinfify'ed thing that doesn't include the meta"
})
```

Then on the Ember side we would pass it into the `Wrapper` on create. This avoids duplication of meta, and avoids `JSON.parse` in the full value eagerly...

krisselden:
@rwjblue one json stringify is just evaluated because the result of this is just concatenated into the module.  There should only be one string, the template string, the idea is to lazily parse it.

packages/glimmer-test-helpers/lib/helpers.ts
```diff
@@ -76,7 +76,7 @@ export function compileRealSpec(string: string, options: TestCompileOptions): Se
 }
 
 export function template(templateSpec: TemplateSpec): SerializedTemplate {
-  return JSON.parse(templateSpec);
+  return JSON.parse(JSON.parse(templateSpec).template);
```
krisselden: should not be parsing it twice here, and encoding it twice above.

packages/glimmer-compiler/lib/compiler.ts
```diff
@@ -19,7 +25,15 @@ export type TemplateSpec = string;
...
+  }
   let ast = preprocess(string, options);
   let program = TemplateCompiler.compile(options, ast);
-  return JSON.stringify(program);
+  let templateModule: TemplateModule = {
+    meta,
+    template: JSON.stringify(program)
+  };
+  return JSON.stringify(templateModule);
```
krisselden: we just want the meta moved outside the stringify here, not an extra stringify.

krisselden: it will be stringify on precompile for going into the module

krisselden: I think my pseudo code was confusing because I was explaining precompile, which currently results in a stringify(stringify(templateSpec)) but doing this in `compileSpec` results in 3 stringifies

asakusuma: @krisselden so you want the return type of this function changed from `String` to `Object`?

krisselden:
Yes, then precompile will turn it into a string for concatenation into module code

asakusuma:
@rwjblue I agree that having `meta` duplicated is less than ideal, but I think relying on the host (i.e. ember) to inject the meta is also not ideal?

@krisselden updated based on your comments

rwjblue:
Closing in favor of #302

PR #299, opened 2016-09-05
Pass caller's self to component manager's create.

rwjblue:
In Ember we are only using `dynamicScope` (since that is all that is exposed), but the current `{{this}}` (aka `self`)  is also needed to allow  us to properly target actions to their calling context.

The `dynamicScope` tracks the direct component heirarchy without regard to blocks/yielding etc. This is a perfect analogue to the way Ember tracks `.parentView` through a component structure.

The caller self tracks the current calling context. This means it is aware of blocks yielding (which changes the `self` in the yielded block), and is the correct analogue to the way Ember tracks the target of any actions within a template block.

For example, given:

```hbs
{{#component-a}}
  {{component-b bar="bar"}}
{{/component-a}}
```

From within `component-b` the `.parentView` (derived from `dynamicScope`) should match `component-a`, but the `targetObject` of `component-b` should not be `component-a` (since `component-a` is not calling it).

This change allows us to completely remove the manual `DynamicScope#targetObject` tracking in Ember.

---

I also considered changing `DynamicScope#child` to accept the `callerScope` which would have let us calculate the proper `targetObject` in Ember, but it seems that calling context is directly related to `ComponentManager#create` (and completely unrelated to `DynamicScope`).

### DIFF
--- a/packages/glimmer-demos/lib/visualizer.ts
+++ b/packages/glimmer-demos/lib/visualizer.ts
@@ -3,7 +3,6 @@ import {
   TestDynamicScope
 } from 'glimmer-test-helpers';
 import { UpdatableReference } from 'glimmer-object-reference';
-
 import { compileSpec } from 'glimmer-compiler';
 
 import { EvaluatedArgs } from 'glimmer-runtime';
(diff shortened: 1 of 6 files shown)


rwjblue:
The required changes in Ember (and a few tests) are in https://github.com/emberjs/ember.js/pull/14217.

packages/glimmer-demos/lib/visualizer.ts
```diff
@@ -372,7 +371,7 @@ function renderContent() {
     let definition = env.getComponentDefinition([component]);
 
     let manager = definition.manager;
-    let instance = manager.create(definition, EvaluatedArgs.empty(), new TestDynamicScope(), false);
+    let instance = manager.create(null, definition, EvaluatedArgs.empty(), new TestDynamicScope(), null, false);
```
chancancode: I think we can pass the env here (we closed over it)

rwjblue: Gotcha, updated!

PR #302, opened 2016-09-07
Refactor compileSpec to precompile

krisselden:
Refactors precompile, moving the laziness JSON parsing into glimmer from ember and the template factory.  Should make using precompiled templates from a standalone glimmer project easier too.

Also caches lazy json parsing per factory.


rwjblue:
This looks great! 

I'd love to see (at least a spike) of the changes required in Ember when we land this. If we used the `templateFactory` we should be able to remove most/all of the `ember-glimmer/template` code in , but I'm not sure how we would pass in the owner if we don't control create. Would Ember just continue to have a custom `template` function (not using the `templateFactory` added here), but remove the custom Wrapper class in favor of TemplateFactory?

krisselden:
@rwjblue the meta is opaque, and it has the environment, is that not good enough?

PR #303, opened 2016-09-07
Modifier install should use tree construction

chadhietala:


chadhietala:
We actually don't want this and instead buffer the modifiers and `install` them in `commit`

PR #305, opened 2016-09-07
[WIP] Cleanup destroyed components before inserting new ones into DOM.

rwjblue:
This supports the Ember lifecycle better, and seems more logical (considering that the destruction may actually want to operate on DOM to clean things up before new components are inserted).

### DIFF
--- a/packages/glimmer-runtime/lib/environment.ts
+++ b/packages/glimmer-runtime/lib/environment.ts
@@ -201,6 +201,10 @@ export abstract class Environment {
   }
 
   commit() {
+    for (let i=0; i<this.destructors.length; i++) {
+      this.destructors[i].destroy();
+    }
+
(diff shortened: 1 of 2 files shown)


chancancode:
@rwjblue I was looking at the Ember code yesterday and it seems to override `didDestroy` to make it synchronous (instead of queueing them), so this doesn't do anything without removing that patch. Have you tried removing it on the Ember side?

rwjblue:
We discussed this in more detail in slack. We need to expose the ability to have sync and async  destruction hooks to the managers. This would be conceptually similar to `create` which does sync instantiation and `didCreateElement` which fires `didInsertElement` and whatnot just before attaching to DOM.

krisselden:
@rwjblue @chancancode what's up with this?

PR #306, opened 2016-09-07
Modifier should be installed during commit

chadhietala:
/cc @rwjblue @chancancode

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
@@ -2,7 +2,7 @@ import { Opcode, OpcodeJSON, UpdatingOpcode } from '../../opcodes';
 import { VM, UpdatingVM } from '../../vm';
 import * as Simple from '../../dom/interfaces';
 import { FIX_REIFICATION } from '../../dom/interfaces';
-import { Environment, DynamicScope } from '../../environment';
+import { Environment } from '../../environment';
 import { FIXME, Option, Opaque, Dict, dict } from 'glimmer-util';
 import {
   CachedReference,
@@ -393,15 +393,9 @@ export class StaticAttrOpcode extends Opcode {
 
 export class ModifierOpcode extends Opcode {
   public type = "modifier";
-  public name: string;
-  public args: CompiledArgs;
-  private manager: ModifierManager<Opaque>;
 
-  constructor({ name, manager, args }: { name: string, manager: ModifierManager<Opaque>, args: CompiledArgs }) {
+  constructor(private name: string, private manager: ModifierManager<Opaque>, private args: CompiledArgs) {
     super();
-    this.name = name;
-    this.manager = manager;
-    this.args = args;
   }
 
   evaluate(vm: VM) {
@@ -410,21 +404,20 @@ export class ModifierOpcode extends Opcode {
     let { constructing: element, updateOperations } = stack;
     let args = this.args.evaluate(vm);
     let dynamicScope = vm.dynamicScope();
+    let modifier = manager.create(element as FIX_REIFICATION<Element>, args, dynamicScope, updateOperations);
 
-    let modifier = manager.install(element as FIX_REIFICATION<Element>, args, updateOperations, dynamicScope);
+    vm.env.scheduleInstallModifier(modifier, manager);
     let destructor = manager.getDestructor(modifier);
 
     if (destructor) {
       vm.newDestroyable(destructor);
     }
 
-    vm.updateWith(new UpdateModifierOpcode({
+    vm.updateWith(new UpdateModifierOpcode(
       manager,
       modifier,
-      element: element as FIX_REIFICATION<Element>,
-      dynamicScope,
       args
-    }));
+    ));
   }
 
   toJSON(): OpcodeJSON {
@@ -442,31 +435,24 @@ export class ModifierOpcode extends Opcode {
 
 export class UpdateModifierOpcode extends UpdatingOpcode {
   public type = "update-modifier";
-
-  private element: Element;
-  private dynamicScope: DynamicScope;
-  private args: EvaluatedArgs;
-  private manager: ModifierManager<Opaque>;
-  private modifier: Opaque;
   private lastUpdated: Revision;
 
-  constructor({ manager, modifier, element, dynamicScope, args }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element, dynamicScope: DynamicScope, args: EvaluatedArgs }) {
+  constructor(
+    private manager: ModifierManager<Opaque>,
+    private modifier: Opaque,
+    private args: EvaluatedArgs
+  ) {
     super();
-    this.modifier = modifier;
-    this.manager = manager;
-    this.element = element;
-    this.dynamicScope = dynamicScope;
-    this.args = args;
     this.tag = args.tag;
     this.lastUpdated = args.tag.value();
   }
 
   evaluate(vm: UpdatingVM) {
-    let { manager, modifier, element, dynamicScope, args, lastUpdated } = this;
+    let { manager, modifier, tag, lastUpdated } = this;
 
-    if (!args.tag.validate(lastUpdated)) {
-      manager.update(modifier, element, args, vm.dom, dynamicScope);
-      this.lastUpdated = args.tag.value();
+    if (!tag.validate(lastUpdated)) {
+      vm.env.scheduleUpdateModifier(modifier, manager);
+      this.lastUpdated = tag.value();
     }
   }
 
--- a/packages/glimmer-runtime/lib/environment.ts
+++ b/packages/glimmer-runtime/lib/environment.ts
@@ -119,6 +119,10 @@ export class Scope {
 export abstract class Environment {
   protected updateOperations: DOMChanges;
   protected appendOperations: DOMTreeConstruction;
+  private scheduledInstallManagers: ModifierManager<Opaque>[] = null;
+  private scheduledInstallModifiers: Object[] = null;
+  private scheduledUpdateModifierManagers: ModifierManager<Opaque>[] = null;
+  private scheduledUpdateModifiers: Object[] = null;
   private createdComponents: Component[] = null;
   private createdManagers: ComponentManager<Component>[] = null;
   private updatedComponents: Component[] = null;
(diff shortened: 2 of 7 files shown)


packages/glimmer-runtime/lib/environment.ts
```diff
@@ -116,9 +116,15 @@ export class Scope {
   }
 }
 
+interface ModifierPair {
```
chadhietala: Not sure if it's better to alloc an object to hold the pair or to create 2 arrays to hold each separately.

packages/glimmer-runtime/lib/environment.ts
```diff
@@ -116,9 +116,15 @@ export class Scope {
...
+  manager: ModifierManager<Opaque>;
+  element: Element;
+}
+
 export abstract class Environment {
   protected updateOperations: DOMChanges;
   protected appendOperations: DOMTreeConstruction;
+  private createdModifierManagers: ModifierPair[] = [];
```
chadhietala: It appears we need to set this to `[]` instead of `null` do to where modifiers are created during evaluation.

packages/glimmer-runtime/lib/environment.ts
chancancode: we should use two arrays like we do for components; in addition to the cost of allocating the object, it has some pretty bad issues with GC driven deopts

packages/glimmer-runtime/lib/environment.ts
chancancode: @chadhietala we do it in `env.begin()` below

packages/glimmer-runtime/lib/modifier/interfaces.ts
```diff
@@ -4,15 +4,16 @@ import { DynamicScope } from '../environment';
 import { Destroyable } from 'glimmer-util';
 
 export interface ModifierManager<T> {
+  create(args: EvaluatedArgs, dynamicScope: DynamicScope, dom: DOMChanges): T;
   // At initial render, the modifier gets a chance to install itself on the
   // element it is managing. It can also return a bucket of state that
   // it could use at update time. From the perspective of Glimmer, this
   // is an opaque token.
-  install(element: Element, args: EvaluatedArgs, dom: DOMChanges, dynamicScope: DynamicScope): T;
+  install(element: Element);
```
chancancode: Need to pass the `T` here, also I _think_ we want to pass `dom: DOMChanges` here instead of the constructor (we hypothesized some scenarios where you might get a different thing each time, e.g. RehydrationDOMChanges -> DOMChanges)

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -42,7 +42,9 @@ function commonSetup() {
 function render(template: Template, context = {}, view: PathReference<Opaque> = null) {
   let options = { appendTo: root, dynamicScope: new TestDynamicScope() };
   self = new UpdatableReference(context);
+  env.begin();
   result = template.render(self, env, options);
+  env.commit();
```
chancancode: good catch! (that means we are not testing any async stuff in this file 😩)

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -407,11 +407,13 @@ export class ModifierOpcode extends Opcode {
...
     let args = this.args.evaluate(vm);
     let dynamicScope = vm.dynamicScope();
+    let modifier = manager.create(args, dynamicScope);
+
+    vm.env.didCreateModifier(modifier, manager, element as FIX_REIFICATION<Element>);
 
-    let modifier = manager.install(element as FIX_REIFICATION<Element>, args, updateOperations, dynamicScope);
     let destructor = manager.getDestructor(modifier);
 
     if (destructor) {
```
chancancode: Can't comment below this line, but we don't need pass the args and dynamicScope to the updating opcode anymore

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -442,30 +441,29 @@ export class ModifierOpcode extends Opcode {
...
+  constructor({ manager, modifier, element }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element }) {
     super();
     this.modifier = modifier;
     this.manager = manager;
     this.element = element;
-    this.dynamicScope = dynamicScope;
-    this.args = args;
-    this.tag = args.tag;
-    this.lastUpdated = args.tag.value();
+    this.dynamicScope = manager.dynamicScope;
```
chancancode: ✂️

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -442,30 +441,29 @@ export class ModifierOpcode extends Opcode {
...
+    this.args = manager.args;
```
chancancode: ✂️

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -442,30 +441,29 @@ export class ModifierOpcode extends Opcode {
...
+    this.lastUpdated = this.args.tag.value();
   }
 
   evaluate(vm: UpdatingVM) {
-    let { manager, modifier, element, dynamicScope, args, lastUpdated } = this;
+    let { manager, modifier, element, args, lastUpdated } = this;
 
     if (!args.tag.validate(lastUpdated)) {
-      manager.update(modifier, element, args, vm.dom, dynamicScope);
+      manager.update(modifier, element, vm.env.getDOM());
```
chancancode: This should probably be async for consistency – the mental model is all modifiers run _after_ component hooks.

So this line should just be `env.didUpdateModifier` (maybe the name should be a `installModifier` and `updateModifier` – because all modifiers are async, and we didn't actually apply the update here)

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -442,30 +441,29 @@ export class ModifierOpcode extends Opcode {
...
-
   private element: Element;
   private dynamicScope: DynamicScope;
   private args: EvaluatedArgs;
   private manager: ModifierManager<Opaque>;
   private modifier: Opaque;
   private lastUpdated: Revision;
 
-  constructor({ manager, modifier, element, dynamicScope, args }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element, dynamicScope: DynamicScope, args: EvaluatedArgs }) {
+  constructor({ manager, modifier, element }: { manager: ModifierManager<Opaque>, modifier: Opaque, element: Element }) {
```
chancancode: Can you change this to positional args while you are at it? We have been doing that whenever we touch existing code.

packages/glimmer-runtime/lib/modifier/interfaces.ts
```diff
@@ -4,15 +4,18 @@ import { DynamicScope } from '../environment';
 import { Destroyable } from 'glimmer-util';
 
 export interface ModifierManager<T> {
+  args: EvaluatedArgs;
+  dynamicScope: DynamicScope;
```
chancancode: This is incorrect. If you need them, you could save it inside the `T` you are creating, but the manager shouldn't have them. The manager is usually a singleton. (i.e. just ✂️  these two lines)

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -623,11 +623,15 @@ class HelperReference implements PathReference<Opaque> {
 }
 
 class InertModifierManager implements ModifierManager<Opaque> {
-  install(element: Element, args: EvaluatedArgs, dom: IDOMChanges): Opaque {
+  args: EvaluatedArgs;
+  dynamicScope: DynamicScope;
```
chancancode: ✂️

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -623,11 +623,15 @@ class HelperReference implements PathReference<Opaque> {
...
+  create() {}
+
+  install(element: Element) {
```
chancancode: `modifier: Opaque, element: Element`

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -623,11 +623,15 @@ class HelperReference implements PathReference<Opaque> {
...
     return;
```
chancancode: ✂️

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -623,11 +623,15 @@ class HelperReference implements PathReference<Opaque> {
...
   }
 
-  update(modifier: Opaque, element: Element, args: EvaluatedArgs, dom: IDOMChanges) {
+  update(modifier: Opaque, element: Element) {
     return;
```
chancancode: ✂️

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -636,59 +640,47 @@ class InertModifierManager implements ModifierManager<Opaque> {
...
 }
 
 export class TestModifierManager implements ModifierManager<TestModifier> {
-  public installedElements: Element[];
-  public updatedElements: Element[];
-  public destroyedModifiers: TestModifier[];
+  public installedElements: Element[] = [];
+  public updatedElements: Element[] = [];
+  public destroyedModifiers: TestModifier[] = [];
+  public args: EvaluatedArgs;
```
chancancode: ✂️

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -636,59 +640,47 @@ class InertModifierManager implements ModifierManager<Opaque> {
...
+  public dynamicScope: DynamicScope;
```
chancancode: ✂️

packages/glimmer-runtime/lib/modifier/interfaces.ts
```diff
@@ -4,15 +4,18 @@ import { DynamicScope } from '../environment';
...
+  create(args: EvaluatedArgs, dynamicScope: DynamicScope): T;
```
chancancode: Actaully, since the element cannot change, we could pass the element to `create` instead, then we don't have pass them to `install` and `update` (consequently, you won't have to buffer the `element` on the env).

PR #307, opened 2016-09-08
Fix {{this.foo}} !== {{foo}}

chancancode:
The handlebars parser simply drops the `this` for all path expressions, but this is incorrect in Glimmer since we have the concept of local variables (block params), so `{{this.foo}}` is semantically different from `{{foo}}` (the former is always a “self lookup”, whereas the latter might be a local variable lookup).

This fundamentally fixes the problem by adding `this` back into the path expressions and allow the runtime to triage it. This also removes the need for the “self-get” expression we previously added to address this issue.


packages/glimmer-syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -140,6 +140,10 @@ export default {
 
     delete path.depth;
 
+    if (original.match(/^this(\..+)?$/)) {
+      path.parts.unshift('this');
```
mmun: The convention in Handlebars is to use an empty string internally for `this` so that you can differentiate it between `{{[this]}}`.

packages/glimmer-test-helpers/lib/environment.ts
```diff
@@ -800,7 +800,7 @@ export class TestEnvironment extends Environment {
       }
     }
 
-    if ((!isSimple && appendType === 'unknown') || appendType === 'self-get') {
```
mmun: What was happening here?

chancancode: This to turn "anything with a dot" to possibly a component invocation because closure components. Previously self-get is different but now they actually _do_ have a dot (parts.length > 1) in them.

mmun:
LGTM :+1:

PR #308, opened 2016-09-08
Fix `assign` utility to skip `null` / `undefined`.

rwjblue:
`Object.assign` ignores `null` / `undefined` arguments, but our internal `assign` does not (prior to these changes it errored if it received `null` / `undefined`).

This now handles it like native `Object.assign`:

screenshot of Object.assign from Chrome


rwjblue:
I stumbled across this when working on updating Ember with the precompilation changes from #302.  If Ember happens to pass no `meta` in options (which should be a valid value since its completely opaque from glimmer's perspective) the call to `assign` would throw an error due to `Object.keys(undefined)`.

PR #311, opened 2016-09-09
Prevent setting properties to an initial value of `undefined`.

rwjblue:
This matches expected Ember and HTMLBars behavior. See the following for where this was implemented in HTMLBars:

* https://github.com/tildeio/htmlbars/blob/master/packages/dom-helper/lib/prop.js#L1-L3
* https://github.com/tildeio/htmlbars/blob/master/packages/morph-attr/lib/main.js#L11

Addresses https://github.com/emberjs/ember.js/issues/14243.


rwjblue:
Per conversation in slack:

Updated property tests to add a "reset" assertion along with a `TODO` comment that the semantics for "resetting" a property may not be correct.

rwjblue:
Updated to use paired assertions in the tests added (i.e. `<div data-foo={{isNull}} /><div data-foo={{isNotNull}} />`).

PR #313, opened 2016-09-12
Lock to typescript@2.1.0-dev.20160911.

rwjblue:
Using typescript@2.1.0-dev.20160912 leads to the following error during
`npm run vscode-build`:

```
packages/glimmer-runtime/lib/compiled/opcodes/content.ts(258,82): error TS2345: Argument of type 'string | T' is not assignable to parameter of type 'T'.
  Type 'string' is not assignable to type 'T'.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "qunit": "^0.7.2",
     "simple-dom": "^0.3.0",
     "simple-html-tokenizer": "^0.2.5",
-    "typescript": "next"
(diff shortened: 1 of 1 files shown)


rwjblue:
Looks like @chadhietala has a fix pending for this in https://github.com/tildeio/glimmer/pull/312/files#diff-4058acbf46527ae6d58d66587448612bR258.  If that is ready to land, we should do that instead of this...

PR #314, opened 2016-09-15
Remove POJO args

chadhietala:
To keep garbage down through the runtime, this removes majority of the usage of pojo args for creating these short lived intermediate objects that are simply used to collect up arguments that are very quickly spread out at fields onto a class.

I need to look at the DSL builder and see if it's worth doing there as well. Seems like there is some maintenance value there.

/cc @krisselden, @chancancode, @wycats


packages/glimmer-runtime/lib/dom/interface.ts
```diff
@@ -1,50 +0,0 @@
-export type Namespace =
```
chadhietala: This was a dup

PR #316, opened 2016-09-15
Fix Partials

chadhietala:


packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -366,7 +378,23 @@ function assertValidYield({ hash }): string {
   }
 }
 
-function assertValidHasBlock({ params }): string {
+function assertValidPartial({ params, hash, escaped }) /* : expr */ {
+  if (params && params.length !== 1) {
+    throw new Error(`Partial found with no arguments. You must specify a template name.`);
```
rwjblue: Needs to include location information, without it the source of this error will very hard to track down in large apps.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -366,7 +378,23 @@ function assertValidYield({ hash }): string {
...
+  } else if (hash && hash.pairs.length > 0) {
+    throw new Error(`partial does not take any named arguments`);
```
rwjblue: Needs to include location information, without it the source of this error will very hard to track down in large apps.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -366,7 +378,23 @@ function assertValidYield({ hash }): string {
...
+  } else if (!escaped) {
+    throw new Error(`{{{partial ...}}} is not supported, please use {{partial ...}} instead`);
```
rwjblue: Needs to include location information, without it the source of this error will very hard to track down in large apps.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -380,7 +408,11 @@ function assertValidHasBlock({ params }): string {
   }
 }
 
-function assertValidHasBlockParams({ params }): string {
+function assertValidHasBlockParams({ params, hash }): string {
+  if (hash && hash.pairs.length > 0) {
+    throw new Error(`has-block-params does not take any named arguments`);
```
rwjblue: Needs to include location information, without it the source of this error will very hard to track down in large apps.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -366,7 +378,23 @@ function assertValidYield({ hash }): string {
...
+  } else if (!escaped) {
+    throw new Error(`{{{partial ...}}} is not supported, please use {{partial ...}} instead`);
+  }
+
+  return params;
+}
+
+function assertValidHasBlock({ params, hash }): string {
+  if (hash && hash.pairs.length > 0) {
+    throw new Error(`has-block does not take any named arguments`);
```
rwjblue: Needs to include location information, without it the source of this error will very hard to track down in large apps.

packages/glimmer-compiler/lib/template-compiler.ts
```diff
@@ -366,7 +378,23 @@ function assertValidYield({ hash }): string {
...
+  } else if (hash && hash.pairs.length > 0) {
+    throw new Error(`partial does not take any named arguments`);
+  } else if (!escaped) {
+    throw new Error(`{{{partial ...}}} is not supported, please use {{partial ...}} instead`);
+  }
+
+  return params;
+}
+
+function assertValidHasBlock({ params, hash }): string {
```
rwjblue: This seems identical to the `assertValidHasBlockParams` function just below.  What am I missing?

packages/glimmer-node/tests/node-test.ts
```diff
@@ -246,8 +250,8 @@ QUnit.module('default template id');
 QUnit.test('generates id in node', function (assert) {
   let template = precompile('hello');
   let obj = JSON.parse(template);
-  assert.equal(obj.id, 'tJmwGiDc', 'short sha of template source');
+  assert.equal(obj.id, 'tjhOuBK+', 'short sha of template source');
```
rwjblue: Why are these changing?  Is this related to the partial work here?

packages/glimmer-runtime/lib/vm/frame.ts
```diff
@@ -4,19 +4,19 @@ import { InlineBlock } from '../compiled/blocks';
...
 import { Opcode, OpSeq } from '../opcodes';
 import { LabelOpcode } from '../compiled/opcodes/vm';
-import { Component, ComponentManager, ComponentDefinition } from '../component/interfaces';
+import { Component, ComponentManager } from '../component/interfaces';
 
 class Frame {
   ops: OpSeq;
   op: Opcode;
   operand: PathReference<any> = null;
+  immediate: any = null;
```
rwjblue: Its not _immediately_ obvious what `immediate` means here (pun intended), maybe an inline comment or more descriptive name?

packages/glimmer-node/tests/node-test.ts
chadhietala: The compiled output now has this `hasPartials` flag in in it.

packages/glimmer-runtime/lib/vm/frame.ts
chadhietala: paging @chancancode

chadhietala:
I believe I have addressed the all of the concerns. I'm shepherding for @chancancode so he will need to comment on the `immediate` stuff.

packages/glimmer-runtime/lib/compiled/blocks.ts
```diff
@@ -60,15 +60,15 @@ export class EntryPoint extends TopLevelTemplate {
 }
 
 export class Layout extends TopLevelTemplate {
-  constructor(program: Program, symbolTable: SymbolTable, public named: string[], public yields: string[]) {
+  constructor(program: Program, symbolTable: SymbolTable, public named: string[], public yields: string[], public hasPartials: boolean) {
     super(program, symbolTable);
   }
 
-  hasNamedParameters(): boolean {
+  get hasNamedParameters(): boolean {
```
chancancode: Actually, since these objects are immutable, should we just assign to these properties in the constructor?

packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -59,6 +59,14 @@ class StatementCompilationBufferProxy implements StatementCompilationBuffer {
...
     return this.inner.hasBlockSymbol(name);
   }
+
+  getPartialArgsSymbol(): number {
+    return this.inner.getPartialArgsSymbol();
+  }
+
+  hasPartialArgsSymbol(): boolean {
+    return this.inner.hasPartialArgsSymbol();
+  }
```
chancancode: I wanted to kill `SymbolLookup` and just use the symbol table, but ran out of time. See https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/helpers/debugger.js#L70-L74.

I believe the main reason for SymbolLookup to exist was that there use to be a "keyword" feature whose information came from the env. But now that the feature is gone, I think all this is doing is just delegating to the symbol table.

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -86,7 +86,7 @@ export class OpenDynamicPrimitiveElementOpcode extends Opcode {
   public type = "open-dynamic-primitive-element";
 
   evaluate(vm: VM) {
-    let tagName = vm.frame.getOperand().value();
+    let tagName = vm.frame.getOperand<string>().value();
```
chancancode: these are somewhat dubious, mighty be missing a `String()` cast

packages/glimmer-runtime/lib/compiled/opcodes/partial.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ReferenceCache, isConst, map } from 'glimmer-reference';
+import { Opcode, OpcodeJSON } from '../../opcodes';
+import { Assert } from './vm';
+import { VM } from '../../vm';
+import { PartialDefinition } from '../../partial';
+import SymbolTable from '../../symbol-table';
+import { PartialBlock } from '../blocks';
+
+export class PutDynamicPartialDefinitionOpcode extends Opcode {
+  public type = "put-dynamic-component-definition";
```
chancancode: oops

packages/glimmer-runtime/lib/environment.ts
```diff
@@ -258,20 +251,21 @@ export abstract class Environment {
...
-  abstract hasPartial(partialName: string[], symbolTable: SymbolTable): boolean;
-  abstract lookupPartial(PartialName: string[], symbolTable: SymbolTable): PartialDefinition<TemplateMeta>;
+  abstract hasHelper(helperName: string[], blockMeta: TemplateMeta): boolean;
+  abstract lookupHelper(helperName: string[], blockMeta: TemplateMeta): Helper;
+
+  abstract hasModifier(modifierName: string[], blockMeta: TemplateMeta): boolean;
+  abstract lookupModifier(modifierName: string[], blockMeta: TemplateMeta): ModifierManager<Opaque>;
+
   abstract hasComponentDefinition(tagName: string[], symbolTable: SymbolTable): boolean;
   abstract getComponentDefinition(tagName: string[], symbolTable: SymbolTable): ComponentDefinition<Opaque>;
```
chancancode: I want to change the rest of these (helpers, modifiers, components) from `string[]` to `string`. The only issue is we need to make `{{f.foo with some="args"}}` illegal in the Glimmer compiler, but Ember actually use that feature. The plan is the make those errors happen in the "JavaScript" compiler, so Ember would have a chance to transform it into `{{component f.foo some="args"}}` before it hits the error. Or we could just allow it for helpers only, but the inconsistency bugs me a little.

packages/glimmer-runtime/lib/syntax/core.ts
```diff
@@ -707,10 +730,15 @@ export class GetArgument<T> extends ExpressionSyntax<T> {
...
-    let path = parts.slice(1);
-    return new CompiledLocalLookup(symbol, path, head);
+    if (lookup.hasNamedSymbol(head)) {
+      let symbol = lookup.getNamedSymbol(head);
+      let path = parts.slice(1);
+      return new CompiledLocalLookup(symbol, path, head);
+    } else if (lookup.hasPartialArgsSymbol()) {
+      let symbol = lookup.getPartialArgsSymbol();
+      return new CompiledLocalLookup(symbol, parts, head);
+    }
```
chancancode: Probably should add an `} else { throw new Error("Compile error..."); }`.

chancancode: (or an assert)

packages/glimmer-node/tests/node-test.ts
rwjblue: Gotcha!

packages/glimmer-runtime/tests/partial-test.ts
```diff
@@ -141,6 +141,19 @@ QUnit.test('dynamic partial with static content', assert => {
   equalTokens(root, `Before <div>Testing</div> After`);
 });
 
+QUnit.test('nested dynamic partial with dynamic content', assert => {
```
rwjblue: 👍

chadhietala:
@chancancode gave 👍 in Slack

PR #317, opened 2016-09-16
Fix issue with components that have multiple yields.

rwjblue:
Previously, we assumed that only `{{#each` needed to deal with "multiple" yields, but in reality any component can also yield many times also.

This moves the `pushChildScope` / `popScope` up out of `EachSyntax` into the main `InlineBlockCompiler`.

In theory, we don't need to do this for *every* block with block params (i.e. `{{#with` will never need this), so we could move these operations to only be for `{{#each` and component invocation with block params.  I'm happy to work on that with direction on where the addition `pushChildScope` / `popScope` needs to go for component statements.

Related to https://github.com/emberjs/ember.js/issues/14284 and https://github.com/emberjs/ember.js/pull/14285.

### DIFF
--- a/packages/glimmer-runtime/lib/compiler.ts
+++ b/packages/glimmer-runtime/lib/compiler.ts
@@ -113,7 +113,10 @@ export class InlineBlockCompiler extends Compiler {
     let { block, ops } = this;
     let { program } = block;
 
-    if (block.hasPositionalParameters()) {
+    let hasPositionalParameters = block.hasPositionalParameters();
+
+    if (hasPositionalParameters) {
+      ops.pushChildScope();
       ops.bindPositionalArgsForBlock(block);
     }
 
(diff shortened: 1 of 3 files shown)


chancancode:
Looks good to me!

> In theory, we don't need to do this for every block with block params (i.e. {{#with will never need this), so we could move these operations to only be for {{#each and component invocation with block params. I'm happy to work on that with direction on where the addition pushChildScope / popScope needs to go for component statements.

`{{#with}}` is the only case we know for sure that doesn't need a child scope. In [here](https://github.com/tildeio/glimmer/blob/master/packages/glimmer-runtime/lib/syntax/builtins/with.ts#L43), `this.evaluate('default', /* reuseParentScope */ true)` and thread it all the way down to [here](https://github.com/tildeio/glimmer/blob/fbe5e0dbf12ca9388eff73eeb29767eee3fd904c/packages/glimmer-runtime/lib/compiler.ts#L116) (`if (!reuseParentScope && block.hasPositionalParameters)`). But it's not mega high priority and we can tackle that in another pass also.

PR #319, opened 2016-09-17
Fix props and attrs again

chadhietala:
This PR aligns props and attrs to how Ember thinks about them.

In the future we want to move away from "props first" but we need to write an RFC.

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/expressions/concat.ts
+++ b/packages/glimmer-runtime/lib/compiled/expressions/concat.ts
@@ -1,5 +1,5 @@
 import { CompiledExpression } from '../expressions';
-import { normalizeTextValue } from '../opcodes/content';
+import { FIXME } from 'glimmer-util';
 import VM from '../../vm/append';
 import { PathReference, CachedReference, RevisionTag, combineTagged } from 'glimmer-reference';
 import { Opaque } from 'glimmer-util';
@@ -31,10 +31,28 @@ class ConcatReference extends CachedReference<string> {
   }
 
   protected compute(): string {
-    let parts = new Array<string>(this.parts.length);
+    let parts = new Array<string>();
+
     for (let i = 0; i < this.parts.length; i++) {
-      parts[i] = normalizeTextValue(this.parts[i].value());
+      let value = this.parts[i].value();
+
+      if (value !== null && value !== undefined) {
+        parts[i] = castToString(this.parts[i].value()) as FIXME<string, 'Coerce falsy values to strings'>;
+      }
     }
-    return parts.join('');
+
+    if (parts.length > 0) {
+      return parts.join('');
+    }
+
+    return null;
   }
+}
+
+function castToString(value) {
+  if (typeof value['toString'] !== 'function') {
+    return '';
+  }
+
+  return String(value);
 }
--- a/packages/glimmer-runtime/lib/dom/attribute-managers.ts
+++ b/packages/glimmer-runtime/lib/dom/attribute-managers.ts
@@ -18,7 +18,7 @@ export interface AttributeManager {
 interface PropertyManager {
   setAttribute(env: Environment, element: Simple.Element, attr: string, value: Opaque, namespace?: string): void;
   updateAttribute(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
-  removeProperty(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
+  removeAttribute(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
 }
 
 export function defaultManagers(element: Simple.Element, attr: string, isTrusting: boolean, namespace: string) {
@@ -29,20 +29,16 @@ export function defaultManagers(element: Simple.Element, attr: string, isTrustin
     return defaultAttributeManagers(tagName, attr);
   }
 
-  let { type } = normalizeProperty(element, attr);
+  let { type, normalized } = normalizeProperty(element, attr);
 
   if (type === 'attr') {
-    return defaultAttributeManagers(tagName, attr);
+    return defaultAttributeManagers(tagName, normalized);
   } else {
-    return defaultPropertyManagers(tagName, attr);
+    return defaultPropertyManagers(tagName, normalized);
   }
 }
 
 export function defaultPropertyManagers(tagName: string, attr: string) {
-  if (attr === 'disabled' || attr === 'checked') {
-    return BooleanPropertyChangeList;
-  }
-
   if (requiresSanitization(tagName, attr)) {
     return SafeHrefPropertyManager;
   }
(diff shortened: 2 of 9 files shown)


rwjblue:
> We want to do attribute for everything and special case the property setting. 

We definitely can not change this without an RFC. Ember 2.9 (and likely 2.10) **must** match the 1.13 - 2.8 behavior.

packages/glimmer-runtime/lib/dom/change-lists.ts
```diff
@@ -57,38 +57,54 @@ export function defaultAttributeChangeLists(tagName: string, attr: string) {
...
+    return element[normalized];
+  }
 };
 
 export const PropertyChangeList: IChangeList = {
   setAttribute(env: Environment, element: Simple.Element, attr: string, value: Opaque, namespace?: DOMNamespace) {
     if (value !== null && value !== undefined) {
       let normalized = attr.toLowerCase();
-      element[normalized] = normalizePropertyValue(value); // TODO: This doesn't work
+      element[normalized] = value;
```
rwjblue: There are some assertions in Ember around `<input disabled>` that we need to add to the tests below.  

If a template contains any of the following:

* `<input disabled=''>`
* `<input disabled>`
* `<input disabled="true">`
* `<input disabled="false">`

The `disabled` property needs to result in `true`, but given the logic here ATM I believe it would not be (happy to be proven wrong by way of added tests for this below).

screenshot of DOM stuffs

However, if the template contains any of these we should end up with `false` (I believe the `null` and `undefined` cases are handled currently):

* `<input disabled={{somethingThatEqualsFalse}}>`
* `<input disabled={{somethingUndefined}}>`
* `<input disabled={{somethingNull}}>`

chadhietala: I fixed this and wrote a test for it.

packages/glimmer-runtime/tests/attributes-test.ts
```diff
@@ -78,6 +78,39 @@ test("helpers shadow self", () => {
   equalTokens(root, '<div data-test="hello"></div>');
 });
 
+test("disable updates properly", () => {
+  let template = compile('<input disabled={{enabled}} />');
```
rwjblue: I think we also need a test with this template: `compile('<input disabled>')`

rwjblue: Can you also add another test just like this one, only with it being quoted?  Basically `<input disabled="{{value}}">` with all the various rerender permutations below.

chadhietala: I wrote tests for these as well.

packages/glimmer-runtime/lib/dom/attribute-managers.ts
```diff
@@ -5,91 +5,140 @@ import {
...
+  setAttribute(env: Environment, element: Simple.Element, attr: string, value: Opaque) {
+    PropertyManager.setAttribute(env, element, attr, sanitizeAttributeValue(env, element, attr, value));
+  }
+
+  updateAttribute(env: Environment, element: Element, attr: string, value: Opaque) {
+    this.setAttribute(env, element, attr, value);
+  }
+};
+
+export const BooleanPropertyChangeList: AttributeManager = new class {
```
rwjblue: Should this be `*Manager` to fit with the others?

packages/glimmer-runtime/lib/dom/attribute-managers.ts
```diff
@@ -5,91 +5,140 @@ import {
...
-      element[normalized] = value;
+    if (shouldRemoveProperty(value)) {
+      this.removeProperty(env, element, attr, value, namespace);
+    }
+    this.setAttribute(env, element, attr, value, namespace);
+  }
+};
+
+function shouldRemoveProperty(value) {
+  return value === null || value === undefined || value === false;
```
rwjblue: In HTMLBars, we only considered `null` and `undefined` to be "removal values".

packages/glimmer-runtime/lib/dom/attribute-managers.ts
```diff
@@ -5,91 +5,140 @@ import {
...
+export interface AttributeManager {
   setAttribute(env: Environment, element: Simple.Element, attr: string, value: Opaque, namespace?: string): void;
   updateAttribute(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
 }
 
-export function defaultChangeLists(element: Simple.Element, attr: string, isTrusting: boolean, namespace: string) {
+interface PropertyManager {
+  setAttribute(env: Environment, element: Simple.Element, attr: string, value: Opaque, namespace?: string): void;
+  updateAttribute(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
+  removeProperty(env: Environment, element: Element, attr: string, value: Opaque, namespace?: string): void;
```
rwjblue: The naming here is a tad odd.  Can we share interfaces with `AttributeManager` (I can't TS for 💩 )?  Naming the extra method `removeProperty` but all the others `*Attribute` "feels" weird.

packages/glimmer-runtime/lib/compiled/expressions/concat.ts
```diff
@@ -31,10 +31,28 @@ class ConcatReference extends CachedReference<string> {
...
+      return parts.join('');
+    }
+
+    return null;
   }
 }
+
+function castToString(value) {
+  if (typeof value['toString'] !== 'function') {
+    return '';
```
rwjblue: So if there is no `toString` we always use `''`? I guess I'd expect `Object.prototype.toString(value)` or something here.

packages/glimmer-runtime/tests/updating-test.ts
```diff
@@ -1371,8 +1371,8 @@ test("<option selected> is normalized and updated correctly", function() {
   let expectedInitialTokens = `
     <select multiple="">
       <option>0</option>
-      <option selected="">1</option>
-      <option selected="">2</option>
+      <option>1</option>
```
rwjblue: Changing these seem wrong, it implies that the options are not selected when they should be (and this was the point of the test originally).

packages/glimmer-runtime/lib/compiled/expressions/concat.ts
```diff
@@ -31,10 +31,28 @@ class ConcatReference extends CachedReference<string> {
   }
 
   protected compute(): string {
-    let parts = new Array<string>(this.parts.length);
+    let parts = new Array<string>();
```
cibernox: Maybe this array could have fixed size? I think I remember reading that making arrays grow deoptimizes.

chadhietala: We can't because we potentially are taking a subset.

cibernox: Yes, but at the end you join the elements with an empty string, so it doesn't matter if there is holes. Actually your implementation might also have holes.

packages/glimmer-runtime/tests/updating-test.ts
chadhietala: If you look at what we output for Ember 2.8 these are the correct semantics.

packages/glimmer-runtime/lib/compiled/expressions/concat.ts
```diff
@@ -31,10 +31,28 @@ class ConcatReference extends CachedReference<string> {
...
+
     for (let i = 0; i < this.parts.length; i++) {
-      parts[i] = normalizeTextValue(this.parts[i].value());
+      let value = this.parts[i].value();
+
+      if (value !== null && value !== undefined) {
+        parts[i] = castToString(this.parts[i].value()) as FIXME<string, 'Coerce falsy values to strings'>;
```
chancancode: IIUC, if we didn't pre-size it (which has a shaping problem, as we discovered recently), we should use push to avoid going OOB, is that correct? /cc @krisselden

PR #321, opened 2016-09-20
Refactor to save off normalized attr/prop name

chadhietala:
This is a part of #283 that we actually need to fix issues we see in Ember with nomalization. This change is to create an instance of the managers that know about the normalized attr/prop name.


chadhietala:
I think we can revisit in the future. The specialized managers are still singletons.

PR #323, opened 2016-09-21
Fix issue with `<img src={{foo}}>`.

rwjblue:
Due to recent changes, the attribute managers were updated to store the attribute names they reference to (so that normalized attribute names are used when setting / updating values).

A few singleton attribute managers were kept around, because we believed that they were always the same attribute so there was no need to create additional instances (seemed wasteful).

Unfortunately, the SafeHrefAttributeManager and SafeHrefPropertyManager are used for much more than just `href` (see full listing in `glimmer-runtime/lib/dom/sanitized-value.ts`) so we cannot make the assumption that all the attributes/properties are the same (and therefore cannot use a singleton).

This updates to remove usage of singleton for the sanitized values, and adds a number of tests around ensuring they are properly handled in various cases.

### DIFF
--- a/packages/glimmer-runtime/index.ts
+++ b/packages/glimmer-runtime/index.ts
@@ -65,8 +65,6 @@ export {
   AttributeManager as IAttributeManager,
   AttributeManager,
   PropertyManager,
-  SAFE_HREF_ATTRIBUTE_MANAGER,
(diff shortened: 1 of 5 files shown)


chadhietala:
LGTM

PR #331, opened 2016-09-25
Implement `{{#-in-element` API.

rwjblue:
This API (which is not used by Glimmer itself) is intended to allow the host environment to provide "wormhole" like behavior.

Specifically, when invoked `{{#-in-element` accepts an Element (that conforms to the `Simple.Element` interface) and renders the block provided into that element.

* The operand (first argument) can be changed to another element and the
  contents are cleared and recreated in the new location.
* Changing the operand to a "falsey" value, will clear the contents and
  skip rendering the block.
* Causing the `in-element` block to be removed (i.e. when inside an
  `{{#if`) the contents that were written to the foreign element are
  cleared.

### DIFF
--- a/packages/glimmer-runtime/lib/builder.ts
+++ b/packages/glimmer-runtime/lib/builder.ts
@@ -27,11 +27,7 @@ export interface LastNode {
 }
 
 class First {
-  private node: Node;
-
-  constructor(node) {
-    this.node = node;
-  }
+  constructor(private node: Node) { }
 
   firstNode(): Node {
     return this.node;
@@ -39,11 +35,7 @@ class First {
 }
 
 class Last {
-  private node: Node;
-
-  constructor(node) {
-    this.node = node;
-  }
+  constructor(private node: Node) { }
 
   lastNode(): Node {
     return this.node;
@@ -151,12 +143,15 @@ export class ElementStack implements Cursor {
     return tracker;
   }
 
-  private pushBlockTracker(tracker: Tracker) {
+  private pushBlockTracker(tracker: Tracker, isRemote = false) {
     let current = this.blockStack.current;
 
     if (current !== null) {
       current.newDestroyable(tracker);
-      current.newBounds(tracker);
+
+      if (!isRemote) {
+        current.newBounds(tracker);
+      }
     }
 
     this.blockStack.push(tracker);
@@ -207,9 +202,13 @@ export class ElementStack implements Cursor {
   pushRemoteElement(element: Simple.Element) {
     this.pushElement(element);
 
-    let tracker = new SimpleBlockTracker(this.element);
-    this.pushBlockTracker(tracker);
-    return tracker;
+    let tracker = new RemoteBlockTracker(element);
+    this.pushBlockTracker(tracker, true);
+  }
+
+  popRemoteElement() {
+    this.popBlock();
+    this.popElement();
   }
 
   private pushElement(element: Simple.Element) {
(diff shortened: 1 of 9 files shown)


packages/glimmer-runtime/lib/builder.ts
```diff
@@ -193,16 +188,35 @@ export class ElementStack implements Cursor {
...
+    this.popBlock();
+    this.popElement();
+  }
+
+  private pushElement(element: Simple.Element) {
+    this.element = element;
     this.elementStack.push(element);
+
+    this.nextSibling = null;
     this.nextSiblingStack.push(null);
```
chancancode: we can make this configurable for bonus points, which would allow inserting into the middle of an element :wink:

but we can do it as a follow-up.

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -62,6 +63,42 @@ export class OpenPrimitiveElementOpcode extends Opcode {
...
 
+export class PushRemoteElementOpcode extends Opcode {
+  public type = "push-remote-element";
+
+  evaluate(vm: VM) {
+    // This is asserting that the operand is an Element.
+    // But how do you know?
+
+    // {{#-with-element "zomg"}} <-- you'll get a string back here, now what?
+    let reference = vm.frame.getOperand<Simple.Element>();
```
chancancode: For future readers: this is unsafe but we decided to punt on error-ing since this is still experimental

packages/glimmer-runtime/tests/in-element-test.ts
```diff
@@ -0,0 +1,414 @@
...
+  set(view, 'foo', 'Double Yips!');
+  rerender();
+
+  equalsElement(externalElement, 'div', {}, `${initialContent}[Double Yips!]`);
+
+  set(view, 'foo', 'Yippie!');
+  rerender();
+
+  equalsElement(externalElement, 'div', {}, `${initialContent}[Yippie!]`);
+});
```
chancancode: Can we test setting the element to `null` as part of the steps here to ensure we aren't removing the initial content?

packages/glimmer-runtime/tests/in-element-test.ts
```diff
@@ -0,0 +1,414 @@
...
+        {{#-in-element externalElement}}[{{destroy-me}}]{{/-in-element}}
+      {{/if}}
+    `,
+    {
+      externalElement,
+      showExternal: false,
+    }
+  );
+
+  equalsElement(externalElement, 'div', {}, stripTight``);
```
chancancode: assert destroyed === 0

packages/glimmer-runtime/tests/in-element-test.ts
```diff
@@ -0,0 +1,414 @@
...
+
+  set(view, 'showExternal', true);
+  rerender();
```
chancancode: assert destroyed === 0

chancancode:
👏

packages/glimmer-runtime/lib/compiled/opcodes/dom.ts
chancancode: Should probably remove those comments tho :P

rwjblue:
Updated based on review feedback. Should be ready for another review...

lukemelia:
Very excited to have a supported API for wormhole and similar behavior! Thanks guys!

PR #332, opened 2016-09-26
Expose dynamicScope to prepareArgs hook

ef4:
This allows a component manager to choose to consume some arguments from the dynamic scope.


rwjblue:
Seems good to me, @chancancode r?

chancancode:
Seems good! Can you add a smoke test to execrise it? You can probably do something similar to the positional args smoke test (I hope we have one!)

ef4:
I added a test by implementing a simple dynamic-scope-dependent feature in EmberishCurlyComponent.

rwjblue:
Test looks good.

@chancancode - Anything else needed here?

chancancode:
@rwjblue @ef4 merged it so it doesn't bit rot, however @wycats told me he has plans to rethink this part of the system and maybe remove the dynamic scope feature and replace it with something else (o_0)

PR #333, opened 2016-10-13
Fix pruning and element availability in willDestroyElement

chadhietala:


rwjblue:
:+1: - Do we know how much this will negatively impact performance? I suspect that it won't be much on initial render, so it may be fine...

krisselden:
@rwjblue it was naming confusion, I realize I see you swapped the order, so didDestroy is not actually didDestroy but this was only on the RenderResult.  This does recurse twice and requires changes on the Ember side, I realize we can just move the destroy() before clear() in the other spots to match RenderResult

rwjblue:
This seems good to me, do you have the Ember side of changes ready and tested against this?

chadhietala:
@rwjblue I'm looking at the Ember side this morning. I'm not confident about the test failures I'm seeing and need to confirm with @krisselden that the failures are expected or not.

PR #335, opened 2016-10-19
Ensure local variables has precedence over helpers

chancancode:


rwjblue:
LGTM

chancancode:
There is a problem with this – this only consider the current level. I need to make it recursively walk up the block stack as well.

PR #337, opened 2016-10-25
Fix timing of expression evaluation in blocks

### LINKED ISSUE #336: stateful helper bug
- [x] if
- [x] unless
- [x] with
- [x] each
    - [ ] ...PutIterable?
- [ ] component
- [ ] partial
- [ ] in-element
- [ ] dot path deopt?

chancancode:
This commit fixes a compiler bug where certain built-in syntaxes would re-evaluate its arguments when the block's assertion causes a re-entry into the append VM, instead of reusing the argument references that were created the first time.

In practice, this is observable when a "stateful helper" is used as an argument to a syntax with an assertion (currently `#if`, `#unless`, `#with`, `#each`, `partial`, `InElementSyntax` and dynamic components).

Because the argument references are created every time, any state stored on the original instance of the helper will be observably wiped away when used as an argument to an asserting block helper. Of course, it is also observable by comparing the JavaScript identity of the helper in question over time.

This commit fixes the compiler bug by moving the evaluation of the expressions passed as arguments to the block syntaxes in question above the `Try` block, and capturing the relevant registers (args and operand, which store the results of evaluating the arguments) so they can be restored when the append VM is re-entered for the same block.

Fixes #336

### DIFF
--- a/packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
+++ b/packages/glimmer-runtime/lib/compiled/opcodes/builder.ts
@@ -1,4 +1,5 @@
 import * as component from './component';
+import * as partial from '../../compiled/opcodes/partial';
 import * as content from './content';
 import * as dom from './dom';
 import * as lists from './lists';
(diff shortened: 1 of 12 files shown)


chancancode:
TODOs:

- [ ] Investigate ember failure
- [ ] Move `Test` opcode outside of the `Enter` block?

PR #340, opened 2016-11-07
Include Handlebars comments in the AST.

rwjblue:
Still need to:

- [x] Add support for `<div {{! foo }}></div>` (adding a
  `elementComments` array to `ElementNode`).
- [x] Tests for `<!-- foo ->` loc info.
- [x] Support and tests for visiting `MustacheCommentStatements` in AST visitors.


rwjblue:
OK, this is ready for review.  I believe I have covered our bases here, but would love help confirming...

bcardarella:
I'm curious to know why hbs comments are disallowed for attribute values. Allowing them is actually very helpful for workflow on teams. For example, when our team gets handed off a design the UX dev who is responsible for the markup and styles will write the `hbs` templates. He'll leave `{{! TODO: add something here}}` comments throughout the template for our engineers to receive and implement. In the event of element attributes the comments were helpful to imply intent of what the values should be based upon the design spec.

Is there a reason other than "let's avoid footguns in production" to disallow comments for attr values?

mmun:
@bcardarella There is discussion about it https://github.com/emberjs/ember.js/issues/14652.

PR #343, opened 2016-11-09
Ensure removal values are respected for safe attr/prop.

rwjblue:
Prior to this change it was not possible to remove one of our sanitized attributes/properties, because we always used the `setAttribute` logic (which explicitly noop's initial sets when a removal value is being used).

This updates to sanitize and call `super.*` properly for both `setAttribute` and `updateAttribute`.

Addresses https://github.com/emberjs/ember.js/issues/14595


rwjblue:
scumbag IE10

rwjblue:
Updated to avoid IE10 issue mentioned above.

PR #346, opened 2016-11-16
Prop removal bugfix

chadhietala:
According to the semantics today we stringify `null` and covert `undefined` to `null`. In some cases we coerce `null` into an empty string. Here is [repro on 2.9.0](https://ember-twiddle.com/8293ea339131a709a3b36ca4ac09876a?openFiles=templates.application.hbs%2C) that confirms this, also illustrates this is broken in the beta if you switch the Ember versions.

I still need to link this to Ember and fix any failing tests as well. /cc @krisselden

### DIFF
--- a/packages/glimmer-runtime/lib/dom/attribute-managers.ts
+++ b/packages/glimmer-runtime/lib/dom/attribute-managers.ts
@@ -18,6 +18,10 @@ export function defaultManagers(element: Simple.Element, attr: string, isTrustin
     return defaultAttributeManagers(tagName, attr);
   }
 
+  if (attr === 'href' || attr === 'src') {
+    return defaultAttributeManagers(tagName, attr);
+  }
+
   let { type, normalized } = normalizeProperty(element, attr);
 
   if (type === 'attr') {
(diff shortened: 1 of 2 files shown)


packages/glimmer-runtime/lib/dom/attribute-managers.ts
```diff
@@ -110,11 +110,16 @@ export class PropertyManager extends AttributeManager {
...
-    if (isAttrRemovalValue(value)) {
-      this.removeAttribute(env, element, namespace);
-    } else {
-      this.setAttribute(env, element, value, namespace);
+    let attr = this.attr;
+    if (value === undefined) {
+      value = null;
+    }
+
+    if (value === null && (attr === 'value' || attr === 'type' || attr === 'src')) {
```
rwjblue: This kinda sucks 😢 

I know that this was what we were doing in HTMLBars, but should we at least tie the `attr` and `tagName` together?

packages/glimmer-runtime/tests/attributes-test.ts
```diff
@@ -232,10 +232,10 @@ test("a[href] can be removed by setting to `null`", () => {
 
   rerender({ foo: null });
 
-  equalTokens(root, '<a></a>');
+  equalTokens(root, '<a href="null"></a>');
```
rwjblue: The test expected value was correct previously (though I realize that this failure cascaded out of a fix for the other things).

If we can't fix this, I understand (I am definitely in favor of staying as close to what was shipped in 2.4 and 2.8 as possible), but it really does seem like a bug to me.

PR #347, opened 2016-11-17
Bump QUnit

chadhietala:
There is a transitive dependency using a branch of esprima in an older
version of Qunit, making hard for us to use this internally. This
version appears to not have this issue.


rwjblue:
I find it super whacky that the version numbers are so out of whack from the qunitjs.com site. But I verified that this is the latest version on npm...

PR #355, opened 2016-12-02
Fix allocing for arg compilation

chadhietala:
This is a very hot path and we need to make sure we are not allocing
context objects each time we compile the args. Still need to alloc
the array but that is far less of an issue than each context created by
`map`.

### DIFF
--- a/packages/glimmer-runtime/lib/syntax/core.ts
+++ b/packages/glimmer-runtime/lib/syntax/core.ts
@@ -1144,7 +1144,13 @@ export class NamedArgs {
 
   compile(compiler: SymbolLookup, env: Environment, symbolTable: SymbolTable): CompiledNamedArgs {
     let { keys, values } = this;
-    return new CompiledNamedArgs(keys, values.map(value => value.compile(compiler, env, symbolTable)));
+    let compiledValues = new Array(values.length);
+
+    for (let i = 0; i < compiledValues.length; i++) {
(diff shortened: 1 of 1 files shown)


rwjblue:
LGTM

chancancode:
We might not want to use `new Array(n)` here: https://twitter.com/chancancode/status/772488333117140994

chancancode:
If you can check (with Chrome's native syntaxes) whether this is still true, that would be great

krisselden:
@chancancode yes, a `new Array(n)` is allocated as fast_holey_elements but it is ok, as long as the methods we construct them by are known before it is JITd.  It is only an issue if it causes bailouts because of new shapes.

chancancode:
@krissden the problem is when the downstream code that uses the value from this function is called with both types of arrays.

There has been cases where it caused depot issues when a function has been called with only the holey version and gets optimized, but eventually got called with the non-holey version (eg EMPTY_*, or vice versa) and get deoptimized.

The problem with this has always been that we cannot guarantee a function will only ever see one of the two types of arrays.

chancancode:
Oops ^ @krisselden, sorry I am on my phone

krisselden:
That's what I meant it only is if that happens downstream of JIT, mixing types doesn't cause issue if the are equally as common the issue you are describing was because of slice converting it on a copy to non holey,  if that is downstream of this we may reconsider but that still may not be the right trade off. Growing arrays can be bad as a deopt and can even cause it with tenuring too.

PR #362, opened 2016-12-19
Better error for stray quotes

ef4:
Currently, a template like

    <div " ></div>

parses as having an attribute whose name is a double quote character. Similarly

    <div foo="bar""></div>

Parses as having two attributes. The correct `foo`, and an additional `"` attribute.

Since these are not caught at compile time, they result in confusing runtime errors when the DOM rejects our `setAttribute` with these invalid names.

This change produces a compile-time error instead, with proper line and column attribution.

### DIFF
--- a/packages/glimmer-compiler/lib/template-compiler.ts
+++ b/packages/glimmer-compiler/lib/template-compiler.ts
@@ -86,13 +86,16 @@ export default class TemplateCompiler<T extends TemplateMeta> {
   }
 
   attribute([action]) {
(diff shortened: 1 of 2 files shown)


Turbo87:
@ef4 shouldn't this result in a parse-time error or is this valid HTML?

ef4:
You are correct, it's not valid HTML, so this fix really belongs in the parser.

ef4:
Closing in favor of #373

Issue #359, opened 2016-12-17
Error with nested components


robbiepitts:
I have a [failing test](https://github.com/robbiepitts/glimmer/commit/f46a8600193474c589971d09de48311fb2ae397f#diff-65b429ee61fd1cb0aa0aa0ceddc81813R95) demonstrating an error I get when trying to render nested angle bracket components:

```
TypeError: Cannot read property 'scan' of null
    at BlockScanner.addStatement (http://localhost:7357/amd/glimmer-runtime.amd.js:6937:54)
    at OpenElement.tagContents (http://localhost:7357/amd/glimmer-runtime.amd.js:8200:25)
    at OpenElement.scan (http://localhost:7357/amd/glimmer-runtime.amd.js:8145:22)
    at BlockScanner.addStatement (http://localhost:7357/amd/glimmer-runtime.amd.js:6937:55)
    at BlockScanner.scan (http://localhost:7357/amd/glimmer-runtime.amd.js:6904:22)
    at buildStatements (http://localhost:7357/amd/glimmer-runtime.amd.js:6885:71)
    at Scanner.scanEntryPoint (http://localhost:7357/amd/glimmer-runtime.amd.js:6850:27)
    at asEntryPoint (http://localhost:7357/amd/glimmer-runtime.amd.js:9104:51)
    at Object.render (http://localhost:7357/amd/glimmer-runtime.amd.js:9117:28)
    at render (http://localhost:7357/amd/glimmer-tests.amd.js:3991:27)
```

I'm able to make the test (and whole suite) pass by checking `scanner.next()` for `null` in `OpenElement#tagContents` but I get the feeling the actual problem is elsewhere. If this is a bug I'd be happy to try to take care of it if I can get a nudge in the right direction.

For context, this is not Ember-related but rather for a little [side project](https://github.com/robbiepitts/sparkles) of mine that uses Glimmer directly.

Thanks for your time!

robbiepitts:
Fixed with #376

PR #363, opened 2016-12-20
Nested angle bracket components test

robbiepitts:
Addresses #359. Note that this is not in the context of Ember but rather [standalone Glimmer](https://github.com/robbiepitts/sparkles).

This works:

```hbs
{{#my-component}}
  {{#my-other-component}}
    Hello!
  {{/my-other-component}}
{{/my-component}}
```

This results in an error:

```xml
<my-component>
  <my-other-component>
    Hello!
  </my-other-component>
</my-component>
```

```
TypeError: Cannot read property 'scan' of null
    at BlockScanner.addStatement (http://localhost:7357/amd/glimmer-runtime.amd.js:6937:54)
    at OpenElement.tagContents (http://localhost:7357/amd/glimmer-runtime.amd.js:8200:25)
    at OpenElement.scan (http://localhost:7357/amd/glimmer-runtime.amd.js:8145:22)
    at BlockScanner.addStatement (http://localhost:7357/amd/glimmer-runtime.amd.js:6937:55)
    at BlockScanner.scan (http://localhost:7357/amd/glimmer-runtime.amd.js:6904:22)
    at buildStatements (http://localhost:7357/amd/glimmer-runtime.amd.js:6885:71)
    at Scanner.scanEntryPoint (http://localhost:7357/amd/glimmer-runtime.amd.js:6850:27)
    at asEntryPoint (http://localhost:7357/amd/glimmer-runtime.amd.js:9104:51)
    at Object.render (http://localhost:7357/amd/glimmer-runtime.amd.js:9117:28)
    at render (http://localhost:7357/amd/glimmer-tests.amd.js:3991:27)
```

My understanding of the problem:

1. `<my-component>` kicks off a scan of its children in `OpenElement#tagContents`.
2. First child is `<my-other-component>`, which also kicks off a scan of its children.
3. The `<my-other-component>` scan completes with `</my-other-component>`.
4. At this point the `<my-component>` scan's `nesting` variable (in `OpenElement#tagContents`) is at a value of 2. It started at one and was incremented when it hit `<my-other-component>`, so 2.
5. We now hit `</my-component>`, but because `nesting` is at 2 and not 1 we cannot decrement to 0 and break the loop; thus we continue…
6. The `BlockScanner` instance is out of statements, so now a call to `#next` results in `null`.
7. `scanner.addStatement(null)`
8. 💥

### DIFF
--- a/packages/@glimmer/runtime/tests/component-test.ts
+++ b/packages/@glimmer/runtime/tests/component-test.ts
@@ -61,6 +61,8 @@ class MyComponent extends BasicComponent {
   }
 }
 
+class MyOtherComponent extends BasicComponent {}
+
 QUnit.test('creating a new component', assert => {
   let template = compile("<my-component color='{{color}}'>hello!</my-component>");
   render(template, { color: 'red' });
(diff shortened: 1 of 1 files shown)


robbiepitts:
The once failing test now passes thanks to https://github.com/tildeio/glimmer/pull/376 🎉

locks:
@robbiepitts hm, I'm missing something here. If you're saying that nested angle bracket component don't work, shouldn't the build fail?

robbiepitts:
@locks Yes, when I originally opened the PR the test failed. But with #376 the test passes, so this PR just adds a passing test now. I updated the title of the PR and the commit message to reflect this.

PR #366, opened 2016-12-27
[glimmer-syntax] add a `chained` argument to program builder since we expect …

sclaxton:
…it in the actual handlebars AST. Change only required altering args list in a single invocation since the builder is currently almost always invoked with only the first argument by tests.

### DIFF
--- a/packages/glimmer-syntax/lib/builders.ts
+++ b/packages/glimmer-syntax/lib/builders.ts
@@ -191,11 +191,12 @@ function buildPair(key, value) {
   };
 }
 
-function buildProgram(body?, blockParams?, loc?) {
+function buildProgram(body?, blockParams?, chained?, loc?) {
   return {
     type: "Program",
     body: body || [],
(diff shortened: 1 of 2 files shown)


krisselden:
looks like something failed trying to launch a sauce labs browser, I restarted the build.

tomdale:
What are the semantics of `chained`?

chadhietala:
@tomdale the semantics are the same as a nested `if` in the the alternative. 

```
{{#if false}}
  Never
{{else if true}}
  Always
{{/if}}
```
Is the same as:

```
{{#if false}}
  Never
{{else}}
  {{#if true}}
     Always
  {{/if}}
{{/if}}
```

That being said we desuggar this when we parse.

tomdale:
It looks like the build is still failing. @sclaxton, would you mind describing the motivation for this change? It doesn't look like this is tested anywhere, and the lack of this change doesn't seem to cause any type errors, so curious what motivated you to make this change.

PR #367, opened 2016-12-28
Closure components tests

chadhietala:
This adds basic tests for closure components that Ember uses.


rwjblue:
@chadhietala - Any objections to merging this one?

PR #371, opened 2017-01-06
Convert Walker to class syntax

### LINKED ISSUE #369: Convert glimmer-syntax Walker to `class`
The glimmer-syntax's `Walker` class is currently using a function constructor instead of the class keyword.

Relevant code: https://github.com/tildeio/glimmer/blob/master/packages/glimmer-syntax/lib/traversal/walker.ts
Example of a class: https://github.com/tildeio/glimmer/blob/fe539562bbaf3229352214a2da645c0269c39e0f/packages/glimmer-util/lib/logger.ts#L24-L61

You can reach me or @tbieniek for help!

jbandura:
Attempts to resolve #369. Not sure about the types for `order` and `stack` in the Walker class though, any help would be greatly appreciated :)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.0.0-beta.31",
+  "lerna": "2.0.0-beta.32",
   "version": "0.21.2",
   "packages": [
     "packages/@glimmer/*"
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "ember-cli-release": "^1.0.0-beta.1",
     "ember-cli-sauce": "^1.3.0",
     "glob": "^7.0.5",
-    "lerna": "2.0.0-beta.31",
+    "lerna": "2.0.0-beta.32",
     "loader.js": "^4.0.10",
     "qunit-tap": "^1.5.1",
     "qunitjs": "^2.0.1",
--- a/packages/@glimmer/syntax/lib/traversal/walker.ts
+++ b/packages/@glimmer/syntax/lib/traversal/walker.ts
@@ -1,24 +1,27 @@
 let visitors = {
-  Program: function(walker, node, callback) {
+  Program(walker, node, callback) {
     for (let i = 0; i < node.body.length; i++) {
       walker.visit(node.body[i], callback);
     }
   },
 
-  ElementNode: function(walker, node, callback) {
+  ElementNode(walker, node, callback) {
     for (let i = 0; i < node.children.length; i++) {
       walker.visit(node.children[i], callback);
     }
   },
 
-  BlockStatement: function(walker, node, callback) {
+  BlockStatement(walker, node, callback) {
     walker.visit(node.program, callback);
     walker.visit(node.inverse, callback);
   }
 };
 
 export default class Walker {
-  constructor(order = undefined) {
+  private order: String | undefined;
+  private stack: Array<any>;
+
+  constructor(order: String | undefined) {
     this.order = order;
     this.stack = [];
   }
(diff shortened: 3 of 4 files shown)


packages/glimmer-syntax/lib/traversal/walker.ts
```diff
@@ -1,28 +1,3 @@
...
-  } else {
-    callback(node, this);
-    this.children(node, callback);
-  }
-
-  this.stack.pop();
-};
-
 let visitors = {
   Program: function(walker, node, callback) {
```
locks: You can use shorthand syntax here and change to `Program(walker, node, callback) {`

packages/glimmer-syntax/lib/traversal/walker.ts
```diff
@@ -42,9 +17,34 @@ let visitors = {
   }
 };
 
-Walker.prototype.children = function(node, callback) {
-  let visitor = visitors[node.type];
-  if (visitor) {
-    visitor(this, node, callback);
+export default class Walker {
+  constructor(order = undefined) {
```
locks: Something like this is the best I could come up with:

```
export default class Walker {
  private order: String | undefined;
  private stack: Array<any>;

  constructor(order: String | undefined) {
    this.order = order;
    this.stack = [];
  }

  …
```

jbandura:
thanks @locks! It did indeed make the tests pass :)

packages/glimmer-syntax/lib/traversal/walker.ts
```diff
@@ -1,50 +1,53 @@
...
     walker.visit(node.inverse, callback);
   }
 };
 
-Walker.prototype.children = function(node, callback) {
-  let visitor = visitors[node.type];
-  if (visitor) {
-    visitor(this, node, callback);
+export default class Walker {
+  private order: String | undefined;
```
chancancode: We don't have strict null check on master yet, so I don't think `| undefined` actually do anything (might in fact be illegal), but we will have it very soon. On that branch I think we can use `Option<string>` or something like that

packages/glimmer-syntax/lib/traversal/walker.ts
```diff
@@ -1,50 +1,53 @@
...
+  private stack: Array<any>;
+
+  constructor(order: String | undefined) {
```
chancancode: this is different than before `| undefined` is about the type, doesn't actually set any default value

packages/glimmer-syntax/lib/traversal/walker.ts
```diff
@@ -1,50 +1,53 @@
...
-    callback(node, this);
-    this.children(node, callback);
-  }
-
-  this.stack.pop();
-};
-
 let visitors = {
-  Program: function(walker, node, callback) {
+  Program(walker, node, callback) {
```
chancancode: If you can type these it would be great, but it's optional (I can see that it might be very difficult)

locks: I tried figuring it out but got kind of confused :\

locks:
It's green :sob: Sorry for the twists and turns @jbandura, thanks for your initial work!

packages/glimmer-syntax/lib/traversal/walker.ts
tomdale: Aren't these implicit `any`s? I thought we had disabled that?

Turbo87:
@chancancode @locks any plans to merge this at some point?

locks:
@Turbo87 I have the same question :C

locks:
Sorry for the messiness @jbandura, and thank you for your efforts. It seems that in the Stack VM branch the files got converted and refactored, so this PR doesn't apply anymore.
Am I correct @Turbo87?

Issue #368, opened 2017-01-06
Convert glimmer-syntax Parser to `class`


locks:
The glimmer-syntax's `Parser` class is currently using a function constructor instead of the `class` keyword.
This should be a mostly mechanical translation, being the `for` loops used for property definitions possibly the trickier bit.

Relevant code: https://github.com/tildeio/glimmer/blob/master/packages/glimmer-syntax/lib/parser.ts#L41-L101
Example of a `class`: https://github.com/tildeio/glimmer/blob/fe539562bbaf3229352214a2da645c0269c39e0f/packages/glimmer-util/lib/logger.ts#L24-L61

You can reach me or @Turbo87 for help!

robbiepitts:
I'll take a stab tonight.

Turbo87:
@rwjblue @mmun I think this can be closed now

PR #372, opened 2017-01-06
Convert `Parser` to a class

robbiepitts:
Addresses #368.

I considered defining the dynamic methods in the `constructor` method but decided against it because that is not a concern of instantiating the class. Going the prototype route also allows us to add the dynamic methods once instead of every time we instantiate.

cc @locks


robbiepitts:
I see what I did wrong. Closing for now.

PR #373, opened 2017-01-06
Report tokenization error states

ef4:
This depends on https://github.com/tildeio/simple-html-tokenizer/pull/37, which adds the new `reportSyntaxError` events.

This supersedes https://github.com/tildeio/glimmer/pull/362

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -85,9 +85,10 @@ module.exports = function(_options) {
   /*
    * ES6 Build
    */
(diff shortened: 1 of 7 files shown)


mmun:
I like this.

ef4:
(The test suite will fail until the simple-html-tokenizer PR is merged and released and we upgrade to it.)

ef4:
Looks like changes are needed in ember-cli-build.js to consume the new layout of simple-html-tokenizer that also shipped in 0.3.0 (unrelated to the change I made in it for this PR).

PR #374, opened 2017-01-07
Convert `Parser` to a class

robbiepitts:
Addresses #368.

I considered defining the dynamic methods in the constructor method but decided against it because that is not a concern of instantiating the class. Going the prototype route also allows us to add the dynamic methods once instead of every time we instantiate.

### DIFF
--- a/packages/@glimmer/syntax/lib/parser.ts
+++ b/packages/@glimmer/syntax/lib/parser.ts
@@ -38,16 +38,61 @@ export function preprocess(html, options?) {
 
 const entityParser = new EntityParser(namedCharRefs);
 
-export function Parser(source, options) {
-  this.options = options || {};
-  this.elementStack = [];
-  this.tokenizer = new EventedTokenizer(this, entityParser);
+export class Parser {
+  private elementStack = [];
+  private options: Object;
+  private source: string[];
+  public currentAttribute = null;
+  public currentNode = null;
+  public tokenizer = new EventedTokenizer(this, entityParser);
 
-  this.currentNode = null;
-  this.currentAttribute = null;
+  constructor(source, options: Object = {}) {
+    this.options = options;
 
-  if (typeof source === 'string') {
-    this.source = source.split(/(?:\r\n?|\n)/g);
+    if (typeof source === 'string') {
+      this.source = source.split(/(?:\r\n?|\n)/g);
+    }
+  }
+
+  acceptNode(node): Object {
+    return this[node.type](node);
+  }
+
+  currentElement(): Object {
+    return this.elementStack[this.elementStack.length - 1];
+  }
+
+  sourceForMustache(mustache): string {
+    let firstLine = mustache.loc.start.line - 1;
+    let lastLine = mustache.loc.end.line - 1;
+    let currentLine = firstLine - 1;
+    let firstColumn = mustache.loc.start.column + 2;
+    let lastColumn = mustache.loc.end.column - 2;
+    let string = [];
+    let line;
+
+    if (!this.source) {
+      return '{{' + mustache.path.id.original + '}}';
+    }
+
+    while (currentLine < lastLine) {
+      currentLine++;
+      line = this.source[currentLine];
+
+      if (currentLine === firstLine) {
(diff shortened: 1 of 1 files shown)


packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
 
 const entityParser = new EntityParser(namedCharRefs);
 
-export function Parser(source, options) {
-  this.options = options || {};
-  this.elementStack = [];
-  this.tokenizer = new EventedTokenizer(this, entityParser);
+export class Parser {
+  private currentAttribute = null;
+  private currentNode = null;
```
robbiepitts: Wondering if I can remove the above two properties. Can't find where they're used anywhere, internally or externally.

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -19,7 +19,7 @@ export const syntax = {
   Walker
 };
 
-export function preprocess(html, options?) {
+export function preprocess(html, options?): Object {
```
chancancode: I think we should either leave out the return type or type it properly, putting `Object` there is quit misleading imo

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
...
-  this.tokenizer = new EventedTokenizer(this, entityParser);
+export class Parser {
+  private elementStack: Object[] = [];
+  private options: Object;
+  private source: string[];
+  public currentAttribute = null;
+  public currentNode = null;
+  public tokenizer: EventedTokenizer = new EventedTokenizer(this, entityParser);
+
+  constructor(source: any, options: Object) {
```
chancancode: options = {}

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
 
 const entityParser = new EntityParser(namedCharRefs);
 
-export function Parser(source, options) {
-  this.options = options || {};
-  this.elementStack = [];
-  this.tokenizer = new EventedTokenizer(this, entityParser);
+export class Parser {
+  private elementStack: Object[] = [];
```
chancancode: ✂️ `Object` or type it properly

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
...
+  private options: Object;
+  private source: string[];
+  public currentAttribute = null;
+  public currentNode = null;
+  public tokenizer: EventedTokenizer = new EventedTokenizer(this, entityParser);
```
chancancode: can ✂️ the explicit `EventedTokenizer` type here, since we are initializing it

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
...
+
+  constructor(source: any, options: Object) {
+    this.options = options || {};
+    if (typeof source === 'string') {
```
chancancode: new line before this

packages/glimmer-syntax/lib/parser.ts
```diff
@@ -38,16 +38,60 @@ export function preprocess(html, options?) {
...
+      this.source = source.split(/(?:\r\n?|\n)/g);
+    }
```
chancancode: If source is not a string, it is uninitialized. What does that mean?

robbiepitts: @chancancode Not sure if this is is the answer you're looking for but it means that `Parser#sourceForMustache` will 💥.

robbiepitts: @chancancode Updated everything except this. Haven't been able to figure out how to answer your question yet. Any help would be appreciated.

robbiepitts:
@chancancode I think this is good to go unless you have any other comments.

PR #376, opened 2017-01-11
Move to Binary VM

chadhietala:


rwjblue:
Is there a tracking Ember update branch?

chadhietala:
@rwjblue https://github.com/emberjs/ember.js/pull/14781

chadhietala:
I'm not actually sure what is going on this branch sense it looks like there are identical commits in the log. I'm wondering if this should just get squashed down.

Turbo87:
@chadhietala can you describe what this PR did? can't make sense of the diff... 🤔

GavinJoyce:
@Turbo87 I'm interested in the description too. One thing I know is that this PR moves the opcode arrays to use typed arrays :

 * https://github.com/tildeio/glimmer/pull/376/files?diff=unified#diff-a533de1c925e49c57bbd7b2a15e1ab4dR203
 * https://github.com/tildeio/glimmer/pull/376/files?diff=unified#diff-a3dc0242300917106778e5ba4c9390bbR28

chadhietala:
@Turbo87 and @GavinJoyce I added a bit more context above.

krisselden:
@Turbo87 got rid of syntax objs, instead just directly works off the wire format to specialize and compile.  Improves the VM loop execution speed by putting the opcodes into an array, instead of a list list off megamorphic object based opcodes, opcodes are now int mapped to functions and args, and paves the way for the upcoming stack based VM.

krisselden:
This is likely to have the bounds check issues with de-structuring within a loop over a jagged array.  If the de-structuring from 2 types inlines and they have different lengths, the hoisting of their bounds check will cause problems.

packages/@glimmer/compiler/lib/javascript-compiler.ts
```diff
@@ -182,6 +236,10 @@ export default class JavaScriptCompiler<T extends TemplateMeta> {
     this.template.block.yields.add(to);
   }
 
+  debugger() {
+    this.push(['debugger', null, null]);
+  }
```
chancancode: @chadhietala what is the reason behind pushing this into the pre-compiler, vs just specializing it at runtime?

chadhietala: No specific reason. I probably should have had a better understanding of the system before implementing. Have to look at the specialization stuff.

PR #382, opened 2017-01-13
[glimmer-syntax][bugfix] codegen fixes

sclaxton:
This PR adds tests and fixes two bugs in the current codegen printer:

1. block params should be separated by spaces
2. comments should alway be generated as multi-line comments

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/print.ts
+++ b/packages/@glimmer/syntax/lib/generation/print.ts
@@ -170,7 +170,7 @@ function compactJoin(array, delimiter?) {
 function blockParams(block) {
   const params = block.program.blockParams;
   if(params.length) {
-    return ` as |${params.join(',')}|`;
+    return ` as |${params.join(' ')}|`;
   }
 }
 
--- a/packages/@glimmer/syntax/lib/generation/print.ts
+++ b/packages/@glimmer/syntax/lib/generation/print.ts
@@ -57,7 +57,7 @@ export default function build(ast) {
     }
     break;
     case 'MustacheCommentStatement': {
-      output.push(compactJoin(['{{!', ast.value, '}}']));
+      output.push(compactJoin(['{{!--', ast.value, '--}}']));
     }
     break;
     case 'ElementModifierStatement': {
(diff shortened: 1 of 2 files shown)


packages/@glimmer/syntax/lib/generation/print.ts
```diff
@@ -57,7 +57,7 @@ export default function build(ast) {
     }
     break;
     case 'MustacheCommentStatement': {
-      output.push(compactJoin(['{{!', ast.value, '}}']));
+      output.push(compactJoin(['{{!--', ast.value, '--}}']));
```
Turbo87: if that necessary? seems like it works fine without them unless you use curlys inside the comment. maybe we should adjust the parser to keep that information in the AST and only add them when they had been used before too.

sclaxton: Agree that the type of comment should be added to the AST. However, this is absolutely a bug, because if you do have curlies in a comment, which is totally feasible and super normal use case, it results in a parse error. My fix is a valid fix for the bug. We can change it to be more "accurate" with the comments once the AST has more information.

sclaxton: Besides, the opposite problem of being "inaccurate" also currently exists since all `{{!--` style comments get printed as `{{!`. So the only thing this change gains is getting rid of a parse error and it loses nothing.

rwjblue: It should be straight forward to notate the type in the AST then use it here (which both fixes the issue you were trying to fix and makes the AST more accurate).

Ping me in slack and I can explain how to do it...

packages/@glimmer/syntax/lib/generation/print.ts
```diff
@@ -170,7 +170,7 @@ function compactJoin(array, delimiter?) {
 function blockParams(block) {
   const params = block.program.blockParams;
   if(params.length) {
-    return ` as |${params.join(',')}|`;
+    return ` as |${params.join(' ')}|`;
```
mmun: Oops!

mmun:
@sclaxton Thank you!

PR #383, opened 2017-01-13
Fixes integration with simple-html-tokenizer v0.3.0

locks:


ember-cli-build.js
```diff
@@ -85,9 +85,10 @@ module.exports = function(_options) {
   /*
    * ES6 Build
    */
-  var tokenizerPath = path.join(require.resolve('simple-html-tokenizer'), '..', '..', 'lib');
+  var tokenizerPath = path.join(require.resolve('simple-html-tokenizer'), '..', '..', 'src');
   // TODO: WAT, why does { } change the output so much....
-  var HTMLTokenizer = find(tokenizerPath, { });
+  var tokenizerTree = find(tokenizerPath, { }), { output: 'tree' });
```
rwjblue: I don't think `stew.find` accepts `output` as an option (perhaps this is left over from a prior `stew.log`)?

locks: I am amazed how I shipped two syntax errors in a 3-line diff. Cleaned up!

rwjblue:
I believe you need to update the simple-html-tokenizer version in package.json also

PR #384, opened 2017-01-13
Remove for..of iterators

chadhietala:


packages/@glimmer/runtime/lib/scanner.ts
```diff
@@ -58,9 +60,10 @@ export class InlineBlock extends Template {
       builder.bindPositionalArgsForLocals(locals);
     }
 
-    for (let statement of this.statements) {
+    for (let i = 0; i < this.statements.length; i++) {
+      let statement = this.statements[i];
       let refined = SPECIALIZE.specialize(statement, table);
-      STATEMENTS.compile(refined, builder);
```
rwjblue: So this was just wrong before, right?

rwjblue: Hmm, no it seems like there is no `b` in this scope, maybe it needs to be reverted?

Turbo87:
@chadhietala why did you remove for-of?

locks:
@Turbo87 performance concerns, and the fact that it falls back to iterables

PR #385, opened 2017-01-14
Preserve original comment style when available

mmun:
Addresses @rwjblue's [comment](https://github.com/tildeio/glimmer/pull/382#discussion_r95986330) in #382.

This PR is a bit gnarly because I had to string the `source` through each function. We should rewrite these functions as a methods on a class [like these guys](https://github.com/tildeio/glimmer/blob/master/packages/@glimmer/syntax/lib/parser.ts#L62-L101).

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/print.ts
+++ b/packages/@glimmer/syntax/lib/generation/print.ts
@@ -1,7 +1,17 @@
-export default function build(ast) {
+export default function print(ast, source) {
+  if (typeof source === 'string') {
+    let lines = source.split(/(?:\r\n?|\n)/g);
+    return build(ast, { lines });
+  } else {
+    return build(ast);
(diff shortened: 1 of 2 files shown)


sclaxton:
Hmm why can't we just add a `isSafeComment` flag to the `MustacheComment` AST node so that we preserve this information in the AST?

mmun:
Because it is irrelevant to the semantics of the template, which is what the Abstract Syntax Tree is supposed to capture. For the same reason, we don't capture whitespace into the AST, whether you use `{{^}}` or `{{else}}`, etc.

krisselden:
@mmun do you need to rebase?

mmun:
Will do.

Turbo87:
closing due to inactivity and conflicts

PR #386, opened 2017-01-14
Disallow dotted helpers/modifiers

chancancode:
We don't actually use this "feature" in Ember and it is making a lot of the lookups more complicated than they need to be.

With this change, `{{foo.bar}}` is statically known to be a `get`, as opposed to an `unknown` before, saving us a `hasHelper` lookup at runtime.

Closes emberjs/ember.js#13404.


rwjblue:
Seems good to me, but how will this get updated for 0.20 branch?

chancancode:
@rwjblue at this point I think I'll just wait for the master branch to get fixed, then re-do the Work there. Doesn't seem worth it to land this for now

PR #390, opened 2017-01-17
Test nested 'else if' body location with whitespace eaten

fhrbek:
This PR introduces a test that should pass but does not with current glimmer code.

Problem:
```
{{#if foo}}
  {{bar}}
{{else if baz~}}
  <div></div>
{{/if}}
```
the snippet above reports the ```div``` tag starting on the same line as ```else if``` (as if it was appended to the line) which is obviously not correct.

Here's a snippet to investigate this structure parsed with glimmer: https://astexplorer.net/#/9zr8ioRZ4u

### DIFF
--- a/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
+++ b/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
@@ -221,15 +221,13 @@ function updateTokenizerLocation(tokenizer, content) {
   let line = content.loc.start.line;
   let column = content.loc.start.column;
 
-  if (content.rightStripped) {
-    let offsets = calculateRightStrippedOffsets(content.original, content.value);
+  let offsets = calculateRightStrippedOffsets(content.original, content.value);
 
-    line = line + offsets.lines;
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thank you for submitting the failing test case for this!

rwjblue:
@fhrbek - OK, I just pushed a commit that fixes the test. Can you submit a previously failing test (like the one here) to ember-template-lint, then `npm link` your branch and confirm it fixes both the new test and passes all the others (ember-template-lint has many more `loc` tests than glimmer itself)?

fhrbek:
@rwjblue Excellent! I'll do what you suggest and will let you know what it looks like.

fhrbek:
@rwjblue Everthing looks good! This is a PR providing (currently failing) tests: https://github.com/rwjblue/ember-template-lint/pull/164. With the fixed glimmer, however, everything works as expected.

Introducing the new glimmer requires a few more changes in the linter code that are not part of the PR. I assume those changes will be done in a separate PR when you decide to upgrade the dependency to glimmer (after new version of glimmer is released).

PR #392, opened 2017-01-19
Represent opcodes as integers in wire-format

chadhietala:
This introduces the idea of representing the wire-format opcodes as an integer instead of a string. Consider the following:

```handlebars
<div>Hello World</div>
```
__Today pre-compiles to:__

```
[
  ['open-element, 'div', []],
  ['flush-element'],
  ['text', 'Hello World'],
  ['close-element']
]
```
__After this PR__

```
[
  [11, 'div', []],
  [13],
  [0, 'Hello World'],
  [14]
]
```

In my experiments this will have a slight decrease in gzip size in a large application, however does has a decent size win post-unzipping and residing in memory. 

__Update__

The second commit introduces an enum for the most common html tags. This would produce the following:

```
[
  [11, 0, []],
  [13],
  [0, 'Hello World'],
  [14]
]
```

### DIFF
--- a/packages/@glimmer/wire-format/lib/opcodes.ts
+++ b/packages/@glimmer/wire-format/lib/opcodes.ts
@@ -1,0 +1,42 @@
+export enum Opcodes {
+  // Statements
+  Text,
+  Append,
+  UnoptimizedAppend,
+  OptimizedAppend,
+  Comment,
+  Modifier,
+  Block,
+  ScannedBlock,
+  NestedBlock,
+  Component,
+  ScannedComponent,
+  OpenElement,
+  OpenPrimitiveElement,
+  FlushElement,
+  CloseElement,
+  StaticAttr,
+  DynamicAttr,
+  AnyDynamicAttr,
+  Yield,
+  Partial,
+  StaticPartial,
+  DynamicPartial,
+
+  DynamicArg,
+  StaticArg,
+  TrustingAttr,
+  Debugger,
+
+  // Expressions
+
+  Unknown,
+  Arg,
+  Get,
+  HasBlock,
+  HasBlockParams,
+  Undefined,
+  Function,
+  Helper,
+  Concat
+}
--- a/packages/@glimmer/compiler/lib/javascript-compiler.ts
+++ b/packages/@glimmer/compiler/lib/javascript-compiler.ts
@@ -11,7 +11,8 @@ import {
   Statement,
   Statements,
   Expression,
-  Expressions
+  Expressions,
+  Ops
 } from '@glimmer/wire-format';
 
 export type str = string;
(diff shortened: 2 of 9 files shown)


packages/@glimmer/wire-format/lib/elements.ts
```diff
@@ -0,0 +1,60 @@
+export enum Tags {
```
cibernox: Would it worth to add svg tags here?
Future idea: Is it possible to statically gather the used tags analysing the templates?

packages/@glimmer/wire-format/lib/opcodes.ts
```diff
@@ -0,0 +1,42 @@
...
+  DynamicPartial,
+
+  DynamicArg,
+  StaticArg,
+  TrustingAttr,
+  Debugger,
+
+  // Expressions
+
+  Unkown,
```
dmfenton: typo here

packages/@glimmer/wire-format/lib/elements.ts
chadhietala: I can add SVG tags.Discovering is pretty difficult to do ahead of time specifically with the custom element stuff. For example we do not now AOT if `<foo-bar />` is a custom element or if it's a glimmer component.

tomdale: @chadhietala Isn't there a plan to provide the compiler with a list of known component names?

chadhietala: This is true, needs module unification.

chadhietala:
Just tried this in one of our apps. Saved 14KB on GZIP size but 800KB post-inflation.

jhartman:
Chad, what is the impact on forwards & backwards compatibility? What happens if there is a new integer the client doesn't understand?

mixonic:
@jhartman the Glimmer always requires the same version compiler and runtime on both sides of the wire format. This applies to Ember as well: You must use the Ember 2.9 template compiler with the Ember 2.9 runtime for example.

topaxi:
Attributes might be interesting too? `class`, `width`, `height`, `fill-rule` etc. should be pretty common.

chadhietala:
Running this against travis-web here are the app sizes: 

Baseline:
```
Ember@2.10.0                                  910.02 KB (185.2 KB gzipped)
```

Commit 1:

```
Ember@master+glimmer#ce58fe8   807.38 KB (182.19 KB gzipped)
```

Commit 2:
```
Ember@master+glimmer#91e4e7b   804.41 KB (182.01 KB gzipped)
```

Commit 2 doesn't really provide as much value  in its current state so I'm likely to just land the opcodes change. I have some ideas to further drop the size and deal with the repetition of user-land strings,  but is out of scope for this set of changes.

chadhietala:
@krisselden +1 in slack

PR #395, opened 2017-01-20
Reduce things noticed from the allocation profiling

krisselden:


packages/@glimmer/runtime/lib/compiled/blocks.ts
```diff
@@ -1,19 +1,10 @@
...
-
-  constructor(slice: Slice) {
-    this.start = slice[0];
-    this.end = slice[1];
-    this.slice = slice;
+  constructor(public start: number, public end: number) {
   }
 }
 
 export class CompiledProgram extends CompiledBlock {
```
chadhietala: Why do we need `CompiledProgram` at all? Where it's used is in the compile side of a block.

krisselden: it adds symbols right?

chadhietala: I think you could just use `CompiledBlock` in the places `CompiledProgram` is being used. It delegates everything to `CompiledBlock`.

PR #401, opened 2017-01-24
Fix loc info of TextNodes within ConcatStatements.

### LINKED ISSUE #399: TextNode in ConcatStatement has no loc=null
Example:

```hbs
<h1 class="{{if active "active" "inactive"}} foo">Test</h1>
```

The "foo" `TextNode` has `loc: null`.

/cc @mmun

rwjblue:
Previously, we would track only the actual chars that made up the plain text portions and would loose the specific location info for each portion.

After this change, we properly track the loc info for each part of the `ConcatStatement`.

Fixes #399.


mmun:
I reviewed the locs in the test and it looks good :+1:

rwjblue:
Thank you @mmun!

PR #403, opened 2017-01-28
syntax: Add type declarations to traverse() function

Turbo87:
/cc @mmun @rwjblue


Turbo87:
this appears to be failing due to unrelated TSLint isses that are fixed by #404

PR #406, opened 2017-02-08
Adding back -in-element and -with-dynamic-vars

chadhietala:


packages/@glimmer/test-helpers/lib/environment.ts
```diff
@@ -1086,34 +1086,6 @@ function populateBlocks(blocks: BlockMacros, inlines: InlineMacros): { blocks: B
     builder.evaluate(sexp[5]);
   });
 
-  blocks.add('-with-dynamic-vars', (sexp, builder) => {
```
rwjblue: Is this removal related to adding `in-element`?

chadhietala: Both of these were removed from the environment, but for some reason implemented just in the test environment.

chadhietala: This puts them back as first class things that an environment gets by default.

rwjblue: 👍

PR #407, opened 2017-02-16
Iterator Interface For VM

chadhietala:


packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -265,28 +271,48 @@ export default class VM implements PublicVM {
...
 
   execute(start: number, end: number, initialize?: (vm: VM) => void): RenderResult {
-    let { elementStack, frame, updatingOpcodeStack, env } = this;
+    this.prepare(start, end, initialize);
+    let result: IteratorResult<RenderResult>;
+
+    while (true) {
+      result = this.next();
+      if (result.done) break;
+    }
```
krisselden: A little more clear
```js
do {
  result = this.next();
} while (!result.done);
```

chadhietala: Resolved

rwjblue:
@krisselden - Can you review?  I'd like to land this soonish, unless you have concerns...

packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -27,6 +27,11 @@ export interface PublicVM {
   newDestroyable(d: Destroyable): void;
 }
 
+export interface IteratorResult<T> {
```
krisselden: isn't this provided by typescript?

chadhietala: I don't think so or at least I couldn't find any docs on it. https://www.typescriptlang.org/docs/handbook/iterators-and-generators.html

krisselden: @chadhietala it should be coming from node_modules/typescript/lib/lib.es2015.d.ts

packages/@glimmer/node/tests/node-test.ts
```diff
@@ -38,7 +38,13 @@ function commonSetup() {
...
   let result;
   env.begin();
-  result = template.render(new UpdatableReference(self), root, new TestDynamicScope());
+  let templateIterator = template.render(new UpdatableReference(self), root, new TestDynamicScope());
+
+  do {
+    result = templateIterator.next();
+  } while (!result.done);
+
+  result = result.value;
```
krisselden: this is because result is typed as any but this doesn't seem like a good thing todo.

packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -35,16 +40,20 @@ export default class VM implements PublicVM {
   public listBlockStack = new Stack<ListBlockOpcode>();
   public frame = new FrameStack();
   public constants: Constants;
+  private notDone = { done: false, value: null };
```
krisselden: The escape analysis in Chrome is much improved with turbofan, and this will troll it.  Please use an object literal as the iterator result should not escape the call frame of the caller.

chadhietala: Fixed

packages/@glimmer/node/tests/node-test.ts
chadhietala: Fixed

packages/@glimmer/runtime/tests/attributes-test.ts
```diff
@@ -26,7 +28,13 @@ function commonSetup() {
 function render<T>(template: Template<T>, context = {}, view: PathReference<Opaque> = null) {
   self = new UpdatableReference(context);
   env.begin();
-  result = template.render(self, root, new TestDynamicScope());
+  let templateIterator = template.render(self, root, new TestDynamicScope());
+
+  do {
+    result = templateIterator.next();
```
krisselden: don't assign the iterator to a closure, this messes up escape analysis.  the `IteratorResult<RenderResult>` should be local var, the `RenderResult` is the closure

chadhietala: This is a test and this entire file uses closures for the test helpers.

PR #411, opened 2017-02-25
Fix package typings

asakusuma:
Reference the build `index.d.ts` declarations file.

### DIFF
--- a/packages/@glimmer/compiler/package.json
+++ b/packages/@glimmer/compiler/package.json
@@ -4,7 +4,7 @@
   "main": "dist/commonjs/index.js",
   "module": "dist/modules/index.js",
   "jsnext:main": "dist/modules/index.js",
-  "typings": "index.ts",
+  "typings": "dist/modules/index.d.ts",
   "scripts": {
     "build": "../../../build/build",
     "prepublish": "npm run build",
--- a/packages/@glimmer/node/package.json
+++ b/packages/@glimmer/node/package.json
@@ -4,7 +4,7 @@
   "main": "dist/commonjs/index.js",
   "module": "dist/modules/index.js",
   "jsnext:main": "dist/modules/index.js",
-  "typings": "index.ts",
+  "typings": "dist/modules/index.d.ts",
   "scripts": {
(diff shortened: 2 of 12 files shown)


asakusuma:
Seems like the build is broken right now: https://github.com/tildeio/glimmer/pull/412

cc @rwjblue

chancancode:
@asakusuma FYI build is fixed in #416

chadhietala:
I'm not sure if this is correct or not. Currently the build situation is pretty odd. `dist` is used for consumers of the library and is built with `tsc`. In development of Glimmer we use the broccoli pipeline and I have a feeling uses the in `index.ts` for the type information. As one can imagine this causes many issues because there are multiple sources of truth in terms of the type definitions.

Before we proceed we need to check to see if this breaks the existing developer flow for people working directly on the engine. @wycats can you try this locally or if @krisselden can confirm or deny what I mentioned is true or not.

asakusuma:
Actually I think if the packages were just bumped and released, this would be fixed.

`index.ts` and `lib` were added to `package.json` `files` after the last time the packages were released: https://github.com/tildeio/glimmer/commit/1bdcdcadee30272ab207231920c358d5ca80fc32

The reason this is broken now is because `index.ts` doesn't exist in the installed packages.

krisselden:
this is correct that it should be the generated declaration files that are used.  The problem is those ones are built only during prepublish not from the ember cli build so they don't update

krisselden:
We need to unify the sources of truth here

asakusuma:
@krisselden so it sounds like the correct solution is to update the generated declaration files during the ember cli build?

wycats:
My preference would be to use `"types": "index.ts"` in development, and change it to `"types": "dist/index.d.ts"` for publication.

Using `dist` files during development makes it way too easy to accidentally change a build artifact, and in practice, you actually **do** want F12 in vscode to jump to the internal definition, not the .d.ts.

tomdale:
These are configured correctly now that Broccoli is producing the packages.

PR #413, opened 2017-02-27
Fix type issues on master

chadhietala:
Maybe there is issue non-nullable types here? Or maybe just more weird build issues.


chancancode:
That seems wrong – if TS is complaining due to the type being non-nullable, undefined would have the same issue.

chancancode:
Fixed by #416

PR #416, opened 2017-03-02
workaround default type issue

krisselden:
Specifying a default value seems to be causing issues with explicit parameter type when it is a union type.   This workaround uses the implicit type of the default by moving the type to a cast of the default value.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -300,9 +300,9 @@ module.exports = function(_options) {
   ];
 
   if (hasBower) {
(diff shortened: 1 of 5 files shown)


chadhietala:
I believe you need to change https://github.com/tildeio/glimmer/blob/master/ember-cli-build.js#L304 to `/loader.js/dist/loader`.

krisselden:
lolz

chancancode:
https://github.com/Microsoft/TypeScript/issues/14406

PR #419, opened 2017-03-03
Fix compiler package build

asakusuma:
`npm run build` in `packages/@glimmer/compiler` now passes


chancancode:
Should be covered by #421 which is now merge into master

PR #420, opened 2017-03-03
TS fixes

tomdale:
Some fixes for TypeScript configuration errors. Also fixes for some issues that are being surfaced now.


tsconfig.json
```diff
@@ -17,7 +17,10 @@
     "include": [
       "packages/*/*.d.ts",
       "packages/*/index.ts",
-      "packages/*/lib/**.ts"
+      "packages/*/lib/**.ts",
+      "packages/@glimmer/*/*.d.ts",
+      "packages/@glimmer/*/index.ts",
+      "packages/@glimmer/*/lib/**.ts"
```
krisselden: this makes this the tsconfig that likely will be used for everything unless you open a runtime file directly

krisselden: we also are still ignore tests for VS Code.

chancancode:
I merged #421 into master, which should cover everything here.

@krisselden sorry I missed your comment when looking at the other PR. Can you fix or revert the relevant things from master?

PR #424, opened 2017-03-12
Build npm packages from Broccoli

tomdale:
**UPDATE 4/19/17**:

I believe this PR is ready for merge now. The `BUILD_PACKAGE=1` environment variable below is no longer needed.

After iterating on this, we now have two distinct build pipelines:

1. Builds for tests, which is all done "locally" with our own Broccoli code.
2. Builds for distribution, which mostly lean on `@glimmer/build` to build each package in the correct { moduleFormat × jsTarget } build matrix.

Test builds should be ever-so-slightly faster than before. PhantomJS support is currently broken on master, which this PR also fixes. (Although note that currently the massive debug log spam causes testem to hang on my machine. If you eliminate that code, the tests do indeed work.)

I think I broke the Node tests but can add them back ASAP.

I've also verified Windows compatibility with @wycats and things seem to be working. I think this is good to merge and we can iterate on further problems as they arise.

---

In the epic saga of publishing each package independently to npm, a new twist arises.

Previously, we attempted to treat each package as a wholly independent entity, giving it its own `package.json`, `tsconfig.json`, build scripts, etc. At the same time, contributors (rightfully) wanted to maintain their workflow of edit-a-file-and-reload-tests-in-the-browser without increasing build time or complexity.

Unfortunately, it appears that VS Code/the TypeScript language server do not handle well the case of multiple `tsconfig.json`s when code from one directory references another. If I open a `.ts` file, it will inherit the settings from the nearest parent `tsconfig.json`. Any files _that_ `.ts` file references, via an `import` for example, will inherit the same settings, even if the referenced file "belongs to" a different `tsconfig.json`.

After (hopefully) gathering the full set of constraints and problems encountered, this is my attempt to tackle the multiple-independent-packages problem from a different angle. Rather than treating everything in `packages/` as the canonical package with its own build step, we instead treat them as "source only." The packages that ultimately get published to npm are generated synthetically as part of the build step. As a bonus, we reuse the [`@glimmer/build` package](https://github.com/glimmerjs/glimmer-build), where @krisselden and @dgeb have done a lot of work to optimize the build output and I believe reflects our most current thinking about how to package libraries.

At a high-level, this PR reads in all of the packages in `packages/` that contain a `package.json` file. Then it topographically sorts them based on their dependencies, builds them in that order, and rewrites their `package.json` to point to the output files. Because of `@glimmer/build`, we get an output something like:

```
dist/
  @glimmer/
    compiler/
      amd/
        es5/
      commonjs/
        es5/
        es2017/
      modules/
        es5/
        es2017/
      types/
        index.d.ts
        lib/
      package.json
    interfaces/
      ...
  ...
```

We also copy the package's `package.json` (where the author can specify things like name, description, dependencies, etc.) and update it to point to the correct artifact on disk:

```js
const PACKAGE_JSON_FIELDS = {
  "main": "commonjs/es5/index.js",
  "jsnext:main": "modules/es5/index.js",
  "module": "modules/es5/index.js",
  "typings": "types/index.d.ts"
};
```

This is a nice decoupling of the more semantic information about the package from the nitty gritty details of how we package and distribute the package, which we can change in the future without having to update all of the `package.json`s.

This should be working-ish but to play around with it you will need to run `ember build` with an environment variable set:

```sh
BUILD_PACKAGES=1 ember build
```

There are some open issues:

* `@glimmer/build` doesn't handle interface-only packages at the moment (i.e., `@glimmer/interfaces`).
* The build process spews a bunch of errors due to suboptimal handling of TypeScript library (`lib.*.d.ts`) files. @krisselden knows how to fix it by improving the host in `broccoli-typescript-compiler`.
* Some build time TypeScript errors are caused by not being able to find the package's dependencies. The dependencies are now topographically sorted so this should be theoretically possible, but I'd like assistance figuring out the best way to provide other `@glimmer/*` packages as dependencies to the TypeScript compiler.
* `broccoli-babel-transpiler is opting out of caching due to a plugin that does not provide a caching strategy`, something related to Babel helper generators.


bin/link-all.js
```diff
@@ -0,0 +1,24 @@
...
+"use strict";
+
+const fs = require('fs');
+const globSync = require('glob').sync;
+const execSync = require('child_process').execSync;
+const cwd = process.cwd();
+
+const LINK_COMMAND = 'yarn link';
+
+globSync('dist/**/@glimmer/*/', { cwd })
```
stefanpenner: small nit-pick, and should not be considered a blocker.

I would recommend using [walkSync](https://www.npmjs.com/package/walk-sync) here, as it prevents duplicating stating of `dir` (as during traversal we must already stat)

```js
const walkSync = require('walk-sync');
walkSync.entries('dist/**/@glimmer/*').forEach( entry => {
  if (entry.isDirectory()) {
    // ....
  }
});
``

Comment (tomdale) [inline bin/link-all.js]: @stefanpenner That is super useful, thank you.

File: bin/problems.js [hunk 2]
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
```
krisselden: why isn't this a package script that is just `tsc --noEmit`?

tsconfig.json
```diff
@@ -26,16 +26,11 @@
...
     }
   },
   "include": [
-    "packages/*/*.d.ts",
-    "packages/*/index.ts",
-    "packages/*/lib/**.ts",
-    "packages/@glimmer/*/*.d.ts",
-    "packages/@glimmer/*/index.ts",
-    "packages/@glimmer/*/lib/**.ts"
+    "packages/**/*.ts"
```
krisselden: there are no more node_modules, dist, or tmp files inside packages now?

bin/problems.js
tomdale: You raise a wise point.

tsconfig.json
tomdale: @krisselden That's correct, those packages are "source-only" and all builds happen in the root `dist/`.

build/broccoli/build-packages.js
```diff
@@ -74,7 +74,8 @@ function topsortPackages() {
 function treeForPackage(packagePath) {
   let packageTree = buildPackage({
     projectPath: packagePath,
-    tsconfigPath: TSCONFIG_PATH
+    tsconfigPath: TSCONFIG_PATH,
+    include: ['lib/**/*.ts', 'index.ts']
```
stefanpenner: are there non `*.ts` files inside lib?

tomdale: @stefanpenner Just a `package.json` file, and `@glimmer/syntax` has a `.js` file but that could be moved somewhere else (it's static and used during the build).

tomdale: @stefanpenner But this line is needed because `@glimmer/build` defaults to `['src/**/*.ts']`. I think it's passed to the tsconfig and not necessarily Broccoli, although I'd need to double check.

tomdale: @stefanpenner OK, sorry for the stream of consciousness. This path is fed to Broccoli. Previously in `@glimmer/build` it was sucking in the entire directory in the TypeScript plugin, including `tmp` and `dist`. You can guess what happened next. So now it filters only for `.ts` files in `src`. But in our case, the TS files live in the root dir, not `src`—and they don't have `dist` or `tmp` directories.

PR #425, opened 2017-03-15
[BUGFIX beta] Fix range element reporting wrong initial value

Serabe:
If a template had an `input` `range` whose type was set before a max or a min
attribute was added, the initial value would be wrong.

In plain HTML, the initial value for `<input max="10" type="range" />` would be
`5` (`(min + max) / 2 = (0 + 10) / 2 = 5`).

In Glimmer, this would be 10 because the code would be equivalent to:

```js
let input = document.createElement('input');
input.type = "range";
input.max = "10";
```

Setting `type` to `range` would make the value sanitization algorihtm to kick
in, setting the value to 50. When setting the `max` to `10`, the algorithm
would sanitize the value (50 by now) to the maximum (10).

More info in whatwg/html#2427
Fixes emberjs/ember.js#14958

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -71,8 +71,18 @@ export default class TemplateCompiler<T extends TemplateMeta> {
 
   openElement([action]) {
     this.opcode('openElement', action, action.tag, action.blockParams);
+    let typeAttr = null;
     for (let i = 0; i < action.attributes.length; i++) {
-      this.attribute([action.attributes[i]]);
+      let attr = action.attributes[i];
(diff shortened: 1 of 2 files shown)


Serabe:
Waiting for master to be stable to rebase.

chancancode:
We can't fix it at this layer because the attribute could come from a few e.g. `attributeBindings` in the Ember.Component API, so we probably need to fix it at runtime. Setting attributes go through the [`ElementOperation`](https://github.com/tildeio/glimmer/blob/master/packages/%40glimmer/runtime/lib/builder.ts#L43-L49) object which would be the right place to do it.

Serabe:
Thank you for the hint. I'll do it tomorrow (kinda late here).

Serabe:
Closed in favor of #726

PR #439, opened 2017-03-24
Mark currently failing tests as QUnit.todo.

rwjblue:
Unlike `QUnit.skip`, `QUnit.todo` still runs the test, if the test passes QUnit will mark it as a failure so that you can remember to change it from `todo` to `test`. If the test fails (or errors) it will not fail the build.

See here for details:

http://api.qunitjs.com/QUnit.todo/

screenshot

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "start": "ember serve --port=7357",
     "pretest": "ember build",
     "test": "ember test",
-    "test:ci": "ember test --test-port=7000 --config-file=testem-sauce.json && npm run test:packages && npm run test:node",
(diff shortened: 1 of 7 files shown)


rwjblue:
The new `tap_quiet_logs` flag was added in https://github.com/testem/testem/pull/1083. I pestered @johanneswuerbach RE: a release (so we don't have to pin to a SHA), but if this passes I think we should land it anyways.  

Not being able to run CI is very scary to me 😨

rwjblue:
Updated to use fork of testem with support for todos (from https://github.com/testem/testem/pull/1101).

rwjblue:
Meh.  Still not quite passing (sauce labs jobs are failing, but local `ember test` and `ember test --server` runs pass).

PR #441, opened 2017-03-25
Switch to a MIPS-based calling convention

chancancode:
In general...

- arguments are passed on the caller's stack frame
- callee saves $fp, $ra
- there are no return values

However, still lots of inconsistencies, HAX, and very hand-crafted
compilations atm.

Fixes many things, notably component hooks and bounds-tracking.

TODO: fix all the things, optimizations.


locks:
👏

packages/@glimmer/runtime/lib/scanner.ts
```diff
@@ -83,14 +84,16 @@ export const ATTRS_BLOCK = '&attrs';
...
 
   let attrsSymbol = symbolTable.symbols.length + 1;
   symbolTable.symbols.push(ATTRS_BLOCK);
 
   let statements = prelude
     .concat([[Ops.Yield, attrsSymbol, EMPTY_ARRAY]])
     .concat(head)
-    .concat(body);
+    .concat(body)
+    .concat([[Ops.ClientSideStatement, ClientSide.Ops.DidRenderLayout]]);
```
chadhietala: Can't you just do 1 `concat`?

chancancode: I tried it quickly but it makes TS sad, feel free to follow up (this thing will probably be removed soon anyway)

PR #450, opened 2017-03-30
Avoid default initializer in hot paths.

krisselden:
The arguments.length check that TypeScript emits for default initializers causes "bad arguments for context" issues in V8.


chancancode:
Is there a better compilation we can upstream to TypeScript?

krisselden:
@chancancode yes, but in these few cases I'm not sure it is worth it, except maybe `Scope` in the case of the iterate, the default path was not being used.

PR #454, opened 2017-04-05
Update repository URLs

zeppelin:
Rename `tildeio/glimmer` -> `glimmerjs/glimmer-vm`


zeppelin:
My pleasure! Also noticed most of these files has no newline at the end. Is this intentional? Asking in case if I run into this again -> fix or leave them as they are?

chancancode:
Please do fix them!

PR #463, opened 2017-04-18
WIP Remove duplicated simple-dom interfaces

josemarluedke:
This is an attempt of removing the duplicated simple-dom inteface.

This also adds `body` to SimpleDOM Document, which it is defined here: https://github.com/ember-fastboot/simple-dom/blob/master/lib/simple-dom/document.js#L12

I'm having some issues exporting `Simple` from a few places, I would love some guidance how to fix that.


packages/@glimmer/runtime/index.ts
```diff
@@ -114,7 +114,7 @@ export {
 } from './lib/modifier/interfaces';
 
 export { default as DOMChanges, DOMChanges as IDOMChanges, DOMTreeConstruction, isWhitespace, insertHTMLBefore } from './lib/dom/helper';
-import  * as Simple from './lib/dom/interfaces';
+import { Simple } from '@glimmer/interfaces';
```
josemarluedke: It seems I cannot import and export `@glimmer/interfaces` from this file. 

Error:
 
```
Could not find module `@glimmer/interfaces` imported from `@glimmer/runtime/index`
```

packages/@glimmer/runtime/lib/dom/helper.ts
```diff
@@ -277,4 +277,4 @@ helper = domChangesSvgElementFix(doc, helper, SVG_NAMESPACE);
 export default helper;
 export const DOMTreeConstruction = DOM.DOMTreeConstruction;
 export type DOMTreeConstruction = DOM.DOMTreeConstruction;
-export { Namespace as DOMNamespace } from './interfaces';
+export { Simple.Namespace as DOMNamespace };
```
josemarluedke: The tests also don't like  this. Same error as bellow: 

```Error: Could not find module `@glimmer/interfaces` imported from `@glimmer/runtime/lib/dom/helper````.

Note that this error only occurs when exporting.

krisselden:
This is no longer applicable

PR #466, opened 2017-04-20
[WIP] Unify append

chadhietala:
There is no need to create a specific object for normalizing, inserting
and updating a value. Instead we should be able to grab a singleton
manager from the VM (currently this seems like overkill) that knows how
to deal with values. We should be able to rationalize this even further
and create a "singleton manager" that knows how to insert the ambiguous
`PathExpression` that evaluates to a component definition.


chadhietala:
Finally got to run some benchmarks on this stuff.

**GC**
gc

**Phases**
phases

**Samples**
samples

Largest win appears to be a pickup in file size reduction and less allocation, which is aligned with the hypothesis.

chadhietala:
This is going to require to drag the stack base VM through the updating side to finish, as we need to store state on the stack so we can determine how we upsert.

tomdale:
@chadhietala Can we close this PR now? I think this has been superseded by your recent work.

chadhietala:
Yea we have to come back to this.

PR #474, opened 2017-04-27
Unify dom

### LINKED ISSUE #471: Dry up DOMTreeConstruction and DOMChanges
We still want to have the distinction between append and update but there is duplicate code here that can be dried up. For example things like [this](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/dom/helper.ts#L174-L197) and [this](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/dom/helper.ts#L78-L101) seems bad. While return types are different, we must find a way to unify the overlapping logic here. Typically what we do is upcast or downcast the types using the `as` operator when we call.

chadhietala:
Closes #471. This drys up majority of the duplication in the DOM construction and updating. This leaves us with 2 classes specifically for the constructing and updating but share a `DOMOperations` super class.


locks:
What's left, peer review?

chadhietala:
So I think we should land this, but we should do #492.

PR #475, opened 2017-05-02
More type fixes

chadhietala:
This cleans up some of the places where we were using implicit `any` types. There are still a great deal of type errors through the pre-compilation stack, and reference to types that don't even exist, for [example](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/syntax/lib/parser.ts#L83-L85) what is a `hbs.AST` type.


rwjblue:
LGTM

PR #478, opened 2017-05-09
Publish packages with licenses

chadhietala:
This adds a LICENSE in the root of each module we publish and also sets the license to MIT in the package.json.

### DIFF
--- a/build/broccoli/build-packages.js
+++ b/build/broccoli/build-packages.js
@@ -7,6 +7,7 @@ const Filter = require('broccoli-persistent-filter');
 const DAGMap = require('dag-map').default;
 const glob = require('glob');
 const path = require('path');
+const writeFile = require('broccoli-file-creator');
 const fs = require('fs');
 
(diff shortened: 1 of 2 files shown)


build/broccoli/build-packages.js
```diff
@@ -98,7 +98,7 @@ function treeForPackage(packagePath) {
 
   let packageJSONTree = treeForPackageJSON(packagePath);
 
-  let tree = merge([packageTree, packageJSONTree]);
+  let tree = merge([packageTree, packageJSONTree, funnel(process.cwd(), { files: ['LICENSE'], destDir: '/' })]);
```
rwjblue: We can't use process.cwd() as a tree (or we get recursive rebuilds depending on watchman).

rwjblue: We could make a temp dir and copy it in, use broccoli-file-creator, maybe use an unwatcheddir, etc? Instead

PR #481, opened 2017-05-10
Fix some of the content deopt cases

chadhietala:
This fixes some of the content deopt cases by emitting the operations to effectively do the following:

```
{{if isComponentReferent (component foo.bar) foo.bar}}
```

This also fixes an issue where the deopt instructions were not being reused, meaning we were compiling new instructions for the deopt effectively creating a slow memory leak.

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
@@ -563,7 +563,6 @@ export default class OpcodeBuilder extends BasicOpcodeBuilder {
   }
 
   guardedAppend(expression: WireFormat.Expression, trusting: boolean) {
-
     this.startLabels();
 
     this.pushFrame();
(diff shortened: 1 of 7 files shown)


packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -560,6 +562,52 @@ export default class OpcodeBuilder extends BasicOpcodeBuilder {
...
+    this.return();
+
+    this.label('END');
+
+    this.popFrame();
+
+    this.stopLabels();
+  }
+
   guardedCautiousAppend(expression: WireFormat.Expression) {
```
chancancode: Are these used?

packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
```diff
@@ -560,6 +562,52 @@ export default class OpcodeBuilder extends BasicOpcodeBuilder {
...
+
+    this.returnTo('END');
+
+    expr(expression, this);
+
+    this.dup();
+
+    this.test((reference) => {
+      return IsComponentDefinitionReference.create(reference);
+    });
```
chancancode: This could just be `this.test(IsComponentDefinitionReference.create)` right?

chadhietala: Yes

packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
chadhietala: I think Kris might have removed these in a PR after this.

PR #483, opened 2017-05-11
Refactor top level

krisselden:


packages/@glimmer/runtime/lib/scanner.ts
```diff
@@ -81,147 +81,103 @@ function compileStatements(statements: WireFormat.Statement[], meta: Compilation
...
-    let { statements: head } = scanBlock({ statements: [...attrs, ...block.head], parameters: EMPTY_ARRAY }, meta, this.env);
-    let { statements: body } = scanBlock({ statements: block.statements, parameters: EMPTY_ARRAY }, meta, this.env);
 
-    return layout(prelude, head, body, symbolTable);
+    let newStatements: WireFormat.Statement[] = [];
+
+    let toplevel: string | undefined;
+    let inTopLevel = false;
+
+    for(let i = 0; i < statements.length; i++) {
```
chadhietala: These seems like a state machine. I'm ok with leaving it but there might be a better way of expressing this. Perhaps it's something we should come back to.

packages/@glimmer/runtime/lib/syntax/client-side.ts
```diff
@@ -0,0 +1,29 @@
+import * as WireFormat from '@glimmer/wire-format';
+import { VersionedPathReference } from '@glimmer/reference';
+import { Opaque, SymbolTable } from '@glimmer/interfaces';
+import { PublicVM } from '../vm';
+
+export const enum Ops {
+  OpenComponentElement,
+  DidCreateElement,
+  DidRenderLayout,
+  FunctionExpression = 9
```
chadhietala: Why is this explicit?

krisselden: I don't remember which enum had an issue with removing an unused value.

packages/@glimmer/runtime/lib/scanner.ts
krisselden: this could should not stay here for sure, I'm hoping this can be done during compile and we could drop the largely unused scanning concept.

packages/@glimmer/runtime/lib/scanner.ts
```diff
@@ -52,6 +52,12 @@ export default class Scanner {
             decorateTopLevelElement(tagName, symbols, attrs, newStatements);
           }
           addFallback(statement, newStatements);
+        } else {
```
chadhietala: `if else`?

krisselden: I don't intend this code to live long like this, I'm just doing the minimum for the tests.  I want to abstract it more when we add better tests and try integrating.

PR #488, opened 2017-05-15
[BUGFIX release] Ensure nested custom elements fallback properly.

rwjblue:
This fixes an issue with Ember versions using glimmer-vm@0.22 when attempting to use nested web components.  This is obviously not correct for angle bracket components but since no consumers are currently using them with glimmer@0.22.x we are hard coding support to just use the fallback case.

The master version of glimmer-vm (v0.23.0-alpha.x) properly handles this without issue (for either true angle bracket invocation or web component fallback invocation).

Thanks for @krisselden for pairing with me to teach me how to do this.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.0.0-beta.31",
+  "lerna": "2.0.0-beta.32",
   "version": "0.22.0",
   "packages": [
     "packages/@glimmer/*"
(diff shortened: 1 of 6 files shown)


rwjblue:
CI is failing due to some lerna stuff I guess?

backspace:
Thank you! 💞

chadhietala:
I'm sorry you had to fight the Lerna. 
image

PR #491, opened 2017-05-19
Fix @glimmer/syntax type errors

tomdale:
Currently, downstream consumers of `@glimmer/compiler` will get an error due to a missing `.d.ts` file in `@glimmer/syntax`, which was not being generated due to type errors at compilation time.

This PR fixes two issues that were preventing the `.d.ts` files from being generated correctly:

1. The global `hbs` namespace (which at one time was provided by `@types/handlebars` but went missing at some point) has been replaced with vendored types for Handlebars AST nodes.
2. TS wasn't able to infer the types of the exported `syntax` object, so I've made them more explicit.

### DIFF
--- a/packages/@glimmer/syntax/lib/parser.ts
+++ b/packages/@glimmer/syntax/lib/parser.ts
@@ -5,6 +5,7 @@ import {
 } from "simple-html-tokenizer";
 import { Program } from "./types/nodes";
 import * as AST from "./types/nodes";
+import * as HandlebarsAST from './types/handlebars-ast';
 import { Option } from '@glimmer/interfaces';
 import { assert, expect } from '@glimmer/util';
 
@@ -80,9 +81,9 @@ export class Parser {
 
   }
 
-  acceptNode(node: hbs.AST.Program): Program;
-  acceptNode<U extends AST.Node>(node: hbs.AST.Node): U;
-  acceptNode(node: hbs.AST.Node): any {
+  acceptNode(node: HandlebarsAST.Program): Program;
+  acceptNode<U extends AST.Node>(node: HandlebarsAST.Node): U;
+  acceptNode(node: HandlebarsAST.Node): any {
     return this[node.type](node);
   }
 
@@ -90,7 +91,7 @@ export class Parser {
     return this.elementStack[this.elementStack.length - 1];
   }
 
-  sourceForNode(node: hbs.AST.Node, endNode?: { loc: hbs.AST.SourceLocation }): string {
+  sourceForNode(node: HandlebarsAST.Node, endNode?: { loc: HandlebarsAST.SourceLocation }): string {
     let firstLine = node.loc.start.line - 1;
     let currentLine = firstLine - 1;
     let firstColumn = node.loc.start.column;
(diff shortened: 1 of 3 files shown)


tomdale:
@rwjblue It exports a global, which is less than ideal. Even then, I'm not sure how reliable including it as a dependency would actually be—if it works, I think the consuming app would have to have their `tsconfig.json` configured to resolve dependencies right.

rwjblue:
> if it works, I think the consuming app would have to have their tsconfig.json configured to resolve dependencies right

Sure, but if they don't then they wouldn't care anyways (because glimmers own types would not be resolved)...

tomdale:
@rwjblue Perhaps, I've had bad experiences with re-exported dependencies before but I can't remember the details. Either way, we need such a small subset of the Handlebars types (just the AST) that, to me, it doesn't make sense to increase our dependencies and push more complexity/install time on consumers when we can include them in ~100LoC.

rwjblue:
Ya, I just worry about them getting out of sync.  At which point we could actually be introducing bugs that the type system was supposed to have been able to catch.

rwjblue:
Ultimately, I'm happy to defer to others on this.  Just trying to make sure I understand the tradeoffs we are making and whatnot.

tomdale:
@rwjblue The other thing to keep in mind is that these are community-contributed types, so they're not coming from Handlebars itself. If Handlebars makes changes to the AST, the types would get out of sync anyway. And us keeping them up-to-date is probably more likely than someone noticing and updating the types on DefinitelyTyped.

rwjblue:
Oh,  sorry, I thought we valued "community solutions", my bad ...:troll: 

In all seriousness, I'm fine with whatever WRT types.

rwjblue:
I believe that this was superceded by https://github.com/glimmerjs/glimmer-vm/pull/518 (by adding `@types/handlebars` to `dependencies`). All type errors in the project are now fixed and adding any new type errors will fail the build.

Closing...

tomdale:
@rwjblue This was closed in error. As you can see, there are not type "errors" but TypeScript cannot build a `.d.ts` file. As I said above, there are a ton of annoying issues with relying on transitive dependencies for types. This is one of them. Currently, even with #518, anyone trying to build with `@glimmer/syntax` as a dependency will get the infamous error:

```
../glimmer-vm/dist/@glimmer/syntax/dist/types/lib/parser/tokenizer-event-handlers.d.ts(1,40): error TS2307: Cannot find module './handlebars-node-visitors'.
```

This problem does not happen with this branch.

rwjblue:
Also, running `tsc` in the repo is able to generate these d.ts files, how does that work if these changes are required?

Perhaps @krisselden has an idea...

krisselden:
@rwjblue because it only generates from `.ts` we are including `.d.ts` so we need to distribute them via copying to dist.

tomdale:
Per discussion yesterday with @rwjblue and others, we are going to land this patch now to fix the broken `@glimmer/syntax` build.

However, the work @krisselden has been doing on [broccoli-typescript-compiler](https://github.com/tildeio/broccoli-typescript-compiler) should allow for re-exporting types from dependencies to work. We can revisit the decision to vendor these Handlebars AST node types once a new release of `broccoli-typescript-compiler` has been cut and we've upgraded in the `glimmer-vm` repo.

PR #497, opened 2017-05-22
Don't leak objects on the stack

chadhietala:
This ensure that we release any items that may have been retained on the stack. Prior to this commit we were leaking many objects.


packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -54,6 +57,7 @@ export class EvaluationStack {
 
   pop<T>(n = 1): T {
     let top = this.stack[this.sp] as T;
+    this.stack[this.sp] = null;
```
chancancode: :scissors:

packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -390,6 +397,9 @@ export default class VM implements PublicVM {
       if (result.done) break;
     }
 
+    // Unload the stack
+    this.stack.reset();
+
```
chancancode: This should be done in next() since that's a lower-level public API

chadhietala:
Addressed the issues.

PR #499, opened 2017-05-23
Ensure positional arguments are accounted for in builder.

rwjblue:
Prior to this commit, positional args are pushed onto the stack during `builder.component.dynamic` but we not properly `pop`'ed.

This resulted in issues when using positional param based API's (e.g. `{{render` or `{{mount` in Ember), because the stack ordering was off.

Tests exist for this in Ember itself (both `{{render` and `{{mount` tests that were failing, are passing now with this change), but adding tests here requires significant fleshing out of the testing infra.

### DIFF
--- a/packages/@glimmer/runtime/lib/syntax/functions.ts
+++ b/packages/@glimmer/runtime/lib/syntax/functions.ts
@@ -191,6 +191,13 @@ export class InvokeDynamicLayout implements DynamicInvoker<ProgramSymbolTable> {
       if (hasEval) lookup![callerNames[i]] = value;
     }
 
+    let numPositionalArgs = stack.pop<number>();
+
+    assert(typeof numPositionalArgs === 'number', '[BUG] Incorrect value of positional argument count found during invoke-dynamic-layout.');
+
+    // Currently we don't support accessing positional args in templates, so just throw them away
+    stack.pop(numPositionalArgs);
+
     let inverseSymbol = symbols.indexOf('&inverse');
     let inverse = stack.pop<Option<Block>>();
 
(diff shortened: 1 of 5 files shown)


packages/@glimmer/test-helpers/lib/environment.ts
```diff
@@ -493,7 +493,7 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
         named[name] = args.positional.at(i);
       }
 
-      return { positional: EMPTY_ARRAY, named };
+      return { positional: args.positional.capture().references, named };
```
chancancode: Revert

packages/@glimmer/test-helpers/lib/environment.ts
```diff
@@ -478,7 +478,7 @@ class EmberishCurlyComponentManager implements ComponentManager<EmberishCurlyCom
       let named = Object.assign({}, args.named.capture().map);
       named[positionalParams] = args.positional.capture();
 
-      return { positional: EMPTY_ARRAY, named };
+      return { positional: args.positional.capture().references, named };
```
chancancode: Revert

packages/@glimmer/test-helpers/lib/environment.ts
rwjblue: 👍 - updated

packages/@glimmer/runtime/lib/syntax/functions.ts
```diff
@@ -191,6 +191,13 @@ export class InvokeDynamicLayout implements DynamicInvoker<ProgramSymbolTable> {
       if (hasEval) lookup![callerNames[i]] = value;
     }
 
+    let numPositionalArgs = stack.pop<number>();
+
+    assert(typeof numPositionalArgs === 'number', 'zomg');
```
chancancode: Could probably drop this too.

chancancode: We could start sprinkling these in too, but we probably want a clearer message like [BUG] Missing argc: ...

rwjblue: Good catch, updated message to be a bit more enlightening...

PR #501, opened 2017-05-23
Ensure return value from `prepareArgs` properly handles @foo interop.

rwjblue:


packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -77,13 +77,18 @@ APPEND_OPCODES.add(Op.PrepareArgs, (vm, { op1: _state }) => {
...
+      let value = named[names[i]];
+      let atName = `@${names[i]}`;
+
+      stack.push(value);
+      atNames.push(atName);
     }
 
-    stack.push(names);
-    args.setup(stack, true);
+    stack.push(atNames);
```
chadhietala: Why do we have to bundle up these and also pass them on to the stack?

rwjblue: This is basically how the args passing convention works. Downstream things expect the stack to include this array of names (so that it can pop all the named / positional references and know what they belong to). 

Some (likely better) inline docs here: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/vm/arguments.ts#L11-L14

PR #502, opened 2017-05-23
Remove usage of `for...of`.

rwjblue:


twokul:
wait why?

rwjblue:
Was being left untranspiled in the `dist/modules/es5` output, and therefore tanking phantom runs of the ember PR.  I will come back and track down why it wasn't being transpiled, but for now its just as easy to use a normal for loop.

twokul:
👌

Issue #484, opened 2017-05-11
Errors regarding nested transactions


kbullaughey:
I've started to observe hard-to-reproduce errors of the form:

    Error: Cannot start a nested transaction

Which seem to come from an assert statement introduced in PR #444 in file:

    `packages/@glimmer/runtime/lib/environment.ts`

I've only noticed this on my production Ember app which is hard to debug given the call stack of the error just goes up backburner/runloop chain and never touches my own code.

The errors often occur a handful at a time.

Anyone have thoughts on where to dig to either reproduce or resolve such errors?

Kilowhisky:
My app is also doing this only in production mode but at the moment i'm only observing it in IE11 users. 

![image](https://cloud.githubusercontent.com/assets/7075666/26007013/9596e1ca-36f3-11e7-9f91-3e7c1ea294f6.png)

User agent string: Browser: Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; rv:11.0) like Gecko

I have my app report JS errors in production to me and one single user is generating 200+ errors per minute.

kbullaughey:
I've also only observed it in production, but also on macs: 

    Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/602.4.8 (KHTML, like Gecko) Version/10.0.3 Safari/602.4.8

It seems to always occur after an error like this:

    NotFoundError (DOM Exception 8): The object can not be found here.

That occurs when something is trying to `removeChild`.

Kilowhisky:
Actually now i'm seeing more users throwing the error in various browsers. I'm going to have to roll back until this gets sorted. 

```
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.96 Safari/537.36
```
```
Mozilla/5.0 (Windows NT 6.1; WOW64; rv:53.0) Gecko/20100101 Firefox/53.0
```

I still can't reproduce the error but the explanation does seem plausible as all my errors are occurring within a page that holds a google map that i have integrated into ember and google maps likes to more DOM nodes around.  (As far as i know i got that all sorted with the initial glimmer 2.0 release though..)

Kilowhisky:
Also here are some sample stack dumps.

```
Stack: Error: Cannot start a nested transaction
[...]

PR #505, opened 2017-05-24
[BUGFIX #484] ensure transactions are always.

stefanpenner:
- [x] need to see if this actually fixes the problem
- [x] tests
- [x] review

### DIFF
--- a/packages/@glimmer/runtime/lib/environment.ts
+++ b/packages/@glimmer/runtime/lib/environment.ts
@@ -330,7 +330,7 @@ export abstract class Environment {
   }
 
   begin() {
-    assert(!this._transaction, 'Cannot start a nested transaction');
+    assert(!this._transaction, 'a glimmer transaction was begun, but one already exists. You may have a nested transaction');
     this._transaction = new Transaction();
   }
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
Although I think ^ should be added, this wont address all cases as.

`begin` -> `commit` are actually the public API for interacting with then VM, so code from the calling library between they two may still fail. Which means that could also needs the `try/finally`

stefanpenner:
Rather then enforcing the caller to be aware of this. I change the PR to take a different approach.

* commit always resets the active transaction to null
* calling begin with an active transaction warns, then commits that transaction, and then starts the next transaction.

stefanpenner:
There are different ways we can skin this cat, would love others opinions.

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -330,7 +330,12 @@ export abstract class Environment {
   }
 
   begin() {
-    assert(!this._transaction, 'Cannot start a nested transaction');
+    let transaction = this._transaction;
+    if (transaction) {
+      this._transaction = null;
+      LOGGER.warn('a glimmer transaction was begun, but one already exists. You may have a nested transaction');
```
stefanpenner: throw in debug

krisselden:
This looks good

krisselden:
I like the fix, I'm not crazy about the test, why not throw on a component hook once like didInsertElement and test that you can rerender again?

PR #512, opened 2017-05-25
Speed up tests by 73x by making logging opt-in

stefanpenner:
## Why?

Improve development sanity for glimmer contributors by reducing test suite runtime by ~73x

With logging: 139s (\w console open)
Without Logging: 1.9s (\w console open)

## How to re-enable logging?

Obviously, logging has value (but maybe not by default), to re-enable your can:

* toggle the Glimmer Logging checkbox in QUnit
* add the glimmer_logging query param

### DIFF
--- a/packages/@glimmer/local-debug-flags/index.ts
+++ b/packages/@glimmer/local-debug-flags/index.ts
@@ -1,5 +1,10 @@
+let location = typeof window !== undefined && window.location;
+let DEBUG = false;
+if (location && /[?&]glimmer_logging/.test(window.location.search) {
+  DEBUG = true;
+}
 
-export const DEBUG = true;
+export const DEBUG = DEBUG;
 
 // TODO this is hacky but requires unifying the build
 export const CI = !!window['Testem'];
--- a/test/index.html
+++ b/test/index.html
@@ -52,6 +52,12 @@
       label: "Disable TSLint",
       tooltip: "Do not include any TSLint tests"
     });
+    QUnit.config.urlConfig.push({
+      id: "glimmer_logging",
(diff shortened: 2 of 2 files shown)


stefanpenner:
@rwjblue if you jot down your ideas somewhere (issue maybe) or here, I can try to land them.

packages/@glimmer/local-debug-flags/index.ts
```diff
@@ -1,5 +1,10 @@
+let location = typeof window !== undefined && window.location;
+let DEBUG = false;
+if (location && /[?&]glimmer_logging/.test(window.location.search) {
```
krisselden: the if is missing a closing paren

packages/@glimmer/local-debug-flags/index.ts
```diff
@@ -1,5 +1,10 @@
...
+  DEBUG = true;
+}
 
-export const DEBUG = true;
+export const DEBUG = DEBUG;
```
krisselden: this conflicts with let DEBUG

packages/@glimmer/local-debug-flags/index.ts
stefanpenner: strange, i swear this worked locally, must have botched a rebase or something

krisselden: I think the compiler was fixing it and just logging an error.

stefanpenner: strange. Anyways, thanks for fixing! Sorry about the extra work.  (hopefully the now faster tests will offset)

PR #516, opened 2017-05-25
bump ember-cli, because it has nice things

stefanpenner:


rwjblue:
We added a `yarn.lock` a little bit ago and since that point we have been using 2.12.1 ([see yarn.lock](https://github.com/glimmerjs/glimmer-vm/blob/04338d17bcdfb299391e70d3928efc2e9206331d/yarn.lock#L2211)). Can you update the `yarn.lock` too?

stefanpenner:
ah sorry, coming right up!

PR #526, opened 2017-05-29
Restore building demos

CvX:
This PR restores the demo building that was lost after #424, and also fixes some of those demos (conway’s, dbmon, uptime, and visualizer).

Ripples and benchmark are still broken. Ripples has an issue with rerendering changes made inside a RingBuffer, and benchmark gets its utf8 methods (e.g. Σ, Π) mangled in the build process.

TBH I just wanted to run visualizer to help myself learning the glimmer compilation process… 😉 

Any feedback is greatly appreciated! 😊

### DIFF
--- a/build/broccoli/build-tests.js
+++ b/build/broccoli/build-tests.js
@@ -22,8 +22,11 @@ module.exports = function(packages) {
   let testTree = merge([
     includeGlimmerAMD(amdTree),
     includeTestFiles(amdTree),
+    includeDemoFiles(amdTree),
+    includeBenchmarkFiles(amdTree),
     includeVendorDependencies(),
-    includeTestHarness()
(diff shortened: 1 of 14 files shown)


rwjblue:
Thank you so much for this!  At a recent glimmer meeting, the team decided to extract all of the demos into a real standalone glimmer app (instead of the bizarre oft-breaking setup here in this repo).  The initial effort was done a little while back (you can see it in [glimmerjs/glimmer-demos](https://github.com/glimmerjs/glimmer-demos)), but still needs help to update and confirm everything is working.

While this PR likely won't be able to land, we would **love** your help updating the glimmer-demos repo to the latest versions and to get things running. Would you be interested in helping with that?

PR #531, opened 2017-06-01
Copy custom element tests forward from 0.22.x.

rwjblue:
All of these tests pass properly, but it seems good to keep the coverage for these various scenarios.


chadhietala:
👍  I actually was going to PR these.

PR #535, opened 2017-06-01
Remove `null` from `Expressions.Value` allowed types.

rwjblue:


rwjblue:
This is the wrong approach, I believe that @krisselden is working on the correct solution and will PR a followup.

PR #536, opened 2017-06-01
Fix typing issue with compiler Syntax constraint.

krisselden:
The compiler only expects tuple syntax, it is the function expr() that separates the 2.

supersedes #535 

```ts
export function expr(expression: WireFormat.Expression, builder: OpcodeBuilder): void {
  if (Array.isArray(expression)) { // Expressions.TupleExpression
    EXPRESSIONS.compile(expression, builder);
  } else { // Expressions.Value
    builder.primitive(expression);
  }
}
```


rwjblue:
Thank you @krisselden!

Issue #348, opened 2016-11-20
Standardize Template Error Structure


jfelchner:
I'd like to begin standardizing the template error structure returned by all of the various Ember components such that we can better utilize the plethora of information that is known at the time the error is thrown.

One of the major end results of this is the ability to generate a very user-friendly error page to developers: https://github.com/ember-cli/broccoli-middleware/pull/4#issuecomment-260112229

The error structure is still up for discussion, but will be documented [here](https://github.com/ember-cli/ember-cli/issues/6441) along with the work required on the various other parts of the Ember ecosystem.

Some of the notable locations here in Glimmer are:

* https://github.com/tildeio/glimmer/blob/46656f09030bcfcb22c825f3585cb667eb5c0fe2/packages/glimmer-syntax/lib/utils.ts#L29
* https://github.com/tildeio/glimmer/blob/46656f09030bcfcb22c825f3585cb667eb5c0fe2/packages/glimmer-syntax/lib/utils.ts#L37
* https://github.com/tildeio/glimmer/blob/c16fe9e2fe16860583d74de701e5245f85f15709/packages/glimmer-syntax/lib/parser/handlebars-node-visitors.ts#L22
* https://github.com/tildeio/glimmer/search?utf8=%E2%9C%93&q=path%3Apackages%2Fglimmer-syntax+throw&type=Code

/cc @rwjblue @stefanpenner

PR #540, opened 2017-06-03
Adds `SyntaxError`

twokul:
Adds a subclass of `Error` with additional `location` property. This helps to deliver error information to `broccoli-builder` and improve error reporting as described [here](https://github.com/ember-cli/ember-cli/issues/6441).

I believe this closes https://github.com/glimmerjs/glimmer-vm/issues/348

### DIFF
--- a/packages/@glimmer/syntax/lib/errors/syntax-error.ts
+++ b/packages/@glimmer/syntax/lib/errors/syntax-error.ts
@@ -1,0 +1,23 @@
+import * as AST from '../types/nodes';
+
+/*
+ * Subclass of `Error` with additional information
+ * about location of incorrect markup.
+ */
+class SyntaxError {
+  message: string;
+  stack: string;
+  location: AST.SourceLocation;
+
+  constructor(message: string, location: AST.SourceLocation) {
+    let error = Error.call(this, message);
+
+    this.message = message;
+    this.stack = error.stack;
+    this.location = location;
+  }
+}
+
+SyntaxError.prototype = Object.create(Error.prototype);
+
+export default SyntaxError;
--- a/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
+++ b/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
@@ -3,6 +3,7 @@ import { appendChild, isLiteral, printLiteral } from "../utils";
 import * as AST from '../types/nodes';
 import { Parser, Tag, Attribute } from '../parser';
 import { Option } from '@glimmer/interfaces';
+import SyntaxError from '../errors/syntax-error';
 
 export abstract class HandlebarsNodeVisitors extends Parser {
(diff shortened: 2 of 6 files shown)


packages/@glimmer/syntax/lib/errors/syntax-error.ts
```diff
@@ -0,0 +1,17 @@
+import * as AST from '../types/nodes';
+
+/*
+ * Subclass of `Error` with additional information
+ * about location of incorrect markup.
+ */
+class SyntaxError extends Error {
```
rwjblue: Can you double check what this transpiles to? I vaguely recall issues with subclassing Error in IE9 and IE10. It may not matter as much here though since this code is primarily ran in node-land...

packages/@glimmer/syntax/lib/errors/syntax-error.ts
```diff
@@ -0,0 +1,17 @@
...
+  location: AST.SourceLocation;
```
rwjblue: Don't we generally call this `loc` everywhere else?

packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -149,7 +148,7 @@ export abstract class HandlebarsNodeVisitors extends Parser {
         break;
 
       default:
-        throw new Error(`Using a Handlebars comment when in the \`${tokenizer.state}\` state is not supported: "${comment.value}" on line ${loc.start.line}:${loc.start.column}`);
+        throw new SyntaxError(`Using a Handlebars comment when in the \`${tokenizer.state}\` state is not supported: "${comment.value}" on line ${loc.start.line}:${loc.start.column}`, rawComment.loc);
```
rwjblue: Do you think we should still include the loc info inside the message now that we thread loc through the rest of the system?

rwjblue:
Overall, looks really good. Just had a few questions/clarifications....

packages/@glimmer/syntax/lib/errors/syntax-error.ts
chadhietala: Yea this does not compile in Babel 6 correctly.

packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
twokul: I think we should; it makes the error message more useful and it would be weird to make it less useful imo

packages/@glimmer/syntax/lib/errors/syntax-error.ts
twokul: `loc` is kind of ambiguous tbh, I'd prefer `location` (also the property name that `broccoli-builder` expects is `location` not `loc`). If you feel strongly that we should stay consistent here, I can rename.

jfelchner:
This is going to be awesome! 👍

PR #541, opened 2017-06-04
Restrict type for `StaticArg` from `Expression` to `Expressions.Value`.

rwjblue:
@krisselden r?


rwjblue:
@krisselden mentioned that `Expressions.Undefined` should be allowed also, maybe making that union its own type also...

```ts
export type DirectValue = Expressions.Value | Expressions.Undefined;
```

krisselden:
is this still applicable?

locks:
It seems like all related code has since vanished. Closing…

PR #545, opened 2017-06-06
Support inverted blocks without program

### LINKED ISSUE #544: Does not support inverse blocks
```handlebars
{{^isLoggedIn}}
    <a href="signup">Sign up for free</a>
{{/isLoggedin}}
```

Errors with:
```
Uncaught TypeError: Cannot read property 'blockParams' of null
    at blockParams 
    at openBlock
    at build
```

danez:
Fixes #544

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/print.ts
+++ b/packages/@glimmer/syntax/lib/generation/print.ts
@@ -92,7 +92,7 @@ export default function build(ast: HBS.Node): string {
 
       lines.push(build(ast.program));
 
-      if(ast.inverse) {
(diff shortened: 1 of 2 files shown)


locks:
Thank you for your work @danez, but I am not sure that this syntax is something that we want to support in glimmer. The `if` and `unless` keywords were introduced to disambiguate syntax, instead of overloading the block sigil.
I shall await confirmation before closing the PR, in case I'm wrong.

wycats:
@danez sorry you closed this before we got back to you. I can confirm that `#unless` was intended to supplant `{{^}}`.

PR #547, opened 2017-06-12
Improve support for XHTML

cbou:
`tagName` are always lower case for xhtml documents.

This is the original problem of https://github.com/emberjs/ember.js/issues/15333


cbou:
would it be possible to merge this PR?

PR #548, opened 2017-06-13
Types for AST transform plugins

tomdale:
You can pass AST transform plugins to the `preprocess`/`precompile` options but previously the options hash was untyped. This PR adds types for the options hash as well as types for AST plugins.


chriseppstein:
Merge & Release? 🙏

PR #550, opened 2017-06-16
Fix handlebars `parse` error

robbiepitts:


rwjblue:
There is an odd MS Edge failure that seems unrelated to this change (likely transient issue with Sauce Labs).

PR #551, opened 2017-06-16
[BREAKING] Use plain functions for AST plugins.

rwjblue:
This changes AST Plugins to be "plain functions" with the following interface:

```ts
import { AST, Syntax } from '@glimmer/syntax';

export type NodeVisitor = {
  [P in keyof AST.Nodes]?: NodeHandler<AST.Nodes[P]>;
};

export interface ASTPluginResult {
  name: string;
  visitors: NodeVisitor;
}

export interface ASTPlugin {
  (env: ASTPluginEnvironment): ASTPluginResult;
}

export interface ASTPluginEnvironment {
  meta?: any;
  syntax: Syntax;
}
```

The current system of instantiation is fairly confusing and makes a number of use cases pretty difficult (where you need to preserve some other outside state that you access from within your plugin's `transform`).

The old API can easily be reimplemented in terms of the plain functions in consumers (e.g. Ember for compatibility):

```js
let uuid = 0;

  function ensurePlugin(FunctionOrPlugin: any): ASTPlugin {
    if (FunctionOrPlugin.prototype && FunctionOrPlugin.prototype.transform) {
      return (env) => {
        return {
          name: 'legacy-transform' + (++uuid) ,
          visitors: {
            Program(node: AST.Program) {
              let plugin = new FunctionOrPlugin(env);

              plugin.syntax = env.syntax;

              return plugin.transform(node);
            }
          }
        };
      };
    } else {
      return FunctionOrPlugin;
    }
  }
```

A test exists for this (to ensure Ember can provide backwards compat).

### DIFF
--- a/packages/@glimmer/compiler/lib/compiler.ts
+++ b/packages/@glimmer/compiler/lib/compiler.ts
@@ -2,17 +2,14 @@ import { preprocess } from "@glimmer/syntax";
 import TemplateCompiler, { CompileOptions } from "./template-compiler";
 import { SerializedTemplateWithLazyBlock, TemplateJavascript, TemplateMeta } from "@glimmer/wire-format";
 import { Option } from "@glimmer/interfaces";
-import { TransformASTPluginFactory } from "@glimmer/syntax";
+import { PreprocessOptions } from "@glimmer/syntax";
 
 export interface TemplateIdFn {
   (src: string): Option<string>;
 }
 
-export interface PrecompileOptions<T extends TemplateMeta> extends CompileOptions<T> {
+export interface PrecompileOptions<T extends TemplateMeta> extends CompileOptions<T>, PreprocessOptions {
   id?: TemplateIdFn;
-  plugins?: {
-    ast?: TransformASTPluginFactory[]
-  };
 }
 
 declare function require(id: string): any;
--- a/packages/@glimmer/syntax/index.ts
+++ b/packages/@glimmer/syntax/index.ts
@@ -1,8 +1,9 @@
 // used by ember-compiler
 export {
   preprocess,
-  TransformASTPlugin,
-  TransformASTPluginFactory,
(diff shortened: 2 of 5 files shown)


mmun:
I've got a couple thoughts.

1) Massage the arguments a bit so we don't have a dangling `syntax` argument.
2) Provide a declarative way to define the visitors. It doesn't cost us much (e.g. you won't be able to use the effectively deprecated `Walker` class) and it gives us a chance to try plugin merging compilation optimizations in the future):

```typescript
export interface ASTPlugin {
  (env: ASTPluginEnvironment): ASTVisitorHandlers;
}

export interface ASTPluginEnvironment {
  meta: {},
  file: FileData,
  syntax: Syntax
}
```

rwjblue:
Updated with feedback after discussing with @mmun.

rwjblue:
Still need to update for:

> Provide a declarative way to define the visitors.

mmun:
It's probably important to agree on a type for `ASTVisitorHandlers` as well.

```typescript
interface ASTVisitorHandlers {
  ComponentStatement: ASTVisitorHandler<ComponentStatement>,
  ...
}

interface ASTVisitorHandler<NodeType> {
  (path: ASTVisitorPath) => NodeType
}
```

ASTVisitorPath is motivated in https://github.com/rwjblue/ember-template-lint/issues/186. It's not clear that we can nail down a path API in a reasonable time frame though.

rwjblue:
OK, just updated for the newer API excluding the migration to using `ASTVisitorPath`, I'll work on that next...

rwjblue:
After discussing with @mmun, I think we have decided to punt on that last point (e.g. using some sort of `Path` object) for now. We are still discussing possible future options here, but those changes likely need more "baking" and will be done in a separate PR...

@mmun - I _think_ this is ready for another review.

PR #555, opened 2017-06-17
Fix static multi select test for Edge 15

### LINKED ISSUE #553: Report MS Edge 15 Issue
Need to research what is going on, but don't have access to MS Edge...

t-sauer:
Fixes #553 for now


rwjblue:
Thank you @t-sauer!!

PR #557, opened 2017-06-22
[BREAKING] Update ASTPluginResult to use `visitor` (no plural form).

rwjblue:
The recent refactor mistakenly used the plural form (`visitors`), and should have used `visitor`.

This fixes the mistake mentioned in https://github.com/glimmerjs/glimmer-vm/commit/28652d4ea214f1c7bf0251d4b4b23853d9f0a787#commitcomment-22696220.

/cc @mmun @chriseppstein

### DIFF
--- a/packages/@glimmer/syntax/index.ts
+++ b/packages/@glimmer/syntax/index.ts
@@ -3,6 +3,7 @@ export {
   preprocess,
   PreprocessOptions,
   ASTPlugin,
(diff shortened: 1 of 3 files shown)


rwjblue:
As background, I did investigate making this a non-breaking change (it would have been fairly easy), but other unreleased changes on master already include breaking changes so I decided to also update the types and avoid the fallback.

PR #558, opened 2017-06-22
[BREAKING] Make the manager decide what tag to use

chancancode:
This is a follow-up from 1908dbe.

Currently we automatically combine the tag from the args object with an optional dirtiness tag returned by Manager#getTag. After this change the component/modifier is responsible for fully deciding what tag to use.

To preserve the existing behavior, just continue to combine the args tag (which you could save off in `create`) with whatever you were returning before. However, this also allows the component to opt-out of updating by returning a `CONSTANT_TAG` if it doesn't use the args.

With this, components will get const-optimization correctly.


chancancode:
@rwjblue @dgeb FYI

PR #560, opened 2017-06-28
More targeted test harnesses

chadhietala:
In order to fan out the test migration we need a bit more cleaner abstractions that people can just take off the shelf and start migrating. This introduces a `RenderTests` and `RehydrationTests` which encapsulate the methods for testing those flows. Also started using Prettier, not sure how people feel about its opinions. 😬


packages/@glimmer/runtime/test/initial-render-test.ts
```diff
@@ -1055,59 +1051,15 @@ abstract class RenderTest extends BaseRenderTest {
...
-class Rehydration extends RenderTest {
-  protected element: HTMLDivElement;
-  protected template: Option<Template<Opaque>>;
-  private isBrowser = false;
-
-  constructor(env = new TestEnvironment({ document: new SimpleDOM.Document() })) {
-    super(env);
-    this.element = env.getDOM().createElement('div') as HTMLDivElement;
-  }
+class Rehydration extends RenderTests implements RehydrationTests {
```
chadhietala: This needs to be `RenderingTests`.

Issue #561, opened 2017-06-28
[Pls send halp again!] Test Porting Guide 


chadhietala:
**Requires** https://github.com/glimmerjs/glimmer-vm/pull/560

## Backstory
The repository originally started off as being a fork of HTMLBars and thus we inherited the test suite. While many tests have been added the test suite has grown somewhat organically over the past year and half. We have seen several test harnesses be introduced making it hard to understand exactly how to introduce new tests 😞 . Luckily we have landed on a test harness that we really like and need halp migrating the rest of the test suite to use it 🤞 😄 🤞 .

## Turning On The Bat Signal 🦇 

People from the internet, we need your halp! This effort is similar to the effort when we originally [introduced the Glimmer VM into Ember](https://github.com/emberjs/ember.js/issues/13127). At this point, the highest value you can do to help is to port the existing tests (and help reviewing these PRs). If you are familiar this the new test harness in Ember you should feel right at home 🏠 . If not, no reason to be worried, we have outlined the migration steps below.

## The Process
In general, and at the high-level, the process looks like this:

1. Pick a test file to port (will be found in `packages/@glimmer/runtime/test`)
2. Determine if you can break apart the tests into a more meaningful file
3. Import `RenderTests` and `module` from `@glimmer/test-helpers`
4. Extend the `RenderTest` creating your own subclass
   - Port each tests in the file and module
   - Make sure you follow the I-N-R-U cycle where appropriate
   - Remove duplicated tests 😬 
5. Pass the new subclass into `module` as such `module('NAME OF MODULE', MyTestClass)`
6. Remove the old test file
7. Create a PR
   - If you have collapsed or deleted test cases please leave comments in the Github issue explaining what was collapsed. This typically is done when the I-N-R-U cycle was spread across multiple test cases.

## How to write good tests

Here are a few general tips/rules that you can follow to improve the test cases.

### The "I-N-U-R" cycle

We would like each test to go through the "I-N-U-R" cycle:

1. Initial render
   
   Render the template you want to test, with the template and initial values of your choice (`this.render('<h1>Hello {{name}}</h1>', { name: 'Chad' })`) and assert the result is what you expect. ([Example](https://github.com/glimmerjs/glimmer-vm/blob/af9294247738503612b707546475a06ba4fdfa0e/packages/%40glimmer/runtime/test/initial-render-test.ts#L650-L651))
2. No-op re-render
   
[...]

PR #564, opened 2017-07-03
New test harness for attributes test

Serabe:


Serabe:
@chadhietala I would appreciate a first input on this. It is mostly the harness and the `renderTemplate` implementation was done mostly by imitation, so any input would be more than welcomed.

Serabe:
#561

packages/@glimmer/runtime/test/attributes-test.ts
```diff
@@ -68,204 +62,214 @@ function nativeValueForElementProperty<T extends keyof HTMLElementTagNameMap, P
...
 
-  equalTokens(root, '<input />');
-
-  rerender({ enabled: true });
-
-  equalTokens(root, '<input disabled />');
+  constructor(env = new TestEnvironment()) {
+    super(env);
+    this.element = env.getDOM().createElement('div') as HTMLDivElement;
+    //this.element.setAttribute('debug-root', 'true');
```
chadhietala: Remove ✂️

packages/@glimmer/runtime/test/attributes-test.ts
```diff
@@ -68,204 +62,214 @@ function nativeValueForElementProperty<T extends keyof HTMLElementTagNameMap, P
...
-  rerender({ enabled: false });
-
-  equalTokens(root, '<input disabled />');
-
-  rerender({ enabled: 'wat' });
-
-  equalTokens(root, '<input disabled />');
-
-  rerender({ enabled: null });
+abstract class RenderTest extends BaseRenderTest {
```
chadhietala: I merged #560 so I'm not sure if you need this any more.

packages/@glimmer/runtime/test/attributes-test.ts
```diff
@@ -68,204 +62,214 @@ function nativeValueForElementProperty<T extends keyof HTMLElementTagNameMap, P
...
-
-  rerender({ foo: 'http://foo.bar' });
+    this.rerender({ foo: 'javascript:foo()' });
+    this.assertHTML('<a href="unsafe:javascript:foo()"></a>');
+    this.assertStableNodes();
+  }
 
-  equalTokens(root, '<a href="http://foo.bar"></a>');
-});
+  @test "a[href] marks javascript: protocol as unsafe on updates"() {
```
chadhietala: This seems to be covered at [L190](https://github.com/glimmerjs/glimmer-vm/pull/564/files?diff=split#diff-37847c52bf54b71f8378543edd6d9407R190)

Serabe:
@chadhietala this is ready for review. The tests related to sauce labs are failing (the four of them). Shall I just restart job like in Ember or is there anything I am doing wrong here?

OTOH, is there any way to remove the TS2300 and related errors?

Thank you!

chadhietala:
@Serabe I think there might be legit errors here with IE. I think you might need to write some type of condition for IE like I did [here](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/test/initial-render-test.ts#L7-L21) looks like `textarea` doesn't behave the same. What is the `TS2300` error?

chadhietala:
@Serabe this looks great otherwise.

Serabe:
TS2300 and related errors are warnings from TypeScript when it finds a duplicated definition. It seems to be coming from the type declaration of es2015 and such. It raises so many that Travis log is hard to use.

I've added the select quirk and I'm looking for the one for textarea.

Thank you!

Serabe:
0003screenshot

Seems that nodes are not stable in `"handles empty string textarea values"` after setting the value to `Alex`

Serabe:
2017-07-11 at 16 59

The errors I mentioned were these. There are over 8.5K lines of this bloating travis log :$

Serabe:
I've been a few hours digging into this in IE11 and cannot find anything :(

Any help available figuring this out?

chadhietala:
@krisselden I check on a VM of IE and it appeared IE and Edge set do different things with CDATA in a textarea c/d?

Serabe:
@chadhietala anything I can do to help fix this?

Serabe:
@chadhietala @krisselden ping?

chadhietala:
@serabe we removed IE9/IE10/Pantom from master. Can you try to rebase and see if the tests just pass?

Serabe:
Done. Still failing IE11 and Edge.

locks:
@Serabe You've Got Conflicts™.

Serabe:
@locks already in Berlin, not sure when I'll be able to rebase again, but the underlying issue is still there.

chadhietala:
Ok I have an answer. Firstly `value` is [not a valid html attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) for `textarea`. Yet this still works if you set the prop, which is what we do today. That being said it appears that browsers differ how they represent the text. In Chrome you don't see a text node with the value instead you just see the following:

screen shot 2017-10-09 at 5 05 04 pm

Notice it does not have a `firstChild`. Now with IE and Edge if we start off with an empty string or `null` we have a similar behavior.

screen shot 2017-10-09 at 5 18 52 pm


However when we do update the `textarea` we have the following:

screen shot 2017-10-09 at 5 05 42 pm

So IE and Edge inject a TextNode when `value` is set.

**Bold What Shall We Do???**
Well because the result is the same between these browsers but the implementation details are slightly different we should just not `assertStableNodes` as they are not going to be stable. Instead we should just keep the assertion that the value updated. I can rebased and fix this if you don't have the time.

chadhietala:
Merged as part of #698. Sorry this took so long @Serabe.

PR #570, opened 2017-07-09
check for value for string type before creating `StringReference`

bekzod:


tomdale:
@bekzod Apologies, I don't think I understand the motivations behind this change. Is this for performance or correctness?

chancancode:
According to TypeScript, it seems like by the time you get to the `else` branch it is impossible for it to be anything other than a string. If we are worried about JavaScript code calling it incorrectly, we can add an assertion instead?

bekzod:
I guess was paranoid to much, closing it :P thanks

PR #574, opened 2017-07-09
refactor constant flags

bekzod:


bekzod:
bitwise refactor also correctly restores negative values:
before: ((2 << 30) | -10) & ~(3 << 30)  === 1073741814 
now: (-10 << 3 | 2) >> 3  === -10
(where 2 is type flag -10 primitive value);

tomdale:
Thank you @bekzod!

PR #575, opened 2017-07-10
Custom element modifiers

robbiepitts:
Here is a failing test for modifiers in custom elements. Is this something for which support is desired (I mean, I want it but not sure if anyone else cares about it)?

Primary use case for me would be adding properties on custom elements:

```ts
class SomeGlimmerComponent {
  foo = "whatever";
}
```

```hbs
<some-glimmer-component>
  <some-custom-element {{prop "foo" foo}}></some-custom-element>
</some-glimmer-component>
```

```js
document.querySelector('some-custom-element').foo;
// => "whatever"
```


tomdale:
This test should be passing now that we no longer have ambiguity between custom elements and Glimmer components. I resolved the merge conflicts, just waiting to see if the tests pass.

rwjblue:
Awesome, thanks @tomdale and @robbiepitts!

PR #583, opened 2017-07-15
Port in element tests

piotrpalek:
**DONE**

Related issue: #561

### DIFF
--- a/packages/@glimmer/runtime/test/in-element-test.ts
+++ b/packages/@glimmer/runtime/test/in-element-test.ts
@@ -127,46 +127,41 @@ class InElementTests extends RenderTests {
     this.assertHTML("<!---->");
     this.assertStableRerender();
   }
+
+  @test "With nextSibling"() {
+    let externalElement = document.createElement("div");
+    externalElement.innerHTML = "<b>Hello</b><em>there!</em>";
+
+    this.render(
+      stripTight`{{#-in-element externalElement nextSibling=nextSibling}}[{{foo}}]{{/-in-element}}`,
+      { externalElement, nextSibling: externalElement.lastChild, foo: "Yippie!" }
+    );
+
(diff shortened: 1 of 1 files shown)


piotrpalek:
@chadhietala would appreciate a quick review to see if I'm going in the right direction with this

chadhietala:
Looking good. For quotes what you've done is fine.

packages/@glimmer/runtime/test/in-element-test.ts
```diff
@@ -1,487 +1,441 @@
...
-    }
-  );
-
-  equalsElement(externalElement, 'div', {}, stripTight``);
-  assert.equal(destroyed, 0, 'component was destroyed');
-
-  set(view, 'showExternal', true);
-  rerender();
-
-  assertElementIsEmberishElement(externalElement.firstElementChild!, 'div', { }, 'destroy me!');
```
piotrpalek: @chadhietala not sure what to do about this, it doesn't seem to render the same way?

chadhietala: Not sure if I follow? Isn't this tested on L428 of the new tests?

piotrpalek: @chadhietala sorry that probably wasn't very clear. What I meant is that I couldn't use the `assertElementIsEmberishElement` method in the new tests and instead just did  `equalsElement`. My question was if this is an issue or if it's fine.

piotrpalek:
@chadhietala other than the one question I had it should be good to go if it looks good to you

packages/@glimmer/runtime/test/in-element-test.ts
chadhietala: This seems fine.

PR #584, opened 2017-07-15
Migrate partial-test following issue #561 guidelines

lbdm44:
Migration of partial-test.ts according to #561.

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/builder.ts
@@ -12,7 +12,7 @@ import {
   LazyConstants
 } from '../../environment/constants';
 import { ComponentBuilder as IComponentBuilder } from '../../opcode-builder';
-import { Primitive } from '../../references';
+import { Primitive, PrimitiveType } from '../../references';
 import { expr, ATTRS_BLOCK } from '../../syntax/functions';
 import { BlockSyntax, CompilableTemplate } from '../../syntax/interfaces';
 import RawInlineBlock from '../../syntax/raw-block';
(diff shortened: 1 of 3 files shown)


packages/@glimmer/runtime/test/partial-test.ts
```diff
@@ -1,470 +1,387 @@
...
-
-  equalTokens(root, `You smaht. You loyal. `);
-  rerender({ qualities: ['smaht', 'loyal'] }, { assertStable: true });
-  equalTokens(root, `You smaht. You loyal. `);
-});
-
-QUnit.test('static partial with named arguments', () => {
-  env.registerBasicComponent('foo-bar', BasicComponent, `<p>{{@foo}}-{{partial 'test'}}</p>`);
+    this.assertStableRerender();
+    this.assertHTML('Respeck my name. When my name come up put some respeck on it.');
```
chadhietala: Can you add an update and reset step for this.

```
this.rerender({ item: 'age' });
this.assertHTML('Respeck my age. When my age come up put some respeck on it.');
this.assertStableNodes();

this.rerender({ item: 'name' });
this.assertStableNodes();
this.assertHTML('Respeck my name. When my name come up put some respeck on it.');
```

packages/@glimmer/runtime/test/partial-test.ts
```diff
@@ -1,470 +1,387 @@
...
 
-  env.registerPartial('test', `{{@foo}}-{{@bar}}`);
-  render(template, { foo: 'foo', bar: 'bar' });
-  equalTokens(root, `<p>foo-foo-bar</p>`);
+    this.render(`{{#each qualities key='id' as |quality|}}{{partial 'test'}}. {{/each}}`, props);
+    this.assertHTML('You smaht. You loyal. ');
+    this.assertStableRerender();
 
-  rerender(null, { assertStable: true });
+    this.rerender(props);
```
chadhietala: This is basically doing the same thing as L39. Can you add an update and reset step like I mentioned above.

chadhietala:
@lbdm44 As part of porting the tests we would like to get more coverage in the cases. In all of the units that are dynamic can you add the update and reset steps like I mentioned above? Also when we are re-rendering we want to use the `assertStableNodes` method instead of the `assertStableRerender`. Here is a good example https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/test/initial-render-test.ts#L237-L253

PR #585, opened 2017-07-16
[CLEANUP] Remove the use of angle type assertion

Serabe:
Most new type assertions added to code are `as` type assertions. Though
we don't care too much about `tsx`, the only linting rule available for
consistent type assertions is `no-angle-bracket-type-assertion`.

Also, some syntax highlighters doesn't work well with angle bracket type
assertions.


tomdale:
@Serabe Sorry for not getting to this sooner. Is there an automated tool we can run on the codebase so we don't have to deal with the merge conflicts that have happened since July?

t-sauer:
@tomdale It can be fixed automatically with TSLint: https://palantir.github.io/tslint/rules/no-angle-bracket-type-assertion/

Serabe:
Updated. It is much simpler to use what @t-sauer said and push force to the same remote branch.

tomdale:
Thank you!

PR #588, opened 2017-07-17
Introduce tests for the test harness

chadhietala:
This places the new test harness under test and fixes some issues when writing the tests for creating component invocations.

### DIFF
--- a/packages/@glimmer/test-helpers/lib/abstract-test-case.ts
+++ b/packages/@glimmer/test-helpers/lib/abstract-test-case.ts
@@ -163,12 +163,38 @@ export abstract class AbstractRenderTest {
     return invocation;
   }
 
-  private buildArgs(args: Object, sigil = ""): string {
-    return `${Object.keys(args).map(arg => `${sigil}${arg}=${args[arg]}`).join(" ")}`;
+  private buildArgs(args: Object): string {
+    let { testType } = this;
+    let sigil = "";
+    let needsCurlies = false;
+
+    if (testType === "Glimmer" || testType === "Basic") {
+      sigil = "@";
+      needsCurlies = true;
+    }
+
+    return `${Object.keys(args)
+      .map(arg => {
+        let rightSide: string;
+
+        if (needsCurlies) {
+          let isString = arg[0] === "'" || arg[0] === '"';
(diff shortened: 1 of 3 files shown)


packages/@glimmer/test-helpers/lib/abstract-test-case.ts
```diff
@@ -163,12 +163,38 @@ export abstract class AbstractRenderTest {
...
+    let { testType } = this;
+    let sigil = "";
+    let needsCurlies = false;
+
+    if (testType === "Glimmer" || testType === "Basic") {
+      sigil = "@";
+      needsCurlies = true;
+    }
+
+    return `${Object.keys(args)
```
scalvert: Does this need to be a template literal?

packages/@glimmer/test-helpers/lib/abstract-test-case.ts
```diff
@@ -163,12 +163,38 @@ export abstract class AbstractRenderTest {
...
+        }
+
+        return `${sigil}${arg}=${rightSide}`;
+      })
+      .join(" ")}`;
   }
 
   private buildBlockParams(blockParams: string[]): string {
-    return `${blockParams.length > 0 ? `as |${blockParams.join(" ")}|` : ""}`;
+    return `${blockParams.length > 0 ? ` as |${blockParams.join(" ")}|` : ""}`;
```
scalvert: Do you think it's worth defining `const SPACE = " ";`. You use it quite a few times in this file (though I realize this is just a test infra file)?

packages/@glimmer/test-helpers/lib/abstract-test-case.ts
```diff
@@ -688,7 +740,7 @@ function componentModule(name: string, klass: typeof AbstractRenderTest & Functi
           case "glimmer":
             tests.curly.push(createTest(prop, test));
             tests.dynamic.push(createTest(prop, test));
-            tests.glimmer.push(createTest(prop, test));
+            tests.glimmer.push(createTest(prop, test, true));
```
scalvert: I noticed this when I was trying to skip a glimmer test. Thanks for fixing.

PR #591, opened 2017-07-20
Improve support for XHTML

cbou:
XML Documents preserves `tagName` cases. EmberJS usually generate tags with lower case tag name. Calling `node.tagName` returns a lower case name like 'input' and glimmer-vm test it against 'INPUT'.

This is the original problem of emberjs/ember.js#15333

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/attributes/dynamic.ts
+++ b/packages/@glimmer/runtime/lib/vm/attributes/dynamic.ts
@@ -173,11 +173,11 @@ export class OptionSelectedDynamicAttribute extends DefaultDynamicProperty {
 }
 
 function isOptionSelected(tagName: string, attribute: string) {
-  return tagName === 'OPTION' && attribute === 'selected';
+  return tagName.toUpperCase() === 'OPTION' && attribute === 'selected';
 }
 
 function isUserInputValue(tagName: string, attribute: string) {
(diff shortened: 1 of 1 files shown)


tomdale:
The only concern I have with this is performance-related. It seems like it will impose some cost on every single attribute on the page for a use case that I think is uncommon. The questions I'd like to answer before merging are:

1. What is the actual performance impact? If the difference is not detectable in benchmarks then it seems fine to normalize tag names like this.
2. Generally speaking, are we equipped to support XHTML compatibility? If we take this patch, will other issues crop up down the road with XHTML compatibility that would be more invasive?

I'm leery of saying we support XHTML as a first-class thing, but am happy to take patches that improve compatibility if the cost is minor.

cbou:
1. I don't measure some performances. Is there anything I can easily run to check this?
2. For the moment it's the only issue I noticed.

cbou:
Is there any chances this will be merged?

tomdale:
@cbou I'm OK with this but I want to be upfront that XHTML compatibility is not a high priority for us.

To merge this, we would need a couple changes:

1. The implementation should normalize `tagName` once. This currently runs `toUpperCase()` on the same string 3 times.
2. Include a test for the normalization case.

Thanks!

cbou:
I create a new PR for this. https://github.com/glimmerjs/glimmer-vm/pull/764

PR #593, opened 2017-07-20
Revive Node tests

tomdale:
When I did the build refactor to use Broccoli to build independent packages, I broke support for testing Node/CommonJS builds. This commit reintroduces those tests and fixes some small regressions that had occurred since then.

It works by using the TypeScript Broccoli plugin to compile all of the Glimmer packages to CommonJS. It saves these into `dist/tests/node_modules/@glimmer/*`. I had to put them in a `node_modules` directory so Node module resolution works; for example, if `@glimmer/runtime` calls `require('@glimmer/util')`, it needs to know how to find that package.

It might be better to do Lerna-style symlinking based on each package's actual dependencies. Right now, if a package forgets to add a dependency to its `package.json`, the tests will still pass even though it will break if users try to install and run it.

I also had to make some changes to get the tests passing again:

1. Putting attribute curlies in a string no longer forces string coercion (i.e., if `foo` is null, `disabled="{{foo}}"` should not emit the attribute; previously it was emitting `disabled="null"`).
2. `typeof window` might return `"undefined"`, but never `undefined` :trollface: 
3. The hash-based ID of compiled templates changed. I assume this is a byproduct of either the hashing function or the input to the function changing and updating the test is harmless, but would love someone to check me on that.


build/broccoli/build-tests.js
```diff
@@ -24,14 +24,43 @@ module.exports = function(packages) {
...
   ]);
 
   return funnel(testTree, {
     destDir: 'tests'
   });
 }
 
+function includeNodeTests() {
+  let tsTree = funnel('packages', {
+    include: ['@glimmer/**/*.ts'],
```
stefanpenner: are there non-test files in the '@glimmer/**/*' ?

package.json
```diff
@@ -18,7 +18,7 @@
     "pretest": "ember build",
     "test": "ember test",
     "test:ci": "ember test --test-port=7000 --config-file=testem-sauce.js && npm run test:node",
-    "test:node": "bin/run-node-tests.js",
+    "test:node": "bin/run-node-tests.sh",
```
stefanpenner: Does this work in windows?

testem-sauce.js
```diff
@@ -8,6 +8,10 @@ module.exports = {
   "parallel": 2,
   "launchers":
     {
+      "Node": {
+        "command": "./bin/run-node-tests.sh",
```
stefanpenner: does this work in windows?

testem.js
```diff
@@ -8,7 +8,7 @@ let config = {
   "disable_watching": true,
   "launchers": {
     "Node": {
-      "command": "./bin/run-node-tests.js",
+      "command": "./bin/run-node-tests.sh",
```
stefanpenner: ^

build/broccoli/build-tests.js
tomdale: @stefanpenner Not sure if I understand your question exactly. This includes all `.ts` files but not `.d.ts` files (which match `.ts`), so any non-TS files aren't included. What do you mean by "non-test files"?

package.json
tomdale: Don't shell scripts work in Windows now?

tomdale: I can rewrite it in JavaScript
*/me cries*

PR #597, opened 2017-07-22
[WIP] - Port ember component tests to new test harness

scalvert:
As per #561


packages/@glimmer/runtime/test/yield-test.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.render({
+      layout: 'In layout -- {{yield}}',
+      template: 'In template',
+    });
+
+    this.assertComponent('In layout -- In template');
+    this.assertStableRerender();
+  }
+
+  /*
```
scalvert: Any thoughts on a good way to translate these data driven tests into class-based tests, @chadhietala?

chadhietala: I would just unroll that loop.

chadhietala: You could also move the loop in a test case and call the test `primitive type asserts`

scalvert: OK - I'm inclined to go for the unroll approach, as it's more explicit what we're testing when we do that.

chadhietala:
Part of #614

PR #602, opened 2017-08-01
[BREAKING] Capital components

chadhietala:
# What Is This
- Removes the concept of dasherized angle bracket component invocations in favor of Capitalized angle bracket component invocations.
- Makes all components effectively fragments, (no need for a top-level element in the component's layout)
- Moves details of wrapping a component invocation the managers capabilities mask.
- Removes the `UnwrappedBuilder`
- Introduces syntax support for `...attributes`

## Wait...You're Removing Dasherized Components!?!?
Angle bracket components were originally theorized as a way to share some symantics of web components. One of those included using dasherized names for component invocations. While this sounded good in theory, it introduced this other problem and that is how do you differentiate between what is a web component and what is a Glimmer component. For various optimizations that we are in the midst of working on, it is required to remove ambiguity in the invocation.

By moving to capitalized component invocations this ambiguity no longer exists. It also allows developers to not be constrained by the requirement of having a dash in the name of the component. This allows you to create single word components, instead of doing `<x-link />` or even `<link- />` you can now just write `<Link />`.

Furthermore, this removes the requirement that the layout of a component requires you to have a wrapping element, for example you could not create components that acted as void elements like `<x-input />` without wrapping it.

```
<span>
   <input value="{{@value}}" type="{{@type}}" />
</span>
```

This PR treats the layout as a fragment so you can now do this.

```
<input value="{{@value}}" type="{{@type}}" />
```


packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -205,12 +206,18 @@ export default class VM implements PublicVM {
...
     self: PathReference<Opaque>,
+    args: Option<ICapturedArguments>,
     dynamicScope: DynamicScope,
     elementStack: ElementBuilder,
     symbolTable: ProgramSymbolTable,
     handle: Handle
   ) {
     let scope = Scope.root(self, symbolTable.symbols.length);
+
+    if (args) {
```
josemarluedke: This `if` seems to be unnecessary.

PR #606, opened 2017-08-06
Unify Broccoli build

tomdale:
Last time I refactored the build pipeline to emit per-package builds from Broccoli, it was split into two totally separate pipelines: one for testing and development, and a different one for doing "production" builds meant to be deployed to npm.

This split was unfortunate because it increased the complexity of the build code and didn't give us high confidence that the tests were passing against what would be the final production build of each package. Additionally, each package was compiled independently via the utilities in the `@glimmer/build` package, which was not designed with the monorepo scenario in mind.

In this PR, I unified the build pipeline and tried to address some of the major paint points with the previous system. Many of the improvements listed below are directly attributable to upgrading to v2 of `broccoli-typescript-compiler`.

## Better Performance

```
# master
yarn run build (cold) - 17.75s user 1.58s system 107% cpu 17.938 total
yarn run build (warm) - 16.04s user 1.29s system 111% cpu 15.571 total

# unified-build
yarn run build (cold) - 10.14s user 1.32s system 124% cpu 9.184 total
yarn run build (warm) - 9.46s user 1.17s system 115% cpu 9.171 total
```

## Clean Output
Previous versions of `broccoli-typescript-compiler` would spew errors because it didn't support the same library and dependency lookup logic as `tsc` itself. By upgrading to v2, we eliminate all of these kinds of spurious errors:

1 time yarn run build node 2017-08-06 12-10-51

## Compact Builds
Because we do a single TypeScript build for the entire project, rather than package-by-package, the compiler can perform smarter optimizations. In addition to being faster to build, it can do things like replace enums with their raw integer values, for example.

## Test Against Production
To improve build performance, we build test modules into individual AMD files and concatenate them together. Previously, we were also building every individual Glimmer VM module into its own AMD module, even though in production builds we use Rollup to build a single, scope-hoisted AMD module. Now we use the single package module (e.g. `@glimmer/runtime`). This necessitated changing some tests to import from the package directly instead of accessing nested files, since those paths are lost once Rollup gets done optimizing the package.

## Customizable Build Matrix
In test environments, we only build AMD and CommonJS ES5 versions of the packages. In production builds, we build the full suite of modules, types, CommonJS and AMD in ES5 and ES2017. There is a simple API for configuring which build combinations should be included:

```js
matrix = [
  ['amd', 'es5'],
  ['commonjs', 'es2017'],
  ['commonjs', 'es5'],
  ['modules', 'es2017'],
  ['modules', 'es5'],
  ['types']
];
buildPackages(jsTree, matrix);
```

## Easier Linking
If you wanted to test something locally by `npm link`ing one of the Glimmer VM packages, it was kind of a nightmare. I wrote a script that would do the symlinking for you, but Broccoli wasn't aware of it and it caused issues as soon as you rebuilt.

In this version, a small in-repo Ember CLI addon removes symlinks before a new build starts and sets them up again after the build has completed. Once you've `npm link`ed a package, you should be able to rebuild to your heart's content without any problems.

## Eslinted Build Tooling

While we have checking for the TypeScript packages, the build tooling did not have any kind of automated checking or linting. I added eslint just for the `build/` directory.

# Verified Working

- [x] `ember test` (with Node tests)
- [x] `ember test --serve` (with Node tests)
- [x] `ember s` (starts development server for running tests)
- [x] `ember build -prod` (builds the full output matrix)

# Open Questions

- [x] @krisselden Do the source maps look good?
- [x] @chadhietala This strips Glimmer debug/util functions from production builds. Is this what we want?


build/broccoli/transpile-to-es5.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const babel = require('broccoli-babel-transpiler');
+
+/**
+ * Transpiles a tree of ES2015+ JavaScript files to ES5 with Babel.
+ */
+module.exports = function transpileToES5(inputNode, modules = false) {
```
chadhietala: I feel like we want to externalize the babel helpers. Every file that uses class (which is a lot) is going to emit this.

```
function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }

function _possibleConstructorReturn(self, call) { if (!self) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return call && (typeof call === "object" || typeof call === "function") ? call : self; }

function _inherits(subClass, superClass) { if (typeof superClass !== "function" && superClass !== null) { throw new TypeError("Super expression must either be null or a function, not " + typeof superClass); } subClass.prototype = Object.create(superClass && superClass.prototype, { constructor: { value: subClass, enumerable: false, writable: true, configurable: true } }); if (superClass) Object.setPrototypeOf ? Object.setPrototypeOf(subClass, superClass) : subClass.__proto__ = superClass; }
```

build/broccoli/strip-glimmer-utilities.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Optimizes out Glimmer utility functions and strips debug code with a set of
+ * Babel plugins.
+ */
+module.exports = function(jsTree) {
+  return babel(jsTree, {
+    annotation: 'Babel - Strip Glimmer Utilities',
+    sourceMaps: 'inline',
+    plugins: [
+      [debugMacros, {
```
chadhietala: This is going to need a plugin to strip the `classCallChecks`. Likely you can move this [one I wrote](https://github.com/emberjs/ember.js/blob/master/broccoli/transforms/strip-class-call-check.js) to this guy https://github.com/ember-cli/ember-build-utilities and just use it directly.

chadhietala:
@tomdale It might be good to leverage or add to this repo https://github.com/ember-cli/ember-build-utilities. Its an attempt to have a single place that contains a lot of utilities for constructing broccoli build pipelines.

tomdale:
@chadhietala A lot of this stuff is vaguely similar to pipelines in `@glimmer/build`. I was planning to extract and upstream to that eventually, but wanted to get it solid here before prematurely extracting. The trickiest thing is that relative paths are very important, particularly with source maps, so you have to make sure your `Funnel` is dropping things into a consistent directory structure the whole way or things will get messed up. I noticed when trying to reuse pieces of `@glimmer/build` that things were designed for compiling a single package didn't translate directly for this monorepo use case.

ember-cli-build.js
```diff
@@ -1,16 +1,74 @@
...
+  // Second, compile all of the TypeScript into ES2017 JavaScript. Because the
+  // TypeScript compiler understands the project as a whole, it's faster to do
+  // this once and use the transpiled JavaScript as the input to any further
+  // transformations.
+  let jsTree = typescript(tsTree);
+
+  // Glimmer includes a number of assertions and logging information that can be
+  // stripped from production builds for better runtime performance.
+  if (PRODUCTION) {
+    // jsTree = stripGlimmerUtilities(jsTree);
```
chadhietala: Uncomment this guy

build/broccoli/transpile-to-es5.js
tomdale: @chadhietala and I talked off-thread and I persuaded him it was not worth optimizing the AMD builds because they are only used in tests.

PR #607, opened 2017-08-07
Intern Items In Constants Pool

chadhietala:
Prior to this commit we would store duplicate primitive types that are treated as immutable. This meant that the constants pool would grow unnecessarily and handout unique addresses for things that could be represented as single stored value. This should reduce the RSS and serialized size of the Constants Pool.


krisselden:
without ref count this means constants can't be unloaded with a template, I think that is ok for now just noting it

PR #610, opened 2017-08-08
[BREAKING] Decouple DOM From Env

chadhietala:
This PR does the following things:

- Makes ElementBuilders the encapsulated primitive for DOM in different environments
- Moves SerializeBuilder to `@glimmer/node`
- Moves things like `attributeFor` and `protocolForURL` to the builder instead of the environment, which cleaned up all the threading of the `env` we did.

## Important

The element builder functions are now to be used as the way of passing in a `Simple.Document` and `Cursor` (the `element` and `Option<nextSibling>` where construction begins).

```ts
type builder = (document: Simple.Document, Cursor) => ElementBuilder
```
There are 4 types of builder functions

**`elementBuilder`**: Used for client-side rendering starting from a blank document.
**`rehydratingBuilder`**: Used for client-side rendering starting from a serialized document
**`serializeBuilder`**: Used to server-side render a glimmer component with markers used for rehydration
**`stringBuilder`**: Used to server-side render a glimmer component without markers (think SEO)

### Usage

```ts
let root = document.createElement('div');
template.renderLayout({
  env: new Environment(),
  self: new UpdatableReference({...});
  elementBuilder: rehydratingBuilder(document, { element: root, nextSibling: null });
  dynamicScope: new TestDynamicScope()
})
```

In the future we may make the classes public so the factory functions are being used to just hide access to them.


bin/run-node-tests.js
```diff
@@ -10,6 +10,7 @@ const NODE_TEST_GLOB = '@glimmer/node/test/**/*-test.js';
 
 // When running inside `ember test`, we already have a build we can use.
 if ('EMBER_CLI_TEST_OUTPUT' in process.env) {
+  console.log("WAT THE FUCK!!!!!!!!!!");
```
chadhietala: @tomdale This isn't getting called at all

chadhietala:
Closing this for now. Partially implemented in #667.

PR #611, opened 2017-08-09
Use `execa`'s `preferLocal` flag to resolve local binaries.

rwjblue:


chadhietala:
Was superseded by #606

PR #614, opened 2017-08-09
Upgrade qunit

chadhietala:
This upgrades us to `qunitjs@^2.4.0`. To do this we needed to move away from private apis that were used in `testComponent` that were removed in `qunitjs@2.3.0`.

It's important to not that the assertion count is less here due to the fact that we used to [assert on the content](https://github.com/glimmerjs/glimmer-vm/compare/upgrade-qunit?expand=1#diff-0cb3427e6b926fffb8ba221f185735ceL391) during `rerender` when going through `testComponent`. In #580 we introduce more generic infrastructure for testing that does not do this assertion as it's can be called with non JSON serializable things like DOM nodes.

Thanks @scalvert and @chelseavalentine for getting this unblocked!


tomdale:
@chadhietala It seems like that extra assert in the old code was purely informational (I'm guessing to make it clear in test output which phase of the test other failures occurred in) so we're definitely not regressing on coverage at all.

packages/@glimmer/runtime/test/ember-component-new-test.ts
```diff
@@ -0,0 +1,113 @@
+import { module, RenderTests, test, classes } from "@glimmer/test-helpers";
```
tomdale: Should probably rename this to something other than `ember-component-new-test`? Maybe `ember-component-new-final-FINAL-2`?

packages/@glimmer/runtime/test/ember-component-test.ts
```diff
@@ -14,8 +14,8 @@ import {
   TestEnvironment,
```
tomdale: Nuke this file.

PR #616, opened 2017-08-10
Start location fix for dynamic attributes

### LINKED ISSUE #615: beginAttributeValue is not called for dynamic mustache attribute value
when parsing `` beginAttributeValue is never called. It does calls finishAttributeValue when it reaches `>` however. This results in the start location of the attribute value node to remain the default (line: 0, column: 0) which is not a valid location. (the first line is line 1)

This was discovered as the start location is not set correctly for the mustache statement, which made created an error in ember-tamplate-lint. See issue here: https://github.com/rwjblue/ember-template-lint/issues/253

initram:
Made sure to call beginAttributeValue when parsing a dynamic mustache attribute. Also added a small test to make sure that it works.

Resolves #615

### DIFF
--- a/packages/@glimmer/syntax/test/loc-node-test.ts
+++ b/packages/@glimmer/syntax/test/loc-node-test.ts
@@ -291,6 +291,18 @@ data-barf="herpy"
   }
 });
 
+test("element dynamic attribute", function() {
+  let ast = parse(`<img src={{blah}}>`);
+
+  let [div] = ast.body;
+  if (assertNodeType(div, 'ElementNode')) {
+    let [src] = div.attributes;
+    locEqual(src, 1, 5, 1, 17);
+    let value = src.value;
+    locEqual(value, 1, 9, 1, 17);
+  }
+});
+
 test("concat statement", function() {
   let ast = parse(`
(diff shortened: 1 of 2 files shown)


packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -291,6 +291,18 @@ data-barf="herpy"
   }
 });
 
+test("element dynamic attribute", function() {
+  let ast = parse(`<img src={{blah}}>`);
+
+  let [div] = ast.body;
```
rwjblue: Can you rename this to local variable to `img` (since it is referencing the `img` not a `div`)?

packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -291,6 +291,18 @@ data-barf="herpy"
...
+  if (assertNodeType(div, 'ElementNode')) {
+    let [src] = div.attributes;
+    locEqual(src, 1, 5, 1, 17);
+    let value = src.value;
```
rwjblue: Can likely use `let { value } = src;` here.

PR #620, opened 2017-08-14
Import tracked property code

tomdale:
This commit moves the tracked property code from `@glimmer/component` into the VM layer itself, and adds ES5 compatibility

### DIFF
--- a/packages/@glimmer/reference/index.ts
+++ b/packages/@glimmer/reference/index.ts
@@ -31,3 +31,9 @@ export {
   IteratorSynchronizer,
   IteratorSynchronizerDelegate
 } from './lib/iterable';
+
+export {
+  tracked,
+  tagForProperty,
+  UntrackedPropertyError
+} from './lib/tracked';
--- a/packages/@glimmer/reference/lib/tracked.ts
+++ b/packages/@glimmer/reference/lib/tracked.ts
@@ -1,0 +1,312 @@
+import { DEBUG } from '@glimmer/local-debug-flags';
+import { dict, Dict } from "@glimmer/util";
+
+import { Tag, DirtyableTag, TagWrapper, combine, CONSTANT_TAG } from './validators';
+
+/**
+ * @decorator
+ *
+ * Marks a property as tracked.
+ *
+ * By default, a component's properties are expected to be static,
+ * meaning you are not able to update them and have the template update accordingly.
+ * Marking a property as tracked means that when that property changes,
+ * a rerender of the component is scheduled so the template is kept up to date.
+ *
+ * @param dependencies Optional dependents to be tracked.
+ */
+export function tracked(...dependencies: string[]): MethodDecorator;
+export function tracked(target: any, key: any): any;
+export function tracked(target: any, key: any, descriptor: PropertyDescriptor): PropertyDescriptor;
+export function tracked(...dependencies: any[]): any {
+  let [target, key, descriptor] = dependencies;
+
+  if (typeof target === "string") {
+    return function(target: any, key: string | Symbol, descriptor: PropertyDescriptor) {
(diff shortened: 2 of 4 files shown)


tomdale:
One question I have:

Should the decorator implementation live here also, or just the primitives? We could allow the "application layer" of glimmer.js implement the actual user-facing decorator API.

tomdale:
@rwjblue Great questions, thank you for asking them and my apologies for not providing any context in the original PR. I'll give some specific context on this and general context that you already know but that may be useful for anyone following along at home.

The main goal of this change is to allow for better interop between different Glimmer implementations. The main one I'm thinking of is Ember.js and Glimmer.js components, but it applies to anyone who wants to create their own custom Glimmer VM environments.

The combination of [References](https://github.com/glimmerjs/glimmer-vm/blob/master/guides/04-references.md) and [Validators](https://github.com/glimmerjs/glimmer-vm/blob/master/guides/05-validators.md) allows for the host Glimmer environment to implement arbitrary change tracking semantics. Types of change tracking include Key-Value Observing or "KVO" (Ember), dirty checking (Angular 1), and component-level change tracking (React and `setState`).

Glimmer.js apps use a novel form of change tracking we call "tracked properties." Tracked properties are unique because they are a class-level annotation that simply tracks whether a property has been mutated.

While the goal of the references/validators abstraction is to allow composition of different change tracking systems, today Glimmer.js is hardcoded to tracked properties and Ember.js is hardcoded to the KVO of the Ember object model.

When a tracked property is mutated, it does two inexpensive operations:

1. It marks the property as being invalidated, or "dirty."
2. It checks to see if a revalidation has been scheduled, and if not, schedules it.

The way we accomplish the invalidation is by sharing a global revision counter and incrementing it when a possible change may have happened. Because Glimmer.js's tracked properties and Ember's KVO use validators exported from Glimmer VM (`DirtyableTag`, etc.) they all use the same shared revision counter and thus can be used together.

Step 2 is where we run into trouble today. The notion of "scheduling a revalidation" is environment-dependent at the moment. Ember.js relies on its runloop implementation and schedules revalidations at the end of every runloop. Glimmer.js has a global `propertyDidChange` hook that each app has to set to catch changes and revalidate itself. Ideally, we could have a shared abstraction for "please let me know when I should revalidate" on environments and "something may have changed, please revalidate" on change tracking implementations.

The other problem remaining for interop is making sure the right `Reference` instances make it into the VM in the first place. Every component has a "root" reference that all values in its template are derived from. Because the root reference comes from the component manager (via the `getSelf()` hook), at the moment every component is hardcoded to the set of references its component manager knows about.

For example, [Ember implements its own set of references](https://github.com/emberjs/ember.js/blob/master/packages/ember-glimmer/lib/utils/references.js) which are similar to the references in Glimmer.js which are similar to the references in the Glimmer VM test environment, and even though _validators_ are shared, references don't know about each other and cannot be composed.

To allow someone to use Ember objects in a Glimmer component, we need a shared set of references along with a protocol for detecting the appropriate reference for a given property. For example, Ember could register a handler that says "when the object is an Ember object, use a validator that observes property changes" and Glimmer.js could register a handler that says "if the descriptor for the property is for a tracked property, use a tracked property validator."

In retrospect, this PR probably puts the cart before the horse. In practice, any Glimmer VM environment will need a change tracking system and we currently believe that tracked properties are the fastest way to do that, so my thinking was that we should just bake it into Glimmer VM so it's there for people to use. However, after thinking about it a little more, it seems like the best option might be to work on the interop functionality I described above and make sure that Glimmer.js can implement tracked properties with that infrastructure. Once that's in place, we'll be better positioned to decide if incorporating tracked properties into Glimmer VM is still the right move.

PR #626, opened 2017-08-20
Remove all Opcode.toJSON references

chadhietala:
These are not used for anything as of right now.


rwjblue:
Hmm, are you certain this is unused? I believe that the stack VM debugging code uses this to emit useful information to the console.  It would be pretty bad to lose what little debuggability we have...

PR #629, opened 2017-08-22
[WIP] Experimental Opcode DCE

chadhietala:
This consolidates the side-effect modules to one place which we can overwrite at build time.

At build time we will know the exact Append opcodes we will execute at runtime. This means we can generate a custom `@glimmer/runtime/compiled/opcodes/index` at build time that clobbers the one we use to do development on the VM. The build would simply import only the operations needed at runtime and create a static operations array instead of one that is built at runtime. This will allow for things like rollup to shed the weight of the unused operations. Also in this mode we would nuke all the `* _MAPPINGS`

I would also argue that this makes moving around in the VM easier, meaning you have an actual symbol to lookup that is the same name as the operation. You will also have better debug-ability and ability to performance tune as you won't just see "anonymous function" in the profiler.

screen shot 2017-08-22 at 4 13 15 pm

### DIFF
--- a/packages/@glimmer/runtime/lib/bootstrap.ts
+++ b/packages/@glimmer/runtime/lib/bootstrap.ts
@@ -1,9 +1,1 @@
 import './opcodes';
-import './compiled/opcodes/expressions';
-import './compiled/opcodes/component';
-import './compiled/opcodes/content';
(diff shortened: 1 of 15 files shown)


chadhietala:
Closing this for now

PR #635, opened 2017-09-01
Load to saved register 0 is wrong

chadhietala:
Results in the following 
screen shot 2017-09-01 at 8 52 39 am


chadhietala:
Fixed in #644

PR #637, opened 2017-09-03
[Improvement] prefer for cloneNode instead of createElement

lifeart:
https://jsperf.com/clonenode-vs-createelement-performance/32

### DIFF
--- a/packages/@glimmer/node/lib/node-dom-helper.ts
+++ b/packages/@glimmer/node/lib/node-dom-helper.ts
@@ -23,9 +23,25 @@ export default class NodeDOMTreeConstruction extends DOMTreeConstruction {
     return new ConcreteBounds(parent, first, last);
   }
 
+  createNodeCache(tag: string) {
+    if (!this.document.nodesCache) {
+      this.document.nodesCache = {};
+    }
+    this.document.nodesCache[tag] = this.document.createElement(tag);
+    return this.getNodeFromCache(tag);
+  }
+  
+  hasNodeInCache(tag: string) {
+    return this.document.nodesCache && this.document.nodesCache[tag];
+  }
+  
+  getNodeFromCache(tag:string) {
+    return this.document.nodesCache[tag].cloneNode(false);
+  }
+
   // override to avoid SVG detection/work when in node (this is not needed in SSR)
   createElement(tag: string) {
-    return this.document.createElement(tag);
+    return this.hasNodeInCache(tag) ? this.getNodeFromCache(tag) : this.createNodeCache(tag);
   }
 
   // override to avoid namespace shenanigans when in node (this is not needed in SSR)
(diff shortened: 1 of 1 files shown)


rwjblue:
Looks like current CI failure is due to types (you can replicate locally via `yarn run problems`):

```
$ yarn run problems
yarn run v0.27.5
$ tsc -p tsconfig.json --noEmit
packages/@glimmer/node/lib/node-dom-helper.ts(27,24): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
packages/@glimmer/node/lib/node-dom-helper.ts(28,21): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
packages/@glimmer/node/lib/node-dom-helper.ts(30,19): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
packages/@glimmer/node/lib/node-dom-helper.ts(35,26): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
packages/@glimmer/node/lib/node-dom-helper.ts(35,54): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
packages/@glimmer/node/lib/node-dom-helper.ts(39,26): error TS2339: Property 'nodesCache' does not exist on type 'Document'.
error Command failed with exit code 1.
```

rwjblue:
Also, we probably will want to run some macro benchmarks against this to see what impact it has (positive or negative).

lifeart:
https://jsperf.com/createelement-vs-clonenode-fragment-cache

rwjblue:
Definitely appreciate the micro benchmark, but we will need to test this in a macro benchmark also before we can land it. In other words, we need to see how this affects real apps in addition to these sorts of tightly constrained micro benchmark scenarios.

tomdale:
I don't think this approach ends up being faster. The microbenchmark does not capture certain aspects of real-world behavior, and given how close the two are (node cache version is ~5% faster on my machine), I wouldn't be surprised if it is actually slower, for these reasons:

1. The microbenchmark does not include the `hasOwnProperty` check to see if the node is already in the dictionary.
2. The microbenchmark does not use a large enough variety of HTML element types to defeat some optimizations that are probably happening here, e.g. putting the node cache into dictionary mode, functions becoming megamorphic, etc.

When a microbenchmark shows me something surprising, I always try to find an explanation for the difference, because of how difficult it is to construct representative synthetic benchmarks. (I even [wrote a blog post about it](https://tomdale.net/2017/07/adventures-in-microbenchmarking/).

In this case, the question I ask myself is: if using `cloneNode(false)` for creating a new, empty element is faster, why would the browser not use this path itself? The performance of `document.createElement()` is definitely in the hot path of many of the browser benchmarks, so why would this not have been optimized?

tomdale:
Gonna close this until we have better benchmarks. Thank you for the PR!

PR #639, opened 2017-09-06
Add stack assertions

chadhietala:
Prior to this commit we were sort of flying blind when it came to making sure the stack was balanced. That being said as we have done larger refactors the integrity of the stack has become less obvious. To ensure the integrity we have added type checking and runtime checks in all of the code that is responsible for pushing and popping from the stack. We also asserting the type at the head of the stack upon exit of the operation and how many positions the stack has moved sense entering the operation. This gives us the ability to ensure the stack is balanced and we are working off of the correct values. 

In production all of this tooling is stripped using [nukable-import](https://github.com/chadhietala/babel-plugin-nukable-import).


packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -22,14 +22,14 @@ import {
...
-import { Macros, OpcodeBuilderConstructor, VMHandle } from "@glimmer/opcode-compiler";
+import { Macros, OpcodeBuilderConstructor, VMHandle, ICompilableTemplate } from "@glimmer/opcode-compiler";
 import { IArguments } from './vm/arguments';
 import { Simple, RuntimeResolver, BlockSymbolTable } from "@glimmer/interfaces";
 import { Component, ComponentManager } from "@glimmer/runtime/lib/internal-interfaces";
 import { Program } from "@glimmer/program";
 
-export type ScopeBlock = [VMHandle, BlockSymbolTable];
-export type ScopeSlot = VersionedPathReference<Opaque> | Option<ScopeBlock>;
+export type ScopeBlock = [VMHandle | ICompilableTemplate<BlockSymbolTable>, BlockSymbolTable];
```
mmun: :+1:

PR #662, opened 2017-09-14
Refactor call protocol

mmun:
- Moves blocks into the call protocol
- Adds a new return value register `v0`
- Simplify the stack checks

### DIFF
--- a/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
+++ b/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
@@ -324,8 +324,11 @@ export abstract class OpcodeBuilder<Specifier> {
     this.push(Op.Comment, comment);
   }
 
-  modifier(specifier: Specifier) {
+  modifier(specifier: Specifier, params: Option<WireFormat.Core.Params>, hash: Option<WireFormat.Core.Hash>) {
+    this.pushFrame();
+    this.compileArgs(params, hash, null, true);
     this.push(Op.Modifier, this.constants.handle(specifier));
+    this.popFrame();
   }
 
   // lists
--- a/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
+++ b/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
@@ -81,6 +81,12 @@ export interface CompileTimeLookup<Specifier> {
   lookupPartial(name: string, referrer: Specifier): Option<number>;
 }
 
+export interface Blocks {
+  main: Option<CompilableBlock>;
+  else: Option<CompilableBlock>;
+  attrs: Option<CompilableBlock>;
+}
+
 export interface OpcodeBuilderConstructor {
   new<Specifier>(program: CompileTimeProgram,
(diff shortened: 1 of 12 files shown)


packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
```diff
@@ -587,9 +596,26 @@ export abstract class OpcodeBuilder<Specifier> {
...
+
     let count = this.compileParams(params);
 
+    let flags = count << 4;
+
+    if (synthetic) flags |= 0b1000;
+
+    if (blocks) {
+      flags |= 0b111;
+      // if (blocks.main) flags |= 0b0100;
```
chadhietala: Remove?

packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
```diff
@@ -733,16 +759,19 @@ export abstract class OpcodeBuilder<Specifier> {
 
     let { symbols } = symbolTable;
 
+    // let blocks = { main: block, else: inverse, attrs };
```
chadhietala: Remove

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -116,21 +116,20 @@ APPEND_OPCODES.add(Op.IsComponent, vm => {
...
 
   let meta = vm.constants.getSerializable<TemplateMeta>(_meta);
   let resolver = vm.constants.resolver;
-  let definition = check(stack.pop(), CheckReference);
 
-  stack.push(new CurryComponentReference(definition, resolver, meta, captured));
+  vm.loadValue(Register.v0, new CurryComponentReference(definition, resolver, meta, captured));
 
-  expectStackChange(vm.stack, -args.length - 1, 'CurryComponent');
+  // expectStackChange(vm.stack, -args.length - 1, 'CurryComponent');
```
chadhietala: This likely needs to be brought back and incorporate the new `pop` from the stack.

packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
```diff
@@ -587,9 +596,26 @@ export abstract class OpcodeBuilder<Specifier> {
...
+  compileArgs(params: Option<WireFormat.Core.Params>, hash: Option<WireFormat.Core.Hash>, blocks: Option<Blocks>, synthetic: boolean) {
+    if (blocks) {
+      this.pushYieldableBlock(blocks.main);
+      this.pushYieldableBlock(blocks.else);
+      this.pushYieldableBlock(blocks.attrs);
+    }
+
     let count = this.compileParams(params);
 
+    let flags = count << 4;
```
chadhietala: I'm a bit worried that the encoding we are introducing is spread between the builder and the encoder. I'm wondering if we should move all the encoding to the encoder so we have a better idea of heap layout and the available and reserved size of different types of operands.

mmun:
Merging as is, but @wycats will circle back on the comments later.

PR #663, opened 2017-09-14
Component terminology cleanup

tomdale:
There were a few pieces of terminology that had emerged organically that we had been wanting to revisit. Generally speaking, it felt like time to do a naming/terminology pass and try to reflect symmetry and other relationships between objects in naming, where possible.

First, I apologize for the size of this PR but it touches some low-level data structures and thus spread rather far. As soon as I realized how long the thread I was tugging at would be once unraveled, I tried to scope it down as fast as I could. But this PR does include some cleanup to the test infrastructure that, while not strictly necessarily related, was convenient to work on while I was in there anyway.

### Glossary

Like any compiler and VM, Glimmer is rather complicated. To help contributors get more comfortable with the codebase, I've included a new Glossary entry in the guides that tries to explain some of the names and concepts they might encounter.

Some entries are shorter than they should be, and others are waaaaay too verbose. In those cases, I figured it was better to err on the side of too much information. All of these subjects really need their own guide, so in the future we can extract verbose entries into a proper guide, and replace them with something terser along with a link to more information.

### VM/Component Terminology

* `ComponentSpec` becomes `ComponentDefinition`, the data structure needed by the runtime to invoke a new instance of a component.
* `ComponentDefinition` becomes `ComponentDefinitionState`, the opaque state bucket provided by the host environment with each `ComponentDefinition` that is passed to the appropriate component manager hooks.
* `ComponentState` becomes `ComponentInstance`, the data structure used by the VM to represent every component instance it creates. (Unlike `ComponentDefinition`s, this one doesn't escape the VM and doesn't impact host implementations.)
* `Component`/`ComponentStateBucket` becomes `ComponentInstanceState`, the opaque state bucket returned from the component manager's `create()` hook and provided back in subsequent component manager hooks.

Note that these state types (ComponentDefinitionState and ComponentInstanceState) are internal and the actual value is opaque to the outside world. Host environments can put any value, or no value, in the state slot.

### Testing Terminology

In the beginning, we had only what we now call the lazy compiler. With the introduction of the eager "bundle" compiler, we now support two different modes of compilation. While the tests were updated to test both paths, the naming itself was not.

This PR attempts to break out more of the testing infrastructure for each "mode" into multiple files and align naming between them, so the symmetry (and, occasionally, lack thereof) is more apparent to the reader.

Additionally, I have tried to be explicit about including either `Lazy` or `Eager` in names to make it clear which mode you're getting, e.g. `TestEnvironment` became `LazyTestEnvironment` and `BundledClientEnvironment` became `EagerTestEnvironment`.


rwjblue:
I'm pretty opposed to more `instance instances`, can we pick something else for `ComponentInstance`?

> Wait, are we talking about component instances or ComponentInstance instances?!?!!?!?!? 🤕

tomdale:
@rwjblue I understand your concern and I thought of you immediately as I changed the name. However, I think it's important to understand the subtle differences between this case and the `ApplicationInstance` case so we don't *ahem* overrotate and incorrectly pattern match against the most appropriate name.

In particular:

1. Unlike `ApplicationInstance`, `ComponentInstance` is a POJO, not a class. You can never instantiate it, so there can never be a `ComponentInstance` instance.
2. Unlike `ApplicationInstance`, this is an internal name that is constrained to a single file and doesn't surface to host environment implementors, let alone app/component developers. We can afford to optimize for clarity in the specific context in which its used.
2. There is no other object created by the VM, so as far as the VM is concerned, this data structure *is* the canonical representation of an instance of a component.
3. Unlike `ApplicationInstance`, where "instance" confusingly referred to a class because of unfortunate backwards-compatibility constraints, in this case the `ComponentInstance` *is* an instance in the system the VM is responsible for managing. Glimmer is implementing a runtime and, I note, [V8 has a `LazyInstance` class to describe JavaScript instances](https://github.com/v8/v8/blob/a580b2f/src/base/lazy-instance.h#L5-L23), and this doesn't strike me as confusing even though C++ has classes and instances as well.

That said, if there is a name that better describes the data structure, I'm eager to adopt it.

packages/@glimmer/interfaces/README.md
```diff
@@ -0,0 +1,17 @@
+# @glimmer/interfaces
```
chadhietala: Maybe remove this?

tomdale: Good call.

PR #668, opened 2017-09-18
Smoke test types before publishing

tomdale:
Before publishing, smoke test that each package can be imported without type errors. This should probably happen as part of the test suite, but unfortunately it requires on symlinking node_modules dependencies across packages which only happens *after* the build has completed, and thus is incompatible with `ember test`.


rwjblue:
@tomdale - Please add a `scripts` entry for this in `package.json`, and then add it to `.travis.yml`.

tomdale:
@rwjblue Great catch, thank you. Will do.

PR #674, opened 2017-09-20
Use simple-html-tokenizer's built-in types

tomdale:
Previously, we were using a version of simple-html-tokenizer that did not ship with type annotations, so we included our own in a .d.ts file.

We've since upgraded to a newer version of SHT that includes types, but we did not remove our custom declaration. This commit removes the conflicting declaration and fixes some type errors that resulted from types that diverged. It also updates each package's dependency to match the version of SHT at the project root.


chadhietala:
I'm "OK" with this sense we aren't maintaining two copies, but we come to terms with the fact that our html parser is using "intimate" APIs to effectively to make this project work. This project seems to be evidence that those "private" APIs actually need a public API.

wycats:
@chadhietala strongly agree

PR #679, opened 2017-09-29
[BUGFIX] Node is not a thing in SSR

chadhietala:
This addition runs tests in SSR in eager and lazy modes, with and without the serialization builder.

### DIFF
--- a/packages/@glimmer/node/lib/serialize-builder.ts
+++ b/packages/@glimmer/node/lib/serialize-builder.ts
@@ -2,6 +2,8 @@ import { NewElementBuilder, ElementBuilder, Bounds, ConcreteBounds, Environment 
 
 import { Simple, Option } from "@glimmer/interfaces";
 
+const TEXT_NODE = 3;
+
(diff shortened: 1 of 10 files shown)


packages/@glimmer/node/lib/serialize-builder.ts
```diff
@@ -39,7 +41,7 @@ class SerializeBuilder extends NewElementBuilder implements ElementBuilder {
 
     if (string === '') {
       return this.__appendComment('%empty%') as any as Simple.Text;
-    } else if (current && current.nodeType === Node.TEXT_NODE) {
+    } else if (current && current.nodeType === TEXT_NODE) {
```
krisselden: simple-dom now has a const enum for this, which will just be replaced with the number and a comment

chadhietala: That was what I was hoping but it is not exposed through the public API.

chadhietala: https://github.com/ember-fastboot/simple-dom/issues/45

PR #680, opened 2017-09-29
Remove support for IE9/IE10/Phantom

chadhietala:
This is the initial pass to remove IE9/IE10/Phantom support. See https://github.com/emberjs/rfcs/pull/252 for details.


packages/@glimmer/runtime/lib/dom/props.ts
```diff
@@ -46,14 +46,7 @@ export function normalizePropertyValue(value: Opaque): Opaque {
...
 const ATTR_OVERRIDES = {
 
-  // phantomjs < 2.0 lets you set it as a prop but won't reflect it
-  // back to the attribute. button.getAttribute('type') === null
-  BUTTON: { type: true, form: true },
-
   INPUT: {
-    // Some version of IE (like IE9) actually throw an exception
-    // if you set input.type = 'something-unknown'
-    type: true,
```
rwjblue: Can you confirm we have tests for this case (that would run under all of our supported browsers)?  I'm slightly concerned that this affected more browsers than just IE9...

testem.js
```diff
@@ -12,12 +12,20 @@ let config = {
       "protocol": "tap"
      }
   },
+  "browser_args": {
+    "Chrome": [
```
rwjblue: We can modify this so that it only does headless in CI (and in dev it does a normal chrome tab).  See https://twitter.com/trentmwillis/status/878277832278802432.

packages/@glimmer/runtime/lib/dom/props.ts
chadhietala: Added a test to make sure that the browser doesn't throw.

rwjblue: ❤️ thank you!

PR #682, opened 2017-09-30
[BUGFIX] Block Rehydration

chadhietala:
This continues the integrity checks for rehydration and "fixes" issues with blocks. Nested blocks would not clear out and blocks that opened on the client would be inserted past the end block markers. While this didn't seem fatal, was pretty weird that the metamorphs were retained.


chadhietala:
There appears to be more issues and/or this is not correct.

Issue #687, opened 2017-10-03
Demos not working


cibernox:
This project contains a demos folder that at some point stopped working. 
The html files reference assets on folders that no longer exist.

locks:
I believe demos were moved to https://github.com/glimmerjs/glimmer-demos ?

CvX:
Yep, all demos (working or not 😉) have been moved to https://github.com/glimmerjs/glimmer-demos. The outdated demos in this repo should probably be removed.

cibernox:
I confirm that the demos in that repo do work, although the one I was interested in, the Glimmer Visualizer, does not show the OpCodes like it did at some point (https://www.youtube.com/watch?v=vL8sCi1Bv6E&t=162s)

CvX:
That's correct, visualizer is currently no-op, see: https://github.com/glimmerjs/glimmer-demos/issues/7 and a note in https://github.com/glimmerjs/glimmer-demos/pull/5

CvX:
Btw. you can get the last working visualizer version from glimmer-vm here: https://github.com/glimmerjs/glimmer-vm/pull/526

cibernox:
@CvX Do you recall some version or point in time this did work? I need the opcodes to illustrate the slides of a talk I'm preparing.

cibernox:
@CvX you were fast! I tried that branch, but I can't seem to get it compiled. It fails during `yarn`, on the `prepublish` compilation
Do I have to use a specific version of node and/or typescript?

CvX:
I've used node 7.x (v7.10.0 specifically).
Then just `yarn && yarn run ember build && open dist/tests/demos/index.html` 😃

PR #688, opened 2017-10-03
Remove Demos And Benches From Repo

chadhietala:
These have had a very hard time being maintained in the repo. We decided it's best to move them out and into their own repositorys. You can find them at:

https://github.com/glimmerjs/glimmer-demos
https://github.com/glimmerjs/glimmer-bench

#687


rwjblue:
Restarted "push" CI job.

PR #690, opened 2017-10-04
Fix Rehydration

chadhietala:
Prior to this commit rehydration had several issues with clearing blocks. In several cases
what would end up happening is that the block depth became invalid and those we would attempt
to remove already attached nodes. This PR uses a `RehydratingCursor` as a means to track where
we are at in rehydration and also if we are rehydrating or clearing/repairing the block.

This PR almost has a better fix for common omitted tags like `<tbody />` by controlling exactly
when to push and pop it off of the CursorStack. In the future we should consider a solution that
emits the tags at compilation time instead of serialization.


testem.js
```diff
@@ -18,8 +18,7 @@ let config = {
       "args": [
         '--disable-gpu',
         '--headless',
-        '--remote-debugging-port=9222',
-        '--window-size=1440,900'
+        '--remote-debugging-port=9222'
```
Turbo87: @chadhietala why did you change this?

.travis.yml
```diff
@@ -2,16 +2,21 @@ language: node_js
 node_js:
   - 6
 dist: trusty
-sudo: false
+sudo: required
```
Turbo87: why is `sudo` mode required now?

testem.js
chadhietala: Doesn't seem required to have the window size for headless tests.

.travis.yml
chadhietala: I don't remember.

testem.js
Turbo87: Depends on what you're testing. For ember apps it's the default because otherwise media queries won't work properly because the window size is 0x0 px by default.

PR #694, opened 2017-10-06
Skip setting attributes if backing reference is not dirty

tomdale:
Fixes glimmerjs/glimmer.js#9. Previously, we were always eagerly setting element attributes even in cases where the backing reference's tag had not been invalidated.

I think we didn't notice this for two reasons:

1. Most attributes sets are idempotent, and the bug is only really noticeable in cases where you're setting things like the `src` attribute of an audio tag.
2. More horrifying, it turns out that the majority of the test suite is built on top of volatile tags. I tried to change these to something a bit more reasonable but it caused hundreds of cascading failures because many tests rely on interior mutation of template context. It's hard to guess exactly how many bugs the volatility is hiding, but we should get in the habit of testing re-renders with both valid and invalid references.


wycats:
lgtm :+1:

PR #695, opened 2017-10-09
Cleanup bundle compiler

chadhietala:
Prior to this change the compiler delegate needed shared state from the BundleCompiler and BundleCompilerLookup to work backwards into a CompilableTemplate. This meant that we had to pass in a `compile` function that closed over the BundleCompiler so that the delegate could compile a specifier. All of this is not neccessary as the lookup can simply pass all the information directly to the delegate, in the event it cannot find a compilable template. By default the delegate can just return an instance of `CompilableTemplate` because everything is now in scope.
ea99e76

### DIFF
--- a/bin/run-types-tests.js
+++ b/bin/run-types-tests.js
@@ -12,7 +12,7 @@ try {
   console.log('TAP version 13');
   console.log('1..1');
   console.log('# Smoke testing types');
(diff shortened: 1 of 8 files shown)


chadhietala:
Talked to Tom about this in Slack. He was +1 once I added an `expect` to make sure we actually have a WF prior to calling the delegate.

PR #697, opened 2017-10-10
Fix Memory Leaks

chadhietala:
This fixes the memory leaks that we were running into. While the test suite infra has been stable for a while we have been adding more and more tests which have actually been leaking instances of the test case for each test. It appears that in IE11 we ran into out of memory limits much sooner than the other browsers. Furthermore, this uses the `qunit-fixture` for a large majority of tests as the place we are appending to instead of doing all of the testing in an orphaned element. The only cases where we do not do this is where we are simulating SSR on the client. We should continue to try to move more of the infra to use this pattern.

### DIFF
--- a/packages/@glimmer/runtime/test/attributes-test.ts
+++ b/packages/@glimmer/runtime/test/attributes-test.ts
@@ -18,7 +18,7 @@ function compile(template: string) {
 
 function commonSetup() {
   env = new TestEnvironment(); // TODO: Support SimpleDOM
-  root = document.createElement('div');
+  root = document.getElementById('qunit-fixture')!;
   root.setAttribute('debug-root', 'true');
 }
 
--- a/packages/@glimmer/runtime/test/attributes-test.ts
+++ b/packages/@glimmer/runtime/test/attributes-test.ts
@@ -18,7 +18,10 @@ function compile(template: string) {
 
 function commonSetup() {
   env = new TestEnvironment(); // TODO: Support SimpleDOM
-  root = document.getElementById('qunit-fixture')!;
+  let fixture = document.getElementById('qunit-fixture')!;
+  let div = document.createElement('div');
+  fixture.appendChild(div);
+  root = div;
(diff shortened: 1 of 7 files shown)


packages/@glimmer/runtime/test/attributes-test.ts
```diff
@@ -18,7 +18,7 @@ function compile(template: string) {
 
 function commonSetup() {
   env = new TestEnvironment(); // TODO: Support SimpleDOM
-  root = document.createElement('div');
+  root = document.getElementById('qunit-fixture')!;
```
rwjblue: FWIW, I think this is not quite right.  There is an issue with qunits cleanup mechanism where attributes added to `qunit-fixture` are not properly cleared (they set inner HTML but do nothing to clear attributes).

I think what you want is:

```js
let testingRoot = document.getElementById('qunit-fixture');
root = testingRoot.createElement('div');
```

packages/@glimmer/runtime/test/partial-test.ts
```diff
@@ -23,7 +23,7 @@ function compile(template: string) {
 
 function commonSetup() {
   env = new TestEnvironment();
-  root = document.createElement('div');
+  root = document.getElementById('qunit-fixture')!;
```
rwjblue: Unclear if this can result in attribute changes, if not then this is totally good (no need to add a nested child).

packages/@glimmer/runtime/test/attributes-test.ts
rwjblue: For reference, this is done here: https://github.com/qunitjs/qunit/blob/bac08a415e514649d9ff30b82f5101c7fec739b0/runner/fixture.js#L37

packages/@glimmer/runtime/test/partial-test.ts
chadhietala: Yea there shouldn't be any attribute setting.

PR #700, opened 2017-10-11
Um... this should have been removed

chadhietala:


tomdale:
How was this not causing a type error?

Edit: Whoops, it's an enum, not an export. Granddad should put on his reading glasses before commenting.

rwjblue:
👴 👓

PR #701, opened 2017-10-11
Initial work towards the renderComponent API

chadhietala:
This introduces the initial work to pass args directly to the top level component. While this implementation is somewhat crude, it exposes all of the primitives need to synthetically create args for the the top level component. For example in Glimmer.js the user facing API for this would be:

```
renderComponent('MyApp', document.body, { name: 'Chad' });
```

Internally we would synthetically turn `{ name: 'Chad' }` into args and thus virtualize the invocation like so:

```
<MyApp @name="Chad" />
```

In the layout you could then refer to the name argument as `@name`.

### DIFF
--- a/packages/@glimmer/bundle-compiler/lib/bundle-compiler.ts
+++ b/packages/@glimmer/bundle-compiler/lib/bundle-compiler.ts
@@ -8,7 +8,8 @@ import {
   Recast,
   SymbolTable,
   VMHandle,
-  ComponentCapabilities
+  ComponentCapabilities,
+  Unique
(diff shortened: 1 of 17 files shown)


rwjblue:
Some of the details elude me (I’m not too familiar with the code here), but I am curious what you expect the API to be from the Glimmer.js side around how to update the initial values?

Happy to chat offline about that though, and it certainly shouldn’t block anything here...

jwlawrence:
I know this is really old and I see that this was merged, but looks like `renderComponent` never updated its signature to allow for passing args. Is there a solution for this now?

PR #707, opened 2017-10-13
Fix path lookup inside helpers

### LINKED ISSUE #705: Using if in partials does not work as expected
# Related Issues
- https://github.com/emberjs/ember.js/issues/15621

# Steps to reproduce
When you use a `{{#if valueOfParentContext}}` statement in a partial the value gets `undefined`.

For example:
```handlebars
// main.hbs
{{#with 'Dennis' as |myAuthor|}}
  {{ partial 'foo' }}
{{/with}}
```

```handlebars
// foo.hbs
{{#if myAuthor}}
  Author: {{ myAuthor }}!
{{else}}
  No author!
{{/if}}
```

# Expected result
The rendered Template should be `Author: Dennis!`.

# Actual result
The rendered Template is `No author!`.

# Comments
I've added a failing [test](https://github.com/liemak-it/glimmer-vm/commit/78358e120c62f3b2d1d153873754fb9bb8a7a00f).

tomdale:
Fixes #705 

MustacheExpressions generate a `get` opcode in cases where the identifier is statically found in their scope, but fall back to `maybeLocal` otherwise to disambiguate at compile time whether they are being invoked inside a partial or not.

PathExpressions however did not go through this path, always using a `get` opcode even in cases that were ambiguous. This caused any PathExpression inside a partial to fail when inheriting an identifier from the outside scope.

This PR changes the PathExpression compiler to generate a `maybeLocal` opcode in the ambiguous case instead of a `get`.

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -268,10 +268,10 @@ export default class TemplateCompiler {
 
       if (expr.this) {
         this.opcode('get', expr, 0, expr.parts);
-      } else  if (symbols.has(head)) {
+      } else if (symbols.has(head)) {
         this.opcode('get', expr, symbols.get(head), expr.parts.slice(1));
       } else {
-        this.opcode('get', expr, 0, expr.parts);
+        this.opcode('maybeLocal',expr, expr.parts);
       }
     }
   }
--- a/packages/@glimmer/runtime/test/partial-test.ts
+++ b/packages/@glimmer/runtime/test/partial-test.ts
@@ -463,6 +463,19 @@ QUnit.test('dynamic partial with local reference (unknown)', () => {
   equalTokens(root, `You smaht. You loyal. `);
 });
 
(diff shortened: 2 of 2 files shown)


chancancode:
Looks good to me. As far as I can tell, we just forgot to do it in https://github.com/glimmerjs/glimmer-vm/commit/702d1de431c5898e3a4cec2ff784f63adeb045a7

GavinJoyce:
This change seems to have caused a regression in the Ember test suite: https://github.com/emberjs/ember.js/pull/15777#issuecomment-339940275

screen shot 2017-10-27 at 11 42 38


https://github.com/emberjs/ember.js/blob/5f4ac602813d8fbd78f26fc955acf693e0b3d62a/packages/ember-glimmer/tests/integration/components/curly-components-test.js#L331-L351

tomdale:
@GavinJoyce The regression was fixed on the Ember.js side in emberjs/ember.js#15777, right, and there is nothing we need to do on the Glimmer VM side?

GavinJoyce:
@tomdale yes, this was fixed in Ember

PR #708, opened 2017-10-13
Check if attribute value actually changed before setting it

t-sauer:
Solves the issue described in https://github.com/glimmerjs/glimmer.js/issues/9#issuecomment-336291506


tomdale:
@t-sauer @chadhietala I don't think this is the correct fix. I am extremely nervous about reading from the DOM in the middle of the render cycle because it can force layout and other side effects we really want to avoid.

If we want to do this I think we should cache the last value in the `DynamicAttribute` itself and check against that.

PR #710, opened 2017-10-15
Cache value instead of reading from the element

chadhietala:
Fixes concerns brought up in #708


packages/@glimmer/runtime/lib/vm/attributes/dynamic.ts
```diff
@@ -81,18 +81,20 @@ export class SimpleDynamicAttribute extends DynamicAttribute {
...
   update(value: Opaque, _env: Environment): void {
     let { element, name } = this.attribute;
 
-    if (element[name] !== value) {
-      element[name] = value;
+    if (this.value !== value) {
+      element[name] = this.value = value;
     }
 
     if (value === null || value === undefined) {
```
tomdale: Seems like we probably want to avoid calling `this.removeAttribute()` if the value is unchanged as well, right?

PR #711, opened 2017-10-16
fix(bundle-compiler): Properly cache compileOptions

tomdale:
The compileOptions object gets passed around to compilable templates and includes relatively-expensive-to-construct objects like BundlingLookup. To avoid doing lots of repeat work, this object was being cached after the first time it was created.

Unfortunately, this object is not cacheable because variants need to be produced for each specifier/asPartial combination. This lead to a bug where the same referrer was always passed to component delegate methods (it was always the first template to be added).

This commit changes the caching strategy to cache a base "template options" object that contains the expensive objects, which gets used to initialize a new, lightweight CompileOptions object tailored to the requested specifier and asPartial setting.


chadhietala:
+1 my bad.

PR #657, opened 2017-09-11
[WIP] Introduce Golden tests for asserting on opcodes and constant pool

chadhietala:
This introduces the notion of [golden tests](https://kseo.github.io/posts/2016-12-15-golden-tests-are-tasty.html) for asserting on the bytecode that is emitted by the bundle compiler. This is largely inspired by [v8's golden test suite](https://github.com/v8/v8/tree/master/test/cctest/interpreter/bytecode_expectations). In practice this allows for us to assert that were are emitting the correct sequence of low-level instructions given a high level language, in our case Handlebars. The next step is to add more tests.

**Failing Test Example**

screen shot 2017-09-11 at 2 58 46 pm

PR #714, opened 2017-10-18
[BUGFIX] Don't continuously recompile the layout

chadhietala:
Prior to this fix we were compiling the layout and returning which meant
the next time we see an invoke we wouldn't just compile the args, but also
compile the layout despite the compilation beind exactly the same e.g.
templates are a pure function.


chadhietala:
We need something like #657 to catch these types of bugs.

PR #718, opened 2017-10-25
[BUGFIX] Fix Frozen Args When Currying

chadhietala:
In #712 we placed a "WELL_KNOWN_EMPTY_ARRAY" into the constants pool to canonicalize
the empty arrays. This allowed us to reduce the number of empty arrays to 1. That being
said it appears we were relying on the mutability of these empty arrays in the constant
pool. This is because an ambigous invocation will start out with 0 args but could have
been yielded with pre-loaded args. We don't have the introspection (today) to realize
this has occured. To fix this when we realize this at runtime we should be able to swap
back in a mutable array when we discover there are more args for the invoke.

Fixes https://github.com/glimmerjs/glimmer.js/issues/40.

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/arguments.ts
+++ b/packages/@glimmer/runtime/lib/vm/arguments.ts
@@ -373,6 +373,10 @@ export class NamedArguments implements INamedArguments {
       let { names, length, stack } = this;
       let { names: extraNames } = other;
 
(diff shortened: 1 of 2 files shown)


rwjblue:
@chadhietala - Can you confirm that this is intended to fix glimmerjs/glimmer.js#40?

chadhietala:
@rwjblue Yes.

PR #726, opened 2017-10-27
[BUGFIX beta] Fix range element reporting wrong initial value

Serabe:
If a template had an `input` `range` whose type was set before a max or a min
attribute was added, the initial value would be wrong.

In plain HTML, the initial value for `<input max="10" type="range" />` would be
`5` (`(min + max) / 2 = (0 + 10) / 2 = 5`).

In Glimmer, this would be 10 because the code would be equivalent to:

```js
let input = document.createElement('input');
input.type = "range";
input.max = "10";
```

Setting `type` to `range` would make the value sanitization algorihtm to kick
in, setting the value to 50. When setting the `max` to `10`, the algorithm
would sanitize the value (50 by now) to the maximum (10).

This was originally discussed in #425, but an alternative solution was
asked for. While this PR keeps part of the original implementation
(could not find a simple way to fix the issue when a template is being
rendered), this PR goes further and fixes it for Emberish components and
`...attributes` in Glimmer components.

More info in whatwg/html#2427
Fixes emberjs/ember.js#14958

While currently the tests contains all the permutations, the final state will only contain tests for the ordering `type`, `value`, `min max`.

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -1,7 +1,7 @@
 import TemplateVisitor, { SymbolTable, Action } from "./template-visitor";
 import JavaScriptCompiler, { Template } from "./javascript-compiler";
 import { Stack } from "@glimmer/util";
-import { assert, expect } from "@glimmer/util";
+import { assert, expect, Option } from "@glimmer/util";
 import { AST, isLiteral, SyntaxError } from '@glimmer/syntax';
 import { getAttrNamespace } from './utils';
 import { Opaque } from "@glimmer/interfaces";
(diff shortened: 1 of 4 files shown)


Serabe:
Supersedes #425

Serabe:
Ping

Serabe:
Rebased master. I would appreciate some review before removing tests.

Thanks!

tomdale:
@Serabe I'm wondering why we need to special case the runtime code if we already re-order opcodes in the compiler? Ideally we could handle ordering issues like this at compile time, so we can avoid doing special cases at runtime.

Serabe:
@tomdale fixing it at several layers was a request of previous PR. You can read it [here](https://github.com/glimmerjs/glimmer-vm/pull/425#issuecomment-286614136).

Thanks!

Serabe:
@tomdale ping!

tomdale:
Thank you @serabe!

PR #727, opened 2017-10-27
[CLEANUP] Use correct environment

Serabe:
In `EmberishRootView` `appendTo` method two different `TestEnvironments`
are being used:

1. In `renderLayout`, the view `env` is being passed as `env` option.
2. In `clientBuilder`, the first parameter is the global `env` declared
   down below.


rwjblue:
Thanks @Serabe!

PR #729, opened 2017-11-01
Update simple-dom

krisselden:
I think this highlights some current misalignments.

Namespaces are completely meaningless for serialization.

I can add it to simple-dom as an ignored property but I think currently this highlights some existing issues.

Also, there seems to be a new dom operations and there are several places we directly do things that don't go through the operations.


krisselden:
I've updated simple-dom on the other side to make this easier. closing this PR will open a new one

Issue #728, opened 2017-10-30
Nested partials is not working as expected


kratiahuja:
# Related issues:

https://github.com/emberjs/ember.js/issues/15791

# Steps to reproduce

The parent template context isn't being passed down to partials that are nested. It is only being passed down to the first partial that is being invoked

Example:

```js
// main.hbs
{{#each model as |record index|}}
  {{partial 'my-partials/some-partials'}}
{{/each}}
```

```js
// my-partials/some-partials.hbs
<p>record id: {{record.id}} </p>
{{partial 'my-partials/name'}}
```

```js
// my-partials/name.hbs
<p> record name: {{record.name}}</p>
```

In the `name.hbs` partial, the record is `undefined`. 

# Expected result:

Rendered template should be:
```
record id: 1
record name: foo
```

# Actual result

```
record id: 1
record name: 
```

GavinJoyce:
failing test: https://github.com/glimmerjs/glimmer-vm/pull/730

GavinJoyce:
Fix here: https://github.com/glimmerjs/glimmer-vm/pull/730

PR #730, opened 2017-11-01
Fix nested partials within a `{{#with}}` block (#728)

GavinJoyce:
fixes https://github.com/glimmerjs/glimmer-vm/issues/728

these changes have already landed in the 0.25 branch: https://github.com/glimmerjs/glimmer-vm/pull/732

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/partial.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/partial.ts
@@ -3,7 +3,7 @@ import { TemplateMeta } from '@glimmer/wire-format';
 import { Op } from '@glimmer/vm';
 import { APPEND_OPCODES } from '../../opcodes';
 import { PartialDefinition } from '../../partial';
-import { assert, dict } from "@glimmer/util";
+import { assert, Dict } from "@glimmer/util";
 import { check, expectStackChange } from '@glimmer/debug';
 import { Opaque } from "@glimmer/interfaces";
 import { CheckReference } from './-debug-strip';
@@ -30,11 +30,12 @@ APPEND_OPCODES.add(Op.InvokePartial, (vm, { op1: _meta, op2: _symbols, op3: _eva
     let partialSymbols = symbolTable.symbols;
     let outerScope = vm.scope();
     let partialScope = vm.pushRootScope(partialSymbols.length, false);
+    let evalScope = outerScope.getEvalScope();
     partialScope.bindCallerScope(outerScope.getCallerScope());
-    partialScope.bindEvalScope(outerScope.getEvalScope());
+    partialScope.bindEvalScope(evalScope);
     partialScope.bindSelf(outerScope.getSelf());
 
-    let locals = dict<VersionedPathReference<Opaque>>();
+    let locals = Object.create(outerScope.getPartialMap()) as Dict<VersionedPathReference<Opaque>>;
 
     for (let i = 0; i < evalInfo.length; i++) {
       let slot = evalInfo[i];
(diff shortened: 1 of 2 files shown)


GavinJoyce:
The outer partial has `locals` containing the `name` reference:

screen shot 2017-11-01 at 09 09 13

The inner partial has empty `locals`:

screen shot 2017-11-01 at 09 09 46

GavinJoyce:
If this is good, we'll need to backport a similar fix to `v0.25` (as a number of things have moved around and this diff wont apply directly)

packages/@glimmer/runtime/lib/compiled/opcodes/partial.ts
```diff
@@ -34,7 +34,10 @@ APPEND_OPCODES.add(Op.InvokePartial, (vm, { op1: _meta, op2: _symbols, op3: _eva
     partialScope.bindEvalScope(outerScope.getEvalScope());
     partialScope.bindSelf(outerScope.getSelf());
 
-    let locals = dict<VersionedPathReference<Opaque>>();
+    let locals = outerScope.getPartialMap();
```
rwjblue: I’m unfamiliar with these internals, we should confirm that this allows arbitrary nesting. Mind adding a test with 4 or 5 levels of partial hell (with each level adding more locals)?

GavinJoyce: As you suspected, there are issues with nested partials within nested `{{with}}` blocks. Looking into these now

GavinJoyce: I added a second test which covers deeply nested partials and with blocks. Interestingly, it fails in glimmer-vm master (with a runtime exception) in a different way to ember 2.16.1 (which doesn't render correctly), I suspect a new regression was introduced in glimmer-vm.

Given this, I suggest that I work on getting both tests to pass in glimmer-vm 0.25. Once I've done that, we can ship a new version of ember with the fix. I can then concentrate on fixing both issues in glimmer-vm master.

GavinJoyce: Here's the glimmer-vm failing test without my changes in this PR (proving that it's preexisting):

https://github.com/glimmerjs/glimmer-vm/pull/731

rwjblue:
> Given this, I suggest that I work on getting both tests to pass in glimmer-vm 0.25. Once I've done that, we can ship a new version of ember with the fix. I can then concentrate on fixing both issues in glimmer-vm master.
  

Sounds good to me. I totally agree that the priority here should be on getting Ember's glimmer-vm version stabilized (as this is biting Ember users in 2.15+). This bug is not affecting glimmer.js users (using `{{partial` in glimmer.js apps throws a helpful error telling the user that partials are not supported).

GavinJoyce:
TODO:

 [ ] mirror these changes: https://github.com/glimmerjs/glimmer-vm/pull/734

GavinJoyce:
@rwjblue I've mirrored @chancancode's changes: https://github.com/glimmerjs/glimmer-vm/pull/734

This is ready for review

packages/@glimmer/runtime/lib/compiled/opcodes/partial.ts
```diff
@@ -30,11 +28,12 @@ APPEND_OPCODES.add(Op.InvokePartial, (vm, { op1: _meta, op2: _symbols, op3: _eva
...
     let outerScope = vm.scope();
     let partialScope = vm.pushRootScope(partialSymbols.length, false);
+    let evalScope = outerScope.getEvalScope();
     partialScope.bindCallerScope(outerScope.getCallerScope());
-    partialScope.bindEvalScope(outerScope.getEvalScope());
+    partialScope.bindEvalScope(evalScope);
     partialScope.bindSelf(outerScope.getSelf());
 
-    let locals = dict<VersionedPathReference<Opaque>>();
+    let locals = Object.create(outerScope.getPartialMap());
```
chancancode: Actually, I just realized this would turn make `locals` `any` (can you confirm?). We should probably write `as dict<VersionedPathReference<Opaque>>` here (or do an assign like you did, etc, anything works as long as it has the right type).

GavinJoyce: updated to 

`let locals = Object.create(outerScope.getPartialMap()) as Dict<VersionedPathReference<Opaque>>;`

chancancode:
LGTM if tests pass

PR #733, opened 2017-11-01
Fix Bounds and ElementBuilder interfaces.

krisselden:


packages/@glimmer/interfaces/lib/dom/simple.d.ts
```diff
@@ -9,7 +9,7 @@ export type Namespace =
   | "http://www.w3.org/XML/1998/namespace"
   | "http://www.w3.org/2000/xmlns/";
 
-export enum NodeType {
+export const enum NodeType {
```
rwjblue: For those wondering what the heck this did (like I wondered), apparently it instructs TypeScript to compile the enum out and replace actual usage with the resulting numbers.  Without `const enum` treatment the usages would result in `Enum.Element` (where `Enum` is an object with various properties that equal the numeric values).

TIL 😄

PR #734, opened 2017-11-01
Revert some unneeded changes to `PartialInvoker`

chancancode:
The original fix is more or less correct, however, we shouldn't have to change anything related to `EvalScope`, because that dictionary is representing what was "passed in" to the component (named args and blocks). Since a partial cannot introduce additional arguments that part of the code is not a problem.

However, there is a separate bug that was caught by the new test. The "top-level" template rendering helper does not bind the `EvalScope` even when `hasPartial` is true. (This is because a "top-level" template cannot  have args and blocks passed in.) However, the `PartialInvoker` assumes it is always present (`!`). So a seconary fix is needed there to correct for the problem.

This is the final diff (against the parent of the fix) to help with porting this to master:

```diff
diff --git a/packages/@glimmer/runtime/lib/syntax/functions.ts b/packages/@glimmer/runtime/lib/syntax/functions.ts
index ad4368f..a662727 100644
--- a/packages/@glimmer/runtime/lib/syntax/functions.ts
+++ b/packages/@glimmer/runtime/lib/syntax/functions.ts
@@ -259,14 +259,15 @@ export class PartialInvoker implements DynamicInvoker<ProgramSymbolTable> {
     let partial = unwrap(_partial);
     let partialSymbols = partial.symbolTable.symbols;
     let outerScope = vm.scope();
+    let evalScope = outerScope.getEvalScope();
     let partialScope = vm.pushRootScope(partialSymbols.length, false);
     partialScope.bindCallerScope(outerScope.getCallerScope());
-    partialScope.bindEvalScope(outerScope.getEvalScope());
+    partialScope.bindEvalScope(evalScope);
     partialScope.bindSelf(outerScope.getSelf());

     let { evalInfo, outerSymbols } = this;

-    let locals = dict<VersionedPathReference<Opaque>>();
+    let locals = Object.create(outerScope.getPartialMap());

     for (let i = 0; i < evalInfo.length; i++) {
       let slot = evalInfo[i];
@@ -275,14 +276,14 @@ export class PartialInvoker implements DynamicInvoker<ProgramSymbolTable> {
       locals[name] = ref;
     }

-    let evalScope = outerScope.getEvalScope()!;
+    if (evalScope) {
+      for (let i = 0; i < partialSymbols.length; i++) {
+        let name = partialSymbols[i];
+        let symbol = i + 1;
+        let value = evalScope[name];

-    for (let i = 0; i < partialSymbols.length; i++) {
-      let name = partialSymbols[i];
-      let symbol = i + 1;
-      let value = evalScope[name];
-
-      if (value !== undefined) partialScope.bind(symbol, value);
+        if (value !== undefined) partialScope.bind(symbol, value);
+      }
     }
```

cc @rwjblue @GavinJoyce

### DIFF
--- a/packages/@glimmer/runtime/lib/syntax/functions.ts
+++ b/packages/@glimmer/runtime/lib/syntax/functions.ts
@@ -5,7 +5,6 @@ import {
 } from '@glimmer/reference';
 import {
   assert,
-  assign,
   Dict,
   dict,
   EMPTY_ARRAY,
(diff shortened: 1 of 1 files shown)


GavinJoyce:
nice one, much neater (and I now have a little more context than before)

PR #735, opened 2017-11-02
Allow passing additional template metadata to BundleCompiler

tomdale:
1. The term "specifier" was used inconsistently in the codebase and was a source of intense confusion at the integration layer in Glimmer.js, where we had to round-trip between Module Unification specifiers and bundle compilers "specifiers." This PR changes the term in the context of the bundle compiler to "module locator."
2. The bundle compiler was hardcoding module semantics, which makes sense during the compilation phase, but is often the incorrect metadata needed to most efficiently resolve lookups at runtime. For example, in the Module Unification world, the most efficient way to resolve dynamic component lookups is with absolute specifier strings.
3. This PR allows users of the bundle compiler to provide additional metadata when adding a template, in addition to the module locator information.
4. There was generally some inconsistent use of terminology throughout the codebase that I've tried to clean up here.
5. In a future PR, I intend to include *only* the supplied template metadata in bytecode-compiled programs, and only for templates that do dynamic lookup, i.e. `{{partial}}` and `{{component}}`. This will significantly reduce the file size of the compiled JSON "data segment" for binary bytecode apps.


chadhietala:
This is a really nice cleanup

PR #737, opened 2017-11-06
Add tests for basic #each use cases

chadhietala:
We should slowly be moving syntax tests from [Ember-Glimmer](https://github.com/emberjs/ember.js/tree/master/packages/ember-glimmer/tests/integration/syntax) into the VM. Libraries/frameworks should not be testing builtin semantics.


rwjblue:
I totally agree that we should be testing these things here, but I do not think that Ember should remove the test coverage.  Ember has custom iterators, custom rules around things like proxies, truthiness, etc. 

It costs us very little to keep the coverage in Ember, and tests in this repo have generally not been super well maintained over the many many iterations that have happened since the HTMLBars days...

PR #740, opened 2017-11-09
Mark @glimmer/test-helpers as private

chadhietala:
This used to be used by @glimmer/application but that is no longer the case. This is also in conflict of https://github.com/glimmerjs/glimmer-test-helpers


rwjblue:
Not sure how we lost the work, but this was already done in https://github.com/glimmerjs/glimmer-vm/pull/563.

PR #741, opened 2017-11-11
Fix bytecode recursive component invocation

tomdale:
Currently, recursive invocation of components is failing when the program is compiled as bytecode. The compiler will crash due to a stack overflow because it tries to recursively assign handles to the still-compiling `CompilableTemplate`.

The compilable template's `compile()` method, which generates a unique handle for each component, guards on the existence of the handle to know whether compilation has already happened. Unfortunately, recursive operations like this cause a bit of a Catch 22 situation. We need the opcode and operands to know how much space to malloc on the heap (and thus assign a VM handle), but we don't know what VM handle to emit as the operand because the malloc hasn't happened yet.

This PR introduces the ability to push placeholder values on to the heap, in addition to normal primitive numbers. Placeholder values increment the current offset like normal, but record the position of the placeholder value and additionally store a value-producing function.

When compilation of the program finishes (as signalled by a call to `capture()`), a final pass (or "linking" step, if you will) occurs, where placeholder values are replaced by the real value, as produced by invoking the originally-passed function.

### DIFF
--- a/packages/@glimmer/encoder/lib/encoder.ts
+++ b/packages/@glimmer/encoder/lib/encoder.ts
@@ -6,11 +6,13 @@ export const TYPE_SIZE        = 0b11111111;
 export const TYPE_MASK        = 0b0000000011111111;
 export const OPERAND_LEN_MASK = 0b0000001100000000;
 
+export type Operand = number | (() => number);
+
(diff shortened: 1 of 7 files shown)


chadhietala:
This is great 👍.

PR #743, opened 2017-11-20
[BREAKING] Move lazy mode constructs to @glimmer/opcode-compiler

chadhietala:
This moves the location of several constructs that are only needed for lazy-mode compilation into `@glimmer/opcode-compiler`. This is where all the code is for running the an application in lazy mode. This does not change any functionality but rather when integrating the code has moved.

/cc @krisselden @smfoote


tomdale:
Can you outline the mental model we should have about what goes where in terms of:

1. `@glimmer/opcode-compiler`
2. `@glimmer/compiler`
3. `@glimmer/bundle-compiler`
4. `@glimmer/runtime`
5. `@glimmer/interfaces`

It seems like that, at a high level, there are "compiler" packages and "runtime" packages, in the sense that a compiler package produces an artifact that is consumed by a runtime package. Exactly which environment the package runs in depends on the mode. We also have "interface" packages that describe common shared data structures between compiler and runtime packages. Wondering if it makes sense to rename/reorganize some of these packages given the changes over the past few months to make these lines clearer. (It's really too bad we can't just rely on treeshaking to produce optimal builds.)

wycats:
@tomdale 


* @glimmer/opcode-compiler - compiles the IR (formerly known as the "wire format") into the executable program
* @glimmer/compiler - compiles the syntax into the IR (formerly known as the "wire format")
* @glimmer/bundle-compiler - AOT compiles the syntax into the executable program, using the compiler and opcode-compiler
* @glimmer/runtime - is used to execute the compiled program
* @glimmer/interfaces - pure interfaces used by the rest of the packages (originally created to break cycles)

The division of responsibilities is mostly derived from this distinction:

| Name | Step | In "classic mode" | In "AOT mode" |
| --- | --- | --- | --- |
| compiler | compile the syntax into the IR | AOT | AOT |
| opcode-compiler | compile the IR into opcodes | at runtime | AOT |
| runtime | execute the program | at runtime | at runtime |

The `bundle-compiler` simple combines the `compiler` and `opcode-compiler` into one ergonomic package when fully AOTing.

wycats:
@Turbo87 what aspect of this PR is breaking?

Turbo87:
I don't know but it said breaking in the title ;)

wycats:
@chadhietala ?

krisselden:
the breaking was only where you import stuff from.

krisselden:
#777 includes this commit

PR #748, opened 2017-11-28
Refine Dynamic Component Invocations

chadhietala:
This fixes an issue where we were eagerly serializing `TemplateMeta` for the append of a closed over component. This is because we had an overly generic `PushDynamicComponentManager` that was being used to resolve and/or pushing the definition for the actual invoke. This PR introduces 2 new opcodes: one to deal with resolving the definition and one for simply pushing a pre-resolved curried component.


wycats:
This is great. Thanks @chadhietala!

PR #752, opened 2017-12-01
Initial stab at porting `asm/stack.ts` to Rust

alexcrichton:
This commit is an initial stab at adding some Rust to glimmer, notably to
implement some bits-and-pieces of the `low-level` package for now. Specifically
this commit takes the `Stack` type in the `low-level` module and ports it to
Rust and then integrates it all back with WebAssembly.

Some notable points of this:

* The error handling isn't great right now and the interface may not be quite
  right just yet.
* Right now the raw wasm instructions are used to allocate memory (the Rust
  standard library's allocator is avoided) and internally a freelist is
  maintained of free'd pages.
* Currently memory is never deallocated in Rust which means that pages are
  forever in use and usage will "grow without bound". Eventually a manual call
  to "free this" will need to be inserted for the `Stack` type.


packages/@glimmer/low-level/rust/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+#![feature(link_llvm_intrinsics)]
+
+#[no_mangle]
+pub extern fn add(a: u32, b: u32) -> u32 {
```
steveklabnik: was this intended to be left in?

wycats: 🤔

alexcrichton: Oops yeah that was just leftover

alexcrichton:
The three currently failing tests I believe are

* [`SL_IE_11`](https://saucelabs.com/jobs/6c3b4adafacf4a4da89b58300c8134ad) - I think that IE doesn't have `WebAssembly` support at all?
* [`SL_Safari_Last`](https://saucelabs.com/jobs/87bc122560cf48bd8e0a5ff52c40fa09) - I think this is using Safari 10 which according to [this](https://developer.mozilla.org/en-US/docs/WebAssembly) support was added in Safari 11.
* [`SL_MS_Edge`](https://saucelabs.com/jobs/2c403cd897e04fb2a0677cbf5f6350b2) - This looks to be Edge 15 but according to [this](https://developer.mozilla.org/en-US/docs/WebAssembly) the webassembly support is behind a flag by default? I tested with what I think is Edge 16 locally and it worked by out-of-the-box.

alexcrichton:
@wycats ok I pushed up the extra commits we talked about here! I think I addressed most of the immediate feedback and left a few questions, but let me know if there's anything else that should be resolved!

chicoxyzzy:
@alexcrichton just in case, I can assure you that this table of wasm support by browsers is absolutely accurate https://caniuse.com/#feat=wasm

wycats:
@chicoxyzzy confirm!

chriskrycho:
@wycats @alexcrichton @tomdale just want to flag up that I'm *very* eager to jump in and help on this push after EmberConf if it's still in the open state. Need to finish getting the story with TypeScript in Ember more solid and get through the workshop I'm doing there, but this is a thing I'd *love* to help drive forward, given my broader ecosystem interests.

lifeart:
is it alive?

cbou:
Any news here?

wycats:
It's somewhat alive, but we've been focusing our efforts on Ember integration lately. @chriskrycho is taking a stab at porting the compiler to Rust/wasm with some very promising results so far!

kripken:
Was this completed, or abandoned? Either way it would be interesting to hear some details of the experience here!

Becavalier:
Is there any update regarding the VM migration to Wasm?

lifeart:
@Becavalier https://discuss.emberjs.com/t/what-is-the-current-state-of-more-advanced-glimmer-vm-features/18114

PR #757, opened 2017-12-15
Fix debugSlice for lazy compile

krisselden:


chadhietala:
🤕

chancancode:
does `program.opcode` mutate? 😐  @chadhietala would it be possible to add a non-mutating version of it for this purpose?

PR #760, opened 2017-12-30
Don't merge: upgrade simple-dom

chancancode:
The actual implementation of `simple-dom` (now fully typed) is not compatible with `Simple.*`. Specifically, it does not support namespaces. Removing them from `Simple.*` yields around type 20 errors:

render-test_ts_ _glimmer

Most of these are in tests. However, I do think there is a real issue. Specifically, in SSR mode, we are wrongly assuming we can call NS methods like `createElementNS`. I don't know if this causes problems in practice – maybe a combination of things causes these things to never be called in practice, but we should get the types to more accurately reflect reality.

Also, since `simple-dom` is now fully typed, and is a strict subset of DOM, do we still need `Simple.*`?


tomdale:
I think this is superseded by #866.

PR #762, opened 2018-01-05
Allow null for iterable keys

smfoote:
Ember does not require a key for the `{{#each}}` helper, so allow null
and do the check in the environment's `iterableFor`.

### DIFF
--- a/packages/@glimmer/test-helpers/lib/environment/environment.ts
+++ b/packages/@glimmer/test-helpers/lib/environment/environment.ts
@@ -1,6 +1,5 @@
 import { Environment, DOMTreeConstruction, IDOMChanges, PrimitiveReference, ConditionalReference } from '@glimmer/runtime';
 import { dict } from '@glimmer/util';
-import { check, CheckString } from '@glimmer/debug';
 import { Dict, RuntimeResolver, Opaque, VMHandle } from '@glimmer/interfaces';
 import { Program } from '@glimmer/program';
 import { Reference, isConst, OpaqueIterable } from '@glimmer/reference';
@@ -39,7 +38,6 @@ export default abstract class TestEnvironment<TemplateMeta> extends Environment 
 
   iterableFor(ref: Reference<Opaque>, keyPath: string): OpaqueIterable {
     let keyFor: KeyFor<Opaque>;
-    check(keyPath, CheckString);
 
     if (!keyPath) {
       throw new Error('Must specify a key for #each');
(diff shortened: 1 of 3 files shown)


packages/@glimmer/test-helpers/lib/environment/environment.ts
```diff
@@ -38,6 +39,7 @@ export default abstract class TestEnvironment<TemplateMeta> extends Environment
 
   iterableFor(ref: Reference<Opaque>, keyPath: string): OpaqueIterable {
     let keyFor: KeyFor<Opaque>;
+    check(keyPath, CheckString);
```
smfoote: I'm not sure how helpful this `check` is, seeing it's in test-helpers and it comes right after `keyPath` has its type declared as `string`.

tomdale:
For some context on this PR:

Glimmer's `{{#each}}` helper takes a "key" to make array diffing more efficient. For example, if you have two arrays:

```js
let arr1 = [{ id: 1 }, { id: 2 }];
let arr2 = [{ id: 1 }, { id: 2 }];
```

A naive implementation of `{{#each}}` that only compares object identity would cause a full re-render if you change from `arr1` to `arr2`, even though each object in `arr2` is the same as in `arr1`, by virtue of containing the same properties with the same values.

To make the above example work, Glimmer.js forces you to provide a key: `{{#each items key="id" as |item|}}`. If you want to opt-in to the more-likely-to-be-slow path of just comparing object identity, you can do that explicitly by setting `{{#each items key="@index"}}`.

However, Ember's API *does not* require to provide an explicit key. If you elide the key, it will fall back to implicit `@index` behavior.

The host `Environment` has a hook called `iterableFor` that creates the iterable wrapper around the underlying array. Because iteration behavior is environment-specific, this is the right hook to do correctness checks, e.g. whether the key was provided or not.

This PR removes an assertion in the lower levels of the VM that throws if no key is present. This is the correct behavior for Glimmer.js but not for Ember. Ironically, this was still working before because Ember was only using production builds of Glimmer VM, which have assertions stripped. Now that we are using debug builds of Glimmer VM in development, the mismatch in behavior became suddenly obvious. 😉 

*TL;DR* Because iteration behavior is environment-specific, it makes sense to do runtime correctness checks in the environment hooks, not the VM. The VM should only make assertions that hold across all possible environments such as Ember and Glimmer.js.

packages/@glimmer/test-helpers/lib/environment/environment.ts
tomdale: I think you can safely remove this, particularly given we have the truthiness check on line 44.

PR #764, opened 2018-01-08
Improve support for XHTML

cbou:
I had to open a new PR because I deleted the old branch. The original PR was https://github.com/glimmerjs/glimmer-vm/pull/591.

@tomdal What did you meant with "Include a test for the normalization case.". Should I write a test for the function `normalizeTagName`?


tomdale:
@cbou I just meant a test that would fail if your change was removed—I can see us accidentally regressing on this because it's somewhat uncommon. I guess it might be tricky to write a test given the test harness HTML document is not XHTML itself. I just want to encode somewhere in the tests what the intended behavior is beyond a comment; otherwise someone will probably delete it on accident.

cbou:
Ok, that was what I understood. Unfortunately I'm not very motivated to invest time in this issue because we switch from XHTML to HTML...

Turbo87:
closing due to inactivity

PR #765, opened 2018-01-09
The modules build and types are needed for yarn link.

krisselden:


rwjblue:
I walked @wycats through doing this locally to link with Ember, thank you for fixing it!

krisselden:
If we add a TURBO mode for normal dev then it should be a flag you pass to skip this, I think by default the modules should be output

wycats:
:+1: 🕺

PR #767, opened 2018-01-12
Fix nested currying.

rwjblue:
Given the following template:

```hbs
{{component (component (component 'foo-bar' 'one') 'two')}}
```

We expect that `foo-bar` receives two positional params: `'one'` and `'two'`.

Prior to the changes made here the _actual_ result was `'one'` and `'one'`. :scream:

The underlying issue here is that the `OpcodeBuilder#curryComponent` building assumes that when it calls `this.pushArgs` _nothing_ else will mutate those args before the `Ops.CurryComponent` is ran. Unfortunately, when we have nested invocations, the processing of the nested `(component` invocation ends up modifying the global shared `ARGS` so that when the outer `(component`'s `Op.CurryComponent` is ran, the contents of `ARGS` is _wrong_.

The fix (as implemented here) is that we push a new `CaptureArgs` opcode so that any arbitrary levels of nested `(component` invocations all end up with the correct set of captured args.

---

A joint @tomdale / @rwjblue production...

### DIFF
--- a/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
+++ b/packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
@@ -1039,6 +1039,7 @@ export abstract class OpcodeBuilder<Locator> extends SimpleOpcodeBuilder {
 
     this.pushFrame();
     this.compileArgs(params, hash, null, synthetic);
+    this.push(Op.CaptureArgs);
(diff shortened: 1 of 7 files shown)


tomdale:
@rwjblue I added a short description in the opcode comment

rwjblue:
Awesome, thank you @tomdale!

PR #768, opened 2018-01-18
Start fleshing out the precompiler docs

smfoote:


wycats:
@smfoote what do you think about including some code snippets illustrating which APIs are in use?

smfoote:
@wycats  I think that makes sense. Are you thinking the APIs used during precompilation (like `preprocess`, `TemplateCompiler.compile`, etc.) or the APIs that _call_ preprocess to get the precompilation started? Or both?

wycats:
@smfoote Mostly I was thinking about the top-level APIs, but anything you think would help clarify the setup 😄

wycats:
@smfoote Thanks! Keep 'em coming!

Issue #755, opened 2017-12-14
Unhelpful error message from unrelated errors


chriskrycho:
When an unhandled error gets thrown by the application, it can end up showing you only an error like this as the final result (from an example I just ran into in our Ember.js application):

```
Error: a glimmer transaction was begun, but one already exists. You may have a nested transaction
debugAssert — util.js:62
begin — runtime.js:6380
begin — environment.js:270
_renderRoots — renderer.js:311
_renderRootsTransaction — renderer.js:379
_revalidate — renderer.js:418
invokeWithOnError — backburner.js:283
flush — backburner.js:153
flush — backburner.js:345
end — backburner.js:455
run — backburner.js:533
_trigger — backburner.js:882
end — backburner.js:468
run — backburner.js:539
wrapped — raygun.js:49
mightThrow — jquery.js:3583
(anonymous function) — jquery.js:3651
(anonymous function) — error.js:155
runInDebug — index.js:246
onerror — error.js:148
invokeWithOnError — backburner.js:286
flush — backburner.js:153
flush — backburner.js:345
end — backburner.js:455
run — backburner.js:533
_trigger — backburner.js:882
end — backburner.js:468
run — backburner.js:539
wrapped — raygun.js:49
mightThrow — jquery.js:3583
(anonymous function) — jquery.js:3651
```

This is thrown here:

https://github.com/glimmerjs/glimmer-vm/blob/27393d1efdc18268e0b67d15ccceeb688b24957a/packages/%40glimmer/runtime/lib/environment.ts#L258

You can debug it the normal ways, and figure out what the original error was, but the error message itself should probably be clearer about indicating that it may result from other errors.

rwjblue:
IMHO, we should (at the least) update the message to tell folks to look at the errors previously thrown.

GavinJoyce:
How about:

`A glimmer transaction was begun, but one already exists. You may have a nested transaction, possibly caused by an earlier runtime exception. Please check your stack trace.`

chriskrycho:
It’s certainly an improvement on what we have now.

That said, with the normal stack trace settings, I can guarantee that won’t actually help: I’ve never seen one of these that included a large enough stack trace to get back to what the actual problem was. (I think the usual stack depth for traces in production is 10?)

PR #771, opened 2018-01-25
improve nested transaction error message

GavinJoyce:
fixes https://github.com/glimmerjs/glimmer-vm/issues/755

**before**

`a glimmer transaction was begun, but one already exists. Please check your console for the stack trace of any prior exceptions.`

**after**

`A glimmer transaction was begun, but one already exists. You may have a nested transaction, possibly caused by an earlier runtime exception while rendering. Please check your console for the stack trace of any prior exceptions.`

### DIFF
--- a/packages/@glimmer/runtime/lib/environment.ts
+++ b/packages/@glimmer/runtime/lib/environment.ts
@@ -255,7 +255,7 @@ export abstract class Environment {
   }
 
   begin() {
-    assert(!this._transaction, 'a glimmer transaction was begun, but one already exists. You may have a nested transaction');
+    assert(!this._transaction, 'A glimmer transaction was begun, but one already exists. You may have a nested transaction, possibly caused by an earlier runtime exception while rendering. Please check your console for the stack trace of any prior exceptions.');
     this._transaction = new Transaction();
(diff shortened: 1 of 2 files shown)


packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -255,7 +255,7 @@ export abstract class Environment {
   }
 
   begin() {
-    assert(!this._transaction, 'a glimmer transaction was begun, but one already exists. You may have a nested transaction');
+    assert(!this._transaction, 'A glimmer transaction was begun, but one already exists. You may have a nested transaction, possibly caused by an earlier runtime exception while rendering. Please check your stack trace.');
```
rwjblue: What do you think about this small tweak to the last sentance:

> Please check your console for the stack trace of any prior exceptions.

GavinJoyce: that's better, thanks. I'll update

GavinJoyce: ^ updated

rwjblue:
Thank you!

PR #773, opened 2018-01-31
Encode component capabilities using bitmasks

tomdale:
Glimmer has the concept of "capabilities," which are per-component features that can be turned on or off and allow the VM to perform both compile-time and run-time optimizations.

Previously, we represented capabilities as a JavaScript object, where each capability was a property whose value was either true or false. This works, but introduces difficulties when interoperating with WebAssembly, which doesn't understand JavaScript objects natively.

This commit introduces a serialization of the `ComponentCapabilities` object into a 32-bit integer, where each capability is represented as a single bit. Note that this change is purely internal to the VM at the moment; component managers still return a JavaScript object from `getComponentCapabilities`.

As an additional optimization, this commit also caches a component's capabilities on the `ComponentInstance` object. Although capabilities were intended to be immutable over the lifetime of a component, we never cached capabilities and opcodes retrieved them from the component manager as needed.

### DIFF
--- a/packages/@glimmer/runtime/lib/capabilities.ts
+++ b/packages/@glimmer/runtime/lib/capabilities.ts
@@ -1,0 +1,31 @@
+import { ComponentCapabilities, Unique, Recast } from "@glimmer/interfaces";
+import { check, CheckNumber } from "@glimmer/debug";
+
+export type CapabilityFlags = Unique<"CapabilityFlag">;
+
+export const enum Capability {
+  DynamicLayout = 0b000001,
+  DynamicTag    = 0b000010,
+  PrepareArgs   = 0b000100,
+  CreateArgs    = 0b001000,
+  AttributeHook = 0b010000,
+  ElementHook   = 0b100000
+}
+
+/**
+ * Converts a ComponentCapabilities object into a 32-bit integer representation.
+ */
+export function capabilityFlagsFrom(capabilities: ComponentCapabilities): CapabilityFlags {
+  return (0 |
+    (capabilities.dynamicLayout ? Capability.DynamicLayout : 0) |
+    (capabilities.dynamicTag ? Capability.DynamicTag : 0) |
+    (capabilities.prepareArgs ? Capability.PrepareArgs : 0) |
+    (capabilities.createArgs ? Capability.CreateArgs : 0) |
+    (capabilities.attributeHook ? Capability.AttributeHook : 0) |
+    (capabilities.elementHook ? Capability.ElementHook : 0)) as Recast<number, CapabilityFlags>;
+}
+
+export function hasCapability(capabilities: CapabilityFlags, capability: Capability): boolean {
+  check(capabilities, CheckNumber);
+  return !!((capabilities as Recast<CapabilityFlags, number>) & capability);
+}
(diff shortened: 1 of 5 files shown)


packages/@glimmer/runtime/lib/capabilities.ts
```diff
@@ -0,0 +1,31 @@
+import { ComponentCapabilities, Unique, Recast } from "@glimmer/interfaces";
+import { check, CheckNumber } from "@glimmer/debug";
+
+export type CapabilityFlags = Unique<"CapabilityFlag">;
+
+export const enum Capability {
```
chadhietala: This should likely be:

```
0b000001
0b000010
0b000011
0b000100
0b000101
0b000111
```

chadhietala: Or just a single 8 bit int

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -258,16 +279,31 @@ APPEND_OPCODES.add(Op.PrepareArgs, (vm, { op1: _state }) => {
   stack.push(args);
 });
 
+function resolveCurriedComponentDefinition(instance: ComponentInstance, definition: CurriedComponentDefinition, args: Arguments): ComponentDefinition {
+  let unwrappedDefinition = instance.definition = definition.unwrap(args);
+  let { manager, state } = unwrappedDefinition;
+
+  assert(instance.manager === null, "component instance manager should not be populated yet");
```
chadhietala: Can't you just use `expect`

packages/@glimmer/runtime/lib/capabilities.ts
tomdale: Not sure I follow… can you say why?

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
tomdale: Doesn't `expect` assert that something is *not* null?

packages/@glimmer/runtime/lib/capabilities.ts
chadhietala: NM

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
wycats: An `assert` is right here.

packages/@glimmer/runtime/lib/capabilities.ts
```diff
@@ -0,0 +1,31 @@
...
+  ElementHook   = 0b100000
+}
+
+/**
+ * Converts a ComponentCapabilities object into a 32-bit integer representation.
+ */
+export function capabilityFlagsFrom(capabilities: ComponentCapabilities): CapabilityFlags {
+  return (0 |
+    (capabilities.dynamicLayout ? Capability.DynamicLayout : 0) |
+    (capabilities.dynamicTag ? Capability.DynamicTag : 0) |
```
wycats: You could use `capabilities.dynamicLayout | 0` to cast a boolean into `0` or `1`.

tomdale: TypeScript forbids bitwise operators on anything but numbers, enums or `any`.

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
tomdale: @wycats Huh?

wycats: I'm agreeing with you.

packages/@glimmer/runtime/lib/capabilities.ts
wycats: Yeah. Too bad. I really wish there was some way to disable this particular error in "boundary modules" like this but oh well.

PR #777, opened 2018-02-07
[BREAKING] Cleanup

krisselden:
This is some cleanup in preparation for rebasing #743 

I have more cleanup coming

Doesn't change how anything works but shifts where you import somethings.


chadhietala:
Is this done? I see some TODOs about stuff not working.

packages/@glimmer/bundle-compiler/lib/bundle-compiler.ts
```diff
@@ -84,6 +85,9 @@ export interface PartialTemplateLocator<TemplateMeta> extends ModuleLocator {
   kind?: 'template';
 }
 
+// to make --declaration happy
```
tomdale: Is this a TypeScript bug?

krisselden: not sure, it felt like it. It only complained in the class initializer, when the alias was in methods it was ok

Issue #782, opened 2018-02-13
Inline `Concat` nodes with single element


Turbo87:
Related: https://github.com/rwjblue/ember-template-lint/issues/350

![bildschirmfoto 2018-02-13 um 13 34 24](https://user-images.githubusercontent.com/141300/36150109-d6ceb682-10c2-11e8-8e42-906621736038.png)

In cases where a `Concat` instruction has a `size` of `1` we should probably be able to inline the content and skip serializing the `Concat` instruction itself. 🤔 

/cc @rwjblue

Turbo87:
closing due to https://github.com/glimmerjs/glimmer-vm/pull/785#issuecomment-366726001

PR #785, opened 2018-02-15
WIP: Drop `ConcatStatement` nodes with only a single part

Turbo87:
This PR is my attempt at implementing https://github.com/glimmerjs/glimmer-vm/issues/782, but unfortunately several tests are now failing and I'm not quite sure why 🤔 

Any guidance would be appreciated :)

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -372,9 +372,14 @@ export default class TemplateCompiler {
         this.attributeMustache([value]);
         return false;
       case 'ConcatStatement':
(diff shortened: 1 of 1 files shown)


tomdale:
To summarize the conversation in Slack: this approach won't work because the `Concat` opcode also serves to coerce non-string values into strings. Theoretically we could still perform this optimization in some cases, such as if we know the reference on the stack is a string literal value.

Turbo87:
I'll close this for now due to the reasons mentioned above

wycats:
@tomdale we could also potentially create a `ToString` opcode that handled this case without the whole Concat setup. It might not be useful or an improvement though,

PR #786, opened 2018-02-23
[BUGFIX] Support Number.MAX_SAFE_INTEGER

chadhietala:
Fixes https://github.com/emberjs/ember.js/issues/16269.

Prior to this fix we would hard error if operands exceeded 16-bits. This was due to the fact that the program is 16-bit aligned. However, we don't have to take this nuclear option. Instead we can just serialize these larger numbers into the constants pool and reify them properly at runtime. This means you are free to use number literals up to `Number.MAX_SAFE_INTEGER`.


rwjblue:
We may want to iterate further on this, but the current implementation is good enough to unblock the reported bugs in Ember 3.1’s beta cycle so I’m going to go ahead and land + release...

Issue #787, opened 2018-02-24
Export isRehydrationComment for rehydration/serialization concerns


rondale-sc:
While working on the rehydration story, outlined here: https://github.com/emberjs/ember.js/pull/16227, I've learned that it is important to be able to recognize when a given string has been created using the serialize element builder.  In Glimmer there is an assertion that ensures that a node must contain a known glimmer comment string which looks like: `'%+b:0%'` which I've used to make decisions in ember-cli-fastboot to determine how the application should behave.

In the Ember.js PR I rely on this comment in tests, and in ember-cli-fastboot I rely on it to change the behavior of the instance-initializer and to modify the boot process to ensure rehydrate mode is set correctly.  

Unfortunately, that means we have that magic string encoded all over many libraries.  I'd like to expose a `isRehydrationComment` function that can be used internally in glimmer-vm, but also in these other libraries so we don't have to hardcode the string and the underlying format can change as needed.

rwjblue:
Totally agreed here. 

Perhaps it could even be `isRehydrationNode` (which even hides the fact that we use comments)...

PR #788, opened 2018-02-27
Expose isSerializationFirstNode and SERIALIZATION_FIRST_NODE_STRING

rondale-sc:
This exposes a mechanicsm that can be used in all the places in ember.js
/ fastboot and anywhere it might be necessary to determine whether or
not a given node is the first serialization node.  Which is useful to
ensure that the actual format can change without affecting the other
libaries who will be able to use this instead of a magic string.

The rationale for this PR can be found in greater detail at the issue
below:
https://github.com/glimmerjs/glimmer-vm/issues/787


rondale-sc:
I wanted to expose a function that could be used to better insulate future changes rather than only export a constant.

rondale-sc:
This is immediately relevant to the following PRs

https://github.com/ember-fastboot/ember-cli-fastboot/pull/580

and

https://github.com/ember-fastboot/fastboot/pull/185

PR #791, opened 2018-03-04
Fix bugs caused by html-safe empty strings

mongoose700:
Adding empty strings as HTML leads to difficulties keeping track of
sibling nodes. It's easier to keep track of as text.

This fixes the issues https://github.com/emberjs/ember.js/issues/14924 and https://github.com/emberjs/ember.js/issues/16314. The first one could be solved by simply removing the `if (html === null || html === '')` special case. The latter issue cannot, because it would still need to deal with a `null` last node.

### DIFF
--- a/packages/@glimmer/runtime/lib/dom/helper.ts
+++ b/packages/@glimmer/runtime/lib/dom/helper.ts
@@ -161,7 +161,7 @@ export function insertHTMLBefore(this: void, useless: HTMLElement, _parent: Simp
   let last: Simple.Node | null;
 
   if (html === null || html === '') {
-    return new ConcreteBounds(parent, null, null);
+    throw new Error('Empty strings should be added as text, not HTML');
   }
 
   if (nextSibling === null) {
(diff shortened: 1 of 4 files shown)


rwjblue:
Awesome, thank you for digging into this!

mongoose700:
It seems to be failing some IE tests in CI, even though those tests don't even encounter my changes. Are they known to be flakey?

rwjblue:
I’ll restart, I’m not sure if they are normally flakey or not...

rwjblue:
I kicked it once more...

rwjblue:
Latest build had a single failure in IE11:

screenshot

mongoose700:
How do you normally go about debugging issues that only happen in IE?

rwjblue:
The "easiest" thing would be to download an IE11 virtual machine (from [modern.ie](https://modern.ie)) and run the tests...

mongoose700:
I've run the tests on a virtual machine repeatedly, and they never fail locally.

mongoose700:
I've tried with Windows 10, and the test still passes. I switched to QUnit 2.0.0 to match what CI is using, and updated IE, which took me past the version used in CI. Is there a way I can use the exact same VM that CI is using?

[
screen shot 2018-03-13 at 12 52 20 am
](url)

chancancode:
@mongoose700 sorry this took so long! I merged in your changes in #852 and credited you for the commit. Would you mind opening a new PR to add the #each test you have here?

PR #798, opened 2018-04-12
[BUGFIX] Grow the Uint16Array if the page size overflows 1MB

chadhietala:


packages/@glimmer/program/lib/program.ts
```diff
@@ -71,14 +74,29 @@ export class Heap implements CompileTimeHeap {
...
 
   push(item: number): void {
     this.heap[this.offset++] = item;
+    this.sizeCheck();
+  }
+
+  private sizeCheck() {
+    this.pageSize++;
+    if (this.pageSize >= PAGE_SIZE) {
+      debugger;
```
rwjblue: 🔪

packages/@glimmer/program/lib/program.ts
```diff
@@ -71,14 +74,26 @@ export class Heap implements CompileTimeHeap {
       this.table = table;
       this.offset = this.heap.length;
       this.handle = handle;
+      this.pageSize = this.heap.length;
```
pieter-v: As `this.pageSize` keeps track of the used space in the last page of the heap, I think this will only work if `serializedHeap.buffer` is a multiple of `PAGE_SIZE`.
If in stead, we keep track of the free space, then this should work.

packages/@glimmer/program/lib/program.ts
```diff
@@ -71,14 +74,26 @@ export class Heap implements CompileTimeHeap {
...
     } else {
-      this.heap = new Uint16Array(0x100000);
+      this.heap = new Uint16Array(PAGE_SIZE);
       this.table = [];
     }
   }
 
   push(item: number): void {
     this.heap[this.offset++] = item;
+    this.sizeCheck();
```
pieter-v: The size should be checked before an element is put on the heap, otherwise a size check should also be executed in the constructor in case of a `serializedHeap`.
This has also the advantage that we don't need to grow the heap in case we end on a page boundary :smiley:

packages/@glimmer/program/lib/program.ts
```diff
@@ -71,14 +74,26 @@ export class Heap implements CompileTimeHeap {
...
+  }
+
+  private sizeCheck() {
+    this.pageSize++;
+    if (this.pageSize >= PAGE_SIZE) {
+      let heap = slice(this.heap, 0, this.offset);
```
pieter-v: I don't understand why the copy of the heap is necessary, is it not enough to keep a reference to the old heap in a local variable?

packages/@glimmer/program/lib/program.ts
```diff
@@ -181,6 +196,7 @@ export class Heap implements CompileTimeHeap {
 
   pushPlaceholder(valueFunc: () => number): void {
     let address = this.offset++;
+    this.sizeCheck();
```
pieter-v: As above, the size should be checked before an element is put on the heap.

rondale-sc:
Is this fix going to land in a point release of Ember 3.1 as a hotfix?

rwjblue:
Yes

PR #799, opened 2018-04-18
add failing test for accessing form attribute

Dhaulagiri:
Attempting to add a failing test that reproduces the problem I'm seeing in https://github.com/emberjs/ember.js/issues/16385

Any insight into if this is the right direction and _where_ I might look in order to fix this would be much appreciated.  It seems like [this code](https://github.com/Dhaulagiri/glimmer-vm/blob/a9fb9ebe35df3e01dfc9b8672aa5cb1d22f00ecf/packages/%40glimmer/runtime/lib/dom/props.ts#L18-L24) is relevant but I'm still trying to learn the codebase.

Fixes https://github.com/emberjs/ember.js/issues/16385

### DIFF
--- a/packages/@glimmer/runtime/lib/dom/props.ts
+++ b/packages/@glimmer/runtime/lib/dom/props.ts
@@ -65,7 +65,8 @@ const ATTR_OVERRIDES = {
   LABEL:    { form: true },
   FIELDSET: { form: true },
   LEGEND:   { form: true },
-  OBJECT:   { form: true }
+  OBJECT:   { form: true },
+  BUTTON:   { form: true }
 };
 
 function preferAttr(tagName: string, propName: string) {
--- a/packages/@glimmer/runtime/test/attributes-test.ts
+++ b/packages/@glimmer/runtime/test/attributes-test.ts
@@ -159,6 +159,12 @@ export class AttributesTests extends RenderTest {
     this.assertStableRerender();
   }
 
+  @test "can read the form attribute"() {
+    this.render('<button form="gnargnar" />');
+    this.assert.equal(this.readDOMAttr('form'), 'gnargnar');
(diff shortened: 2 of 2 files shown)


Dhaulagiri:
Adding `BUTTON` to the list of overrides [here](https://github.com/Dhaulagiri/glimmer-vm/blob/a9fb9ebe35df3e01dfc9b8672aa5cb1d22f00ecf/packages/%40glimmer/runtime/lib/dom/props.ts#L62) makes my test pass, but it'd be good to get confirmation that I am even looking at the right thing.

rwjblue:
That seems good, but we should try to figure out what changed between 0.25.6 (which is what was used in ember-source@3.0.0) and now that would have made it just start failing.  Perhaps a starting point would be to send another PR with this exact test but against the release-0-25 branch?

Dhaulagiri:
Digging in more, it looks like the changes [here](https://github.com/Dhaulagiri/glimmer-vm/commit/f8170a30fa29ec5bdbf1151c68a3c5297f3b6e81#diff-9622119753418be6ba2e05653dd68876) to remove phantom support also removed the `BUTTON` override.  Putting that code back into props.ts fixes this failing test.

rwjblue:
Yep, I think you nailed it.  It seems (in hindsight) that that commit should only have removed the `type` entry for `button` not both `type` and `form`.  Adding it back seems completely correct (along with this test to make sure we know that we still need it).

Awesome job sleuthing this out!

Dhaulagiri:
@rwjblue thanks for helping me track it down.  I added a commit to reinstate that override and now the test is passing.  If I can do anything else let me know.

Dhaulagiri:
(saucelabs seems to have died???)

rwjblue:
restarted

PR #800, opened 2018-04-20
WrappedBuilder should cache compiled

krisselden:
WrappedBuilder should cache compiled like the other CompilableProgram implementations.


urbany:
Hi @krisselden will this fix https://github.com/emberjs/ember.js/issues/16532 ?

rwjblue:
Yep, thats the goal.

PR #804, opened 2018-04-26
Parser should strip leading newline from <pre> and <textarea>

ef4:
This accompanies https://github.com/tildeio/simple-html-tokenizer/pull/59 and won't pass CI until that is released.

tl;dr this closes a gap in our parser's spec compliance, details in the simple-html-tokenizer PR.


ef4:
Closing this because I redid the simple-html-tokenizer PR such that it is no longer a breaking change and nothing is needed here in glimmer-vm.

Issue #807, opened 2018-05-02
Rehydration Builder errors when helper produces empty string


rondale-sc:
I've created a reproduction of the issue here: https://github.com/rondale-sc/translation-related-rehydration-bug

In short, when using "ember-i18n"'s `t` helper with a key that returns an empty string causes this failure:

```
vendor.js:19840 Uncaught TypeError: Cannot read property 'parentNode' of null
    at RehydrateBuilder.remove (vendor.js:19840)
    at RehydrateBuilder.__closeBlock (vendor.js:19807)
    at RehydrateBuilder.popBlock (vendor.js:18263)
    at VM.exit (vendor.js:19416)
    at Object.evaluate (vendor.js:15864)
    at AppendOpcodes.evaluate (vendor.js:14781)
    at LowLevelVM.evaluateSyscall (vendor.js:18159)
    at LowLevelVM.evaluateInner (vendor.js:18131)
    at LowLevelVM.evaluateOuter (vendor.js:18123)
    at VM.next (vendor.js:19512)
```

rondale-sc:
Investigated this and found that it errors when "trusted content" is passed with empty string.

```
{{{""}}}
```

or

```
{{someHTMLSafeString}}
```

Will throw the error.  In GlimmerVM this throws an assertion, but by the time they are included in ember-glimmer package those assertions are stripped so you get the strange cryptic error above.

I've created some tests that illustrate the failure and 90% of the fix, but the resultant rehydrated code doesn't remove the "placeholder" node I install in the serialization build.  View more on that here:

https://github.com/rondale-sc/glimmer-vm/tree/fix-trusted-content-empty-string-bug

Hopefully, I'll figure out how to remove that node, advance the candidate / cursor, **and** remove the empty comment node.  But first, lunch. 🍽

rwjblue:
@rondale-sc - Can you confirm that this is now resolved?

rondale-sc:
@rwjblue I believe it is confirmed.  My usecase was inside ember-glimmer though, so I'm not sure I can test it fully yet.  I tried the known test failure in try.glimmerjs.org and as far as I can tell it didn't fail there.

rondale-sc:
And the test case is also satisfied.

rwjblue:
@rondale-sc ember@3.2.0-beta.5 should include these changes so you can test there

rondale-sc:
@rwjblue Didn't see this, but yes I can confirm this was resolved.

PR #809, opened 2018-05-07
Fix trusted content empty string bug

rondale-sc:


rondale-sc:
This fixes: https://github.com/glimmerjs/glimmer-vm/issues/807

The TL;DR is 

When in serialization mode we notice an empty string inside trusted content (either `{{{''}}}` or `{{<someHTMLSafeString>}}` rather than do nothing we send an empty comment node `<!-- % % -->` which in the rehydration builder we check for.  If found in the rehydration builder we remove it and advance the candidate.  

I found that since when this happens the `newBounds` (a type of Bounds) is set its `first` and `last` are the same element.  To ensure this doesn't error when `__closeBlock` is called I used `this.clearMismatch` to fudge it along.

Please let me know if this isn't the right approach.  :beers:

rondale-sc:
Rebased after prettier commit 👍

rwjblue:
Whoops! Sorry about that!

PR #811, opened 2018-05-19
[BUGFIX] Fix void element printing

chadhietala:
This PR fixes an issue where output from the printer was not actually valid HTML.
For instance preprocessing `<br>` and then printing would emit `<br></br>`, upon
parsing this the template compiler would emit an "invalid end tag" error.

This PR fixes this by propigating the `selfClosing` property from the tokenizer
into the `ElementNode`. To make sure that the syntax builder is back compat, I've
allowed the `elementNode` builder to optionally take a `TagDescriptor`, which
encapsulates the tag name and if it was authored as `selfClosing`.

### DIFF
--- a/packages/@glimmer/syntax/lib/builders.ts
+++ b/packages/@glimmer/syntax/lib/builders.ts
@@ -4,6 +4,7 @@ import { Option } from '@glimmer/interfaces';
 // Statements
 
 export type BuilderPath = string | AST.PathExpression;
+export type TagDescriptor = string | { name: string; selfClosing: boolean };
 
 function buildMustache(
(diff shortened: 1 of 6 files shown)


packages/@glimmer/syntax/lib/builders.ts
```diff
@@ -41,7 +42,7 @@ function buildBlock(
     hash: hash || buildHash([]),
     program: program || null,
     inverse: inverse || null,
-    loc: buildLoc(loc || null),
+    loc: buildLoc(losc || null),
```
rwjblue: seems like a typo? what is `losc`?

PR #812, opened 2018-05-19
Loosen path type in buildSexpr

csantero:
`ember-template-compiler` currently passes a raw string to `buildSexpr`, but the current typing for the `path` parameter is too restrictive. It only allows a `PathExpression`, even though a string can get converted to one via `buildPath`.


krisselden:
released as 0.34.4

PR #814, opened 2018-05-25
fix @glimmer/compiler types

csantero:
https://github.com/emberjs/ember.js/pull/16660 is failing because ember.js is using stricter tsconfig settings than glimmer-vm. For example:

```
node_modules/@glimmer/compiler/dist/types/lib/allocate-symbols.d.ts(22,5): error TS2416: Property 'attrSplat' in type 'SymbolAllocator' is not assignable to the same property in base type 'Processor<CompilerOps<string | 0>, number, CompilerOps<number>>'.
  Type '(_op: null) => Op<number, CompilerOps<number>, "attrSplat">' is not assignable to type '(operand: Option<string | 0>) => void | Op<number, CompilerOps<number>, "text" | "unknown" | "lit...'.
    Types of parameters '_op' and 'operand' are incompatible.
      Type 'Option<string | 0>' is not assignable to type 'null'.
        Type 'string' is not assignable to type 'null'.
```

Since I'm not particularly familiar with the `glimmer-vm` codebase, I'm not sure whether loosening parameter types and adding definite assignment assertions is the safest approach, but it does make the TS compiler happy when strict mode is on.

### DIFF
--- a/packages/@glimmer/compiler/lib/allocate-symbols.ts
+++ b/packages/@glimmer/compiler/lib/allocate-symbols.ts
@@ -77,7 +77,7 @@ export class SymbolAllocator
     this.symbolStack.pop();
   }
 
-  attrSplat(_op: null): OutOp<'attrSplat'> {
+  attrSplat(_op: Option<InVariable>): OutOp<'attrSplat'> {
     return ['attrSplat', this.symbols.allocateBlock('attrs')];
   }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Can you create an issue specifically around making glimmer-vm's config as strict as possible (at least as strict as Ember's)?

csantero:
@rwjblue Sure! https://github.com/glimmerjs/glimmer-vm/issues/815

PR #816, opened 2018-05-26
Enable stricter TS settings

### LINKED ISSUE #815: Glimmer's TS config should be at least as strict as Ember's
I ran into problems importing certain glimmer packages from TypeScript files in ember.js due to ember.js having stricter TypeScript settings than glimmer-vm.

Ember's current tsconfig:
```json
    "strict": true,
    "noImplicitReturns": true,
    "allowUnreachableCode": false,
    "suppressImplicitAnyIndexErrors": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
```

Glimmer-vm's current tsconfig:
```json
    "noImplicitAny": true,
    "suppressImplicitAnyIndexErrors": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "allowUnreachableCode": false,
    "strictNullChecks": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
```

I have a fix in for `@glimmer/compiler` at #814, but this will become more of a problem as more of ember's codebase switches to TypeScript.

csantero:
Fixes #815

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,14 +15,10 @@
 
     // Enhance Strictness
     "strict": true,
-    "noImplicitAny": true,
     "suppressImplicitAnyIndexErrors": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
-    "allowUnreachableCode": false,
-    "strictNullChecks": true,
     "noImplicitReturns": true,
-    "noImplicitThis": true,
 
     "newLine": "LF",
     "noEmit": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,6 +14,7 @@
     "moduleResolution": "node",
 
     // Enhance Strictness
+    "strict": true,
     "noImplicitAny": true,
     "suppressImplicitAnyIndexErrors": true,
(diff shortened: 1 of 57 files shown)


tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "moduleResolution": "node",
 
     // Enhance Strictness
+    "strict": true,
```
t-sauer: I think you can remove the other "strict" compiler options (like `noImplicitAny`) since `stricts` activates them all anyway.

rwjblue: Nice! I didn’t realize that. Do you know where the docs are for that (just so I can read up on it)?

csantero: @rwjblue https://www.typescriptlang.org/docs/handbook/compiler-options.html

rwjblue:
I'm going to go ahead and land this (don't want to cause merge conflicts with other changes that are in the works...), I'm happy to work on any cleanups though...

rwjblue:
Thank you again @csantero!

PR #817, opened 2018-05-27
Restore Node tests

tomdale:
Testem swallows non-zero exit codes and other error messages with the TAP reporter enabled. The Node.js tests have not been running for a few months because of this.

This PR traps process exit from the test runner, and prints a TAP-compatible "Bail out" message if it detects a non-zero exit code. With this message present, errors starting up the test runner are now reported as test failures.

I also fixed several other bugs that were preventing the CommonJS versions of the packages from working. There are a handful of remaining tests related to rehydration failing. I'm not sure how they were ever intended to work. Maybe @chadhietala can take a look at these.

### DIFF
--- a/bin/run-node-tests.js
+++ b/bin/run-node-tests.js
@@ -8,6 +8,16 @@ const EMBER_BIN = 'ember';
 const QUNIT_BIN = 'qunit';
 const NODE_TEST_GLOB = '@glimmer/{node,bundle-compiler}/test/**/*node-test.js';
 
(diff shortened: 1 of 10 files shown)


rwjblue:
Looks like we need a rebase...

rwjblue:
This seems fine generally speaking, but we should migrate to just use `qunit` for this directly...

tomdale:
@rwjblue The script shells out to the QUnit CLI utility already. Did you mean something else?

PR #820, opened 2018-05-29
Decouple `didCreateElement` from `...attributes`.

rwjblue:
This changes the generated opcodes such that `<div ...attributes></div>` no longer invokes the `didCreateElement` hook on the associated manager.

`didCreateElement` is intended only for the "wrapping element" in scenarios like `Ember.Component` (which specifically uses the `WrappedBuilder` et al).

~~glimmer.js will need some tweaking (specifically around how it represents `this.element`) before updating (hence the breaking change label).~~

### DIFF
--- a/packages/@glimmer/runtime/test/initial-render-test.ts
+++ b/packages/@glimmer/runtime/test/initial-render-test.ts
@@ -18,6 +18,7 @@ import {
   ComponentBlueprint,
   GLIMMER_TEST_COMPONENT,
   assertEmberishElement,
+  assertElement,
   assertSerializedInElement,
 } from '@glimmer/test-helpers';
 import { expect } from '@glimmer/util';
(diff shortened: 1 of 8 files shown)


rwjblue:
Hmm, I was wrong apparently we had reverted the usage of `didCreateElement` in `glimmer.js` apps (see https://github.com/glimmerjs/glimmer.js/commit/8f42d335784f4aaa7ee8ede83baa54261b4f2838). 

Given that, I would actually **not** consider this a breaking change then...

packages/@glimmer/runtime/test/initial-render-test.ts
```diff
@@ -503,10 +504,18 @@ class RehydratingComponents extends AbstractRehydrationTests {
   }
 
   assertServerComponent(html: string, attrs: Object = {}) {
+    let element: HTMLElement;
+
     if (this.testType === 'Dynamic') {
-      assertEmberishElement(this.element.childNodes[3] as HTMLElement, 'div', attrs, html);
+      element = this.element.childNodes[3] as HTMLElement;
```
tomdale: This is not a blocker, but this seems a little brittle. Do you have any ideas for how we could find the right element without special casing offsets?

packages/@glimmer/test-helpers/lib/render-test.ts
```diff
@@ -992,6 +1000,19 @@ export function content(list: Content[]): string {
   return out.join('');
 }
 
+export function processAssertElementArgs(args: any[]) {
```
tomdale: Would you mind adding a few comments here describing expected inputs and outputs? It's hard for my poor brain to reverse the intended data structures from the code below.

packages/@glimmer/runtime/test/initial-render-test.ts
tomdale: Ah, that sounds like a likely explanation. Again, I know you're inheriting this bit of code, so please don't consider it urgent or a blocker, unless there's an obvious, trivial improvement we can sneak in.

packages/@glimmer/test-helpers/lib/render-test.ts
rwjblue: OK, updated with some docs (and a TODO to make it clearer that this needs future cleanup "when we have time").

packages/@glimmer/runtime/test/initial-render-test.ts
rwjblue: OK, so I confirmed our suspicion here. When `{{component` is rendered in SSR we add comment node markers for it and therefore have to account for that when reading which of the childNodes we care about. I cleaned up the logic a little and left a comment to that effect.

PR #822, opened 2018-05-30
Implement dynamic angle bracket invocation.

rwjblue:
This supports the following dynamic angle bracket invocations:

* Invoking a local (block param):

```
{{#with (component 'special-sauce') as |OtherThing|}}
  <OtherThing data-foo={{thingsHere}} @namedStuff="goes here" />
{{/with}}
```

* Invoking a named argument:

```
<@someThing data-foo="wheeeeee" />
```

* Invoking a path expression:

```
<this.thing.here />
```

All forms support a template, named arguments, and attributes.

---

Significant changes (implementation wise):

* Move Component detection from JavascriptCompiler to TemplateCompiler (required to support emitting different opcodes for dynamic invocation vs static invocation).
* Add `openComponent`, `closeDynamicComponent`, and `closeComponent` template compiler opcodes.
* Add `DynamicComponent` wire-format opcode. During runtime opcode building the `DynamicComponent` opcode invokes `builder.dynamicComponent` (very similarly to `{{component`).
* Add support for `AttrSplat` nodes passed during component invocation (was previously disabled due to implementation details around `openElement` and `openSplattedElement` which were simplified in this PR).
* Add support for an `AttrBlock` to `Builder.prototype.dynamicComponent`.
* Add error when using `{{` in the `TokenizerStates.tagOpen` or `TokenizerState.tagName` states.
* Add tests for various combinations of invocation source (path, local, named arg) and invocation capabilities (passing args, attributes, template).

---

TODO:

- [x] Add tests for using paths `<this.foo />`
- [x] Add tests for using blocks with all three dynamic types
- [x] Remove `rwjblue` prefix from tests added
- [x] Add tests for a component with `...attributes` on another component invocation
- [x] Add template compilation error for things like `<{{"asdf"}}></{{"asdf"}}>`
- [x] Add tests confirming `<foo/bar></foo/bar>` and `<Foo/Bar></Foo/Bar>` does not work
- [x] ~~Update `ElementNode`'s `tag` property to be a new type: `Identifier` (roughly similar to path expression)~~ moving to another PR...

Related to https://github.com/emberjs/ember.js/issues/16688.

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -81,7 +81,11 @@ export default class TemplateCompiler {
       }
     }
 
-    if (hasSplat) {
+    if (isComponent(action)) {
+      //let [head, ...rest] = action.tag;
+      //this.opcode(['get', [head, rest]]);
+      this.opcode(['openComponent', action], action);
+    } else if (hasSplat) {
       this.opcode(['openSplattedElement', action], action);
     } else {
       this.opcode(['openElement', action], action);
@@ -108,7 +112,11 @@ export default class TemplateCompiler {
   }
 
   closeElement([action]: [AST.ElementNode]) {
-    this.opcode(['closeElement', action], action);
+    if (isComponent(action)) {
+      this.opcode(['closeComponent', action], action);
+    } else {
+      this.opcode(['closeElement', action], action);
+    }
   }
 
   attribute([action]: [AST.AttrNode]) {
@@ -441,6 +449,24 @@ function isBuiltInHelper(path: AST.PathExpression) {
 
 function isArg(path: AST.PathExpression): boolean {
   return !!path['data'];
+}
+
+function isDynamicComponent(element: AST.ElementNode): boolean {
+  let open = element.tag.charAt(0);
+
+  let isNamedArgument = open === '@';
+  let isLocal = element['symbols'].has(element.tag);
+  let isPath = element.tag.indexOf('.') > -1;
+
+  return isLocal || isNamedArgument || isPath;
+}
+
+function isComponent(element: AST.ElementNode): boolean {
+  let open = element.tag.charAt(0);
+
+  let isUpperCase = open === open.toUpperCase() && open !== open.toLowerCase();
+
+  return isUpperCase || isDynamicComponent(element);
 }
 
 function assertIsSimplePath(path: AST.PathExpression, loc: AST.SourceLocation, context: string) {
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -5,6 +5,7 @@ import { AST, isLiteral, SyntaxError } from '@glimmer/syntax';
 import { getAttrNamespace } from './utils';
 import { Opaque } from '@glimmer/interfaces';
 import { SymbolAllocator, InOp as SymbolInOp, OutOp as SymbolOutOp } from './allocate-symbols';
+import { PathHead } from './compiler-ops';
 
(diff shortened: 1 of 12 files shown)


rwjblue:
OK, I believe this is ready for review.

packages/@glimmer/test-helpers/lib/suites/components.ts
```diff
@@ -163,4 +164,382 @@ export class BasicComponents extends RenderTest {
     this.assertHTML("<div color='red'>left - hello! yield me</div>");
     this.assertStableNodes();
   }
+
+  @test({
```
chadhietala: Aren't all of these testing Glimmer Components?

rwjblue: I can change these to say `type: 'glimmer'`, but the reason I used `basic` was that _all_ of the other tests in this class were listed as `basic` and the name of the class is `BasicComponents`...

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -81,7 +81,17 @@ export default class TemplateCompiler {
       }
     }
 
-    if (hasSplat) {
+    if (isDynamicComponent(action)) {
+      let head: string | 0, rest: string[];
```
tomdale: Is there a better type we can use here than `string | 0`? This took me a few attempts to read to understand what was going on, particularly since the zero literal type looked like a bitwise operation at first.

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -443,6 +459,24 @@ function isArg(path: AST.PathExpression): boolean {
   return !!path['data'];
 }
 
+function isDynamicComponent(element: AST.ElementNode): boolean {
+  let open = element.tag.charAt(0);
+
+  let isNamedArgument = open === '@';
+  let isLocal = element['symbols'].has(element.tag);
```
tomdale: Why do we need to use bracket notation here? Does `symbols` need to be added to the type?

packages/@glimmer/opcode-compiler/lib/opcode-builder.ts
```diff
@@ -883,7 +883,8 @@ export abstract class OpcodeBuilder<Locator = Opaque> extends StdOpcodeBuilder {
 
   dynamicComponent(
     definition: WireFormat.Expression,
-    /* TODO: attrs: Option<RawInlineBlock>, */ params: Option<WireFormat.Core.Params>,
```
tomdale: The first useful TODO in the history of the world. 😁

packages/@glimmer/test-helpers/lib/suites/components.ts
tomdale: I am also confused about the intent of basic components.

packages/@glimmer/compiler/lib/template-compiler.ts
rwjblue: Ah, yes, great point.  This was a bit of fumbling on my part actually. We actually expect either `0` (which is what `this` is represented as) or the string for the "head" of the path.  I'll try to do some tweaking to make this type more intention revealing...

packages/@glimmer/test-helpers/lib/suites/components.ts
chadhietala: @wycats or @chancancode should confirm, but "basic" started out as component type that didn't have all the capabilities that a `EmberishGlimmerComponent` or a `EmberishCurlyComponent` had. Was used as quick way to just test the mechanics of the VM. This concept may not matter any more, in that case we should purge the type of the test suite. Likely will simplify things.

PR #823, opened 2018-05-31
Add implicit ...attributes to wrapped layouts.

rwjblue:
This implements the implicit `...attributes` support for wrapped layouts (which is what `Ember.Component` uses) as specified in the [Angle Bracket Invocation RFC](https://github.com/emberjs/rfcs/blob/master/text/0311-angle-bracket-invocation.md#html-attributes):

> Classic components (Ember.Component) will implicitly have an ...attributes added to the end of the wrapper element (if tagName is not an empty string), after any attributes added by the component itself (using attributeBindings, classNames etc). This means that attributes provided by the caller will override (replace) those added by the component (except for class, which is merged).

TODO:

- [x] add tests confirming merging behaviors (invocation attributes clobber, except `class` which merges)

### DIFF
--- a/packages/@glimmer/opcode-compiler/lib/wrapped-component.ts
+++ b/packages/@glimmer/opcode-compiler/lib/wrapped-component.ts
@@ -22,6 +22,7 @@ import { EMPTY_ARRAY } from '@glimmer/util';
 export class WrappedBuilder<Locator> implements CompilableProgram {
   public symbolTable: ProgramSymbolTable;
   private compiled: Option<number> = null;
+  private attrsBlockNumber: number;
 
   constructor(
     private compiler: Compiler<OpcodeBuilder<Locator>>,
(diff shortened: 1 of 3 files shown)


packages/@glimmer/opcode-compiler/lib/wrapped-component.ts
```diff
@@ -22,17 +22,27 @@ import { EMPTY_ARRAY } from '@glimmer/util';
...
+  private attrsBlockNumber: number;
 
   constructor(
     private compiler: Compiler<OpcodeBuilder<Locator>>,
     private layout: LayoutWithContext<Locator>
   ) {
     let { block } = layout;
 
+    let symbols = block.symbols.slice();
     this.symbolTable = {
```
chadhietala: Can you move this assignment to L45 just so it's easier to understand that you are going to mutate `symbols` and place it in the `symbolTable`.

rwjblue: Yep, makes sense. Will update.

rwjblue: Done! Thanks for reviewing!

PR #832, opened 2018-06-19
Add `value` method to `ICapturedArguments`.

rwjblue:
This seems to have been an oversight (likely hasn't mattered to date?), both `ICapturedNamedArguments` and `ICapturedPositionalArguments` already support `.value()`. This brings `ICapturedArguments` in line.

Supports changes in https://github.com/emberjs/ember.js/pull/16747.


krisselden:
@rwjblue you need to update the Checker<> type with this change.

rwjblue:
Updated to make the checker pass, and pinged @chancancode and @krisselden for review. If there is a better way, I'm happy to iterate...

chancancode:
seems fine to me. not directly related, but we used to do more hand-crafted tuning for the empty cases because we believed them to be common (e.g. `value` could return the same const empty object every time), did we decide it's not worth it?

PR #841, opened 2018-08-13
Remove attribute if property is `falsy` when rehydrating

bobisjan:
👋,

we've tried rehydration in our application and found 🐛, when some property (eq. `fastboot.isFastBoot`) has changed between server side and client side rendering.

This is an attempt to fix issue as I'm not familiar with Glimmer VM internals 🙈. It seems fine to reuse `normalizeValue()` method here, but it requires to update some other assertions, because of the `true` => `""` normalization rule.

Failing test before fix was applied:

screen shot 2018-08-13 at 13 12 07

Should fix https://github.com/ember-fastboot/ember-cli-fastboot/issues/594.


bobisjan:
@glimmerjs/glimmer.js what do you think about this?

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -325,7 +326,12 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
       let attr = findByName(unmatched, name);
       if (attr) {
         if (attr.value !== value) {
-          attr.value = value;
+          let normalizedValue = normalizeValue(value);
```
chadhietala: The value should be normalized by the time you get here.

bobisjan:
@chadhietala, can you please take a look on the updated approach?

I can not normalize values in `DefaultDynamicProperty`, because of other attributes (eq. `name="false"`), thus `BooleanDynamicAttribute` class with `isBooleanAttribute` helper (which needs to be properly updated if this solution is appropriate).

bobisjan:
@chadhietala friendly ping about what do you think about proposed solution ☺️.

packages/@glimmer/runtime/lib/vm/attributes/dynamic.ts
```diff
@@ -188,6 +192,22 @@ export class OptionSelectedDynamicAttribute extends DefaultDynamicProperty {
...
+  }
+
+  update(value: Opaque, env: Environment): void {
+    let normalizedValue = normalizeValue(value);
+    super.update(normalizedValue, env);
+  }
+}
+
+function isBooleanAttribute(tagName: string, attribute: string) {
+  return tagName === 'BUTTON' && attribute === 'disabled';
```
chadhietala: Why is this narrowed to `BUTTON`?

bobisjan: Just for clarity in this POC i would say, the complete list of attributes is 
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes, and disabled attribute is applicable to more tags, like i said this is just an approach which needs to be updated according to the list of Boolean attributes. I’ll update the list if this approach seems fine to you ☺️

packages/@glimmer/runtime/lib/vm/attributes/dynamic.ts
```diff
@@ -215,3 +235,20 @@ function normalizeValue(value: Opaque): Option<string> {
 
   return String(value);
 }
+
+function normalizeBooleanValue(value: Opaque): Option<boolean> {
```
bobisjan: Based on https://github.com/glimmerjs/glimmer-vm/pull/841/files#diff-3ce638612baaf18381842996696f1f37R219, but for `boolean` values.

packages/@glimmer/test-helpers/lib/suites/initial-render.ts
```diff
@@ -256,6 +256,50 @@ export class InitialRenderSuite extends RenderTest {
     this.assertStableRerender();
   }
 
+  @test
```
bobisjan: These tests are similar (copy&paste) to [Unquoted attribute null values do not disable](https://github.com/glimmerjs/glimmer-vm/pull/841/files#diff-c826bfbc3bab3c7ef42bc6f7884b1ebbR238), but updated for other `falsy` values.

tomdale:
@bobisjan Thank you for the hard work on this!

I think we need to give this some more discussion, because we're trying to move away from the guess-the-right-behavior whackamole of the DOM API (which as you note is poorly specified) and instead move to a model where developers can explicitly set, add, and remove properties and/or attributes, and Glimmer just reflects that into the DOM without trying to guess what you really want.

@wycats is at TC39 this week but let me try to follow up with him and try to give an answer here one way or another.

bobisjan:
Hi @tomdale and @wycats,

do we have some update on this?

mydea:
FYI I just stumbled over this issue (when trying to make a button disabled in fastboot and enable it only when clicking the button actually does something), so this seems to still be an issue!

NullVoxPopuli:
Anyone want to rebase and get this across the finish line?

PR #843, opened 2018-08-20
Introduce ModifierDefinitonState

chadhietala:
This introduces the notion of `ModifierDefinitionState`. The purpose of this state
is to encapsulate values that are required for creation of the modifier. For instance the `ModifierDefinitionState` can contain the factory that backs a user defined modifier. This is not a new concept and is in fact why `ComponentDefinitionState` exists.

In the future there is opportunity to unify these ideas. This is the first step.

### DIFF
--- a/packages/@glimmer/runtime/lib/internal-interfaces.d.ts
+++ b/packages/@glimmer/runtime/lib/internal-interfaces.d.ts
@@ -6,3 +6,8 @@ export {
   ComponentDefinitionState,
   InternalComponentManager as ComponentManager,
 } from './component/interfaces';
+
+export {
+  InternalModifierManager as ModifierManager,
+  ModifierInstanceState as Modifier,
+} from './modifier/interfaces';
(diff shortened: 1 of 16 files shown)


packages/@glimmer/runtime/lib/internal-interfaces.d.ts
```diff
@@ -6,3 +6,8 @@ export {
   ComponentDefinitionState,
   InternalComponentManager as ComponentManager,
 } from './component/interfaces';
+
+export {
+  InernalModifierManager as ModifierManager,
```
tomdale: s/Inernal/Internal/

packages/@glimmer/runtime/lib/modifier/interfaces.ts
```diff
@@ -2,30 +2,57 @@ import { IArguments } from '../vm/arguments';
...
+  ModifierDefinitionState = Opaque,
+  Manager = ModifierManager<Opaque, ModifierDefinitionState>
+> {
+  state: ModifierDefinitionState;
+  manager: Manager;
+}
+
+/* @internal */
+export interface ModifierDefinition {
+  manager: InernalModifierManager;
```
tomdale: s/Inernal/Internal/

packages/@glimmer/runtime/lib/modifier/interfaces.ts
```diff
@@ -2,30 +2,57 @@ import { IArguments } from '../vm/arguments';
...
+  state: ModifierDefinitionState;
+}
+
+/* @internal */
+export type InernalModifierManager = ModifierManager<
```
tomdale: s/Inernal/Internal/

PR #844, opened 2018-08-25
Async-rerender proof of concept

lifeart:
Donno why `commit` in env running before `didCreate/didUpdate` etc.
Checking for `transaction != null` solve this issue, but I don't like this approach

part of work in https://github.com/glimmerjs/glimmer.js/pull/134

### DIFF
--- a/packages/@glimmer/runtime/lib/environment.ts
+++ b/packages/@glimmer/runtime/lib/environment.ts
@@ -257,10 +257,16 @@ export abstract class Environment {
   }
 
   didCreate(component: Component, manager: ComponentManager) {
+    if (!this.transaction) {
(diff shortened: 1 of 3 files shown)


tomdale:
@lifeart Cool experiment! Our thinking was that async revalidation should be modeled similarly to the initial render, where an iterator is returned, and the host environment can call `.next()` synchronously until revalidation is complete. This allows the host environment to customize and experiment with async scheduling, without coupling the VM to any particular approach.

Anything we can do to help unblock this?

lifeart:
@tomdale thank you for pointing! I don't have time in December to work on this :(
Feel free to close this pr

PR #847, opened 2018-09-05
Ensure dynamic component invocations follow splattribute rules.

rwjblue:
When the `DynamicComponent` opcode was initial introduce, it did not properly replicate the `SetComponentAttrs` toggling system that `Ops.Component` has. This has the unfortunate result of making dynamic component invocations not follow the established rules for splattributes (invocation "wins" for all attributes other than `class` and `class` is merged). The implementation here for attrs block created here, is **directly** taken from the `Ops.Component` opcode just below...

In this context `DynamicComponent` is specifically referring to angle bracket invocation with either a named argument or path.

---

Once merged, this needs to be backported into the version of glimmer-vm in use by Ember 3.4.x to fix the bug reported by @cibernox in https://github.com/emberjs/ember.js/pull/16933 and https://github.com/rwjblue/sparkles-component/issues/7.

### DIFF
--- a/packages/@glimmer/test-helpers/lib/suites/components.ts
+++ b/packages/@glimmer/test-helpers/lib/suites/components.ts
@@ -201,6 +201,64 @@ export class BasicComponents extends RenderTest {
   @test({
     kind: 'glimmer',
   })
+  'invoking curried component with attributes via angle brackets (invocation attributes clobber)'() {
+    this.registerHelper('hash', (_positional, named) => named);
+    this.registerComponent(
+      'Glimmer',
+      'Foo',
+      '<p data-foo="default" ...attributes>hello world!</p>'
+    );
+    this.render({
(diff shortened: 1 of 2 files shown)


packages/@glimmer/test-helpers/lib/suites/components.ts
```diff
@@ -198,6 +198,60 @@ export class BasicComponents extends RenderTest {
     this.assertStableRerender();
   }
 
+  @test({
+    kind: 'glimmer',
+  })
+  'invoking curried component with attributes via angle brackets (invocation attributes clobber)'() {
+    this.registerHelper('hash', (_positional, named) => named);
+    this.registerComponent('Glimmer', 'Foo', '<p data-foo="default"...attributes>hello world!</p>');
```
lennyburdette: nitpicking: space between `"default"` and `...attributes`

PR #848, opened 2018-09-06
Refactor publishing script to allow for OTP timeouts.

rwjblue:
The recent requirement to use OTP's for the @glimmer scope required changes to `bin/publish.js` in order to maintain functionality. An earlier commit (f58ee9a) add initial support, but publishing all of our packages simply doesn't fit within a single OTP window (it seems to take at least 2 full OTP windows, but I've had to do as many as 4 when either my local network was slow or npm registry was slow).

This PR refactors things a bit to fix this issue by allowing recovering of a E401 publishing error and prompting for an updated OTP token then retrying.

It also ensures that the correct dist tag is used (the prior logic only accounted for `next` and `latest` but caused many issues when we need to do patch releases of older versions).

---

Note, this now requires using Node 10.3 or higher (which seems perfectly fine to me...).

### DIFF
--- a/bin/publish.js
+++ b/bin/publish.js
@@ -135,6 +135,18 @@ function gitCommitAndTag() {
   execWithSideEffects(`git tag "v${newVersion}"`);
 }
 
+async function getOTPToken() {
+  let token = await question(chalk.green('\nPlease provide OTP token '));
+
+  return token.trim();
+}
+
+function publishPackage(distTag, otp, cwd) {
+  execWithSideEffects(`npm publish --tag ${distTag} --access public --otp ${otp}`, {
+    cwd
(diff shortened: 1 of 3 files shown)


bin/publish.js
```diff
@@ -125,35 +134,53 @@ function gitCommitAndTag() {
...
 
-      console.log(chalk.green(`\nv${newVersion} deployed!`));
-      console.log(chalk.green('Done.'));
-      cli.close();
-    });
-  });
+    try {
+      publishPackage(distTag, otp, package.absolutePath);
+    } catch(e) {
+      debugger
```
rwjblue: Whoops, stray debugger!

PR #849, opened 2018-09-07
Failing test: incorrect loc for attrs following valueless attr.

rwjblue:
Given the following:

```hbs
<div
  data-foo
  class="blah"
></div>
```

We currently suggest that the `AttrNode` for `class` starts _after_ the `c` in `class`, this is definitely incorrect. The fix will need to be made in simple-html-tokenizer.

Originally reported in https://github.com/ember-template-lint/ember-template-lint/issues/483

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "babel-plugin-debug-macros": "^0.1.11",
     "handlebars": "^4.0.6",
     "simple-dom": "^0.3.0",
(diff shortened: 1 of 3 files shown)


rwjblue:
Will be fixed by https://github.com/tildeio/simple-html-tokenizer/pull/65 once released and bumped here...

PR #850, opened 2018-09-14
[WIP] Exposing a primitive renderComponent API

chiragpat:
Currently in order to render a component from the server with a json blob requires a lot of work arounds and has applications use the the dynamic component API. This pull request aim's to create a new renderComponent API in the glimmer-vm that the host environments can use to provide a simpler interface to support such a use case.


packages/@glimmer/runtime/lib/render.ts
```diff
@@ -27,3 +31,69 @@ export default function render<T>(
...
+
+  // Get a list of tuples of argument names and references, like
+  // [['title', reference], ['name', reference]]
+  const argList = Object.keys(args).map(key => [key, args[key]]);
+
+  const blockNames = ['main', 'else', 'attrs'];
+  // Prefix argument names with `@` symbol
+  const argNames = argList.map(([name]) => `@${name}`);
+
+  vm.pushFrame();
```
chadhietala: I'm trying to remember if there is a `popFrame` when the iterator is exhausted or not. Do you happen to know?

chiragpat: I walked through the opcodes the final set of ops the vm is running in order are,

PopFrame, PopScope, PopDynamicScope, CommitComponentTransaction, Return

PR #852, opened 2018-09-23
Fix emberjs/ember.js#14978

chancancode:
emberjs/ember.js#14978

Likely related:

emberjs/ember.js#14924
emberjs/ember.js#16172
emberjs/ember.js#16314

Closes #791

### DIFF
--- a/packages/@glimmer/runtime/lib/dom/helper.ts
+++ b/packages/@glimmer/runtime/lib/dom/helper.ts
@@ -210,22 +210,15 @@ export function insertHTMLBefore(
   useless: HTMLElement,
   _parent: Simple.Element,
   _nextSibling: Option<Simple.Node>,
-  html: string
+  _html: string
 ): Bounds {
-  // tslint:disable-line
-  // TypeScript vendored an old version of the DOM spec where `insertAdjacentHTML`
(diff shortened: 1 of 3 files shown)


packages/@glimmer/runtime/lib/dom/helper.ts
```diff
@@ -210,22 +210,15 @@ export function insertHTMLBefore(
...
+  let parent = _parent as Element;
+  let nextSibling = _nextSibling as Option<Node>;
 
   let prev = nextSibling ? nextSibling.previousSibling : parent.lastChild;
   let last: Simple.Node | null;
 
-  if (html === null || html === '') {
-    return new ConcreteBounds(parent, null, null);
-  }
+  let html = _html || '<!---->';
```
chancancode: Actually this may not be good enough since blank but non-empty strings may have the same problem?

chancancode: I thought deleting the conditional would do the job, but at least on chrome, `insertAdjecentHTML` with an empty string is a no-op (is that spec compliant)? Unsure if blank strings has the same “optimization” (hence problem).

chancancode:
Digging through the history, it turns out @mongoose700 and I independently arrived at the same fix, so I added them as a co-author.

PR #853, opened 2018-09-25
Move stack interface, to interfaces.

billybonks:
While exporting stack from `packages/@glimmer/runtime/lib/vm/low-level.ts` is good for people who code against low level, it doesn't help find the implementation of the interface.

To resolve this i moved the interface interfaces, and marked `EvaluationStack` as implementing it. it seems that wasmStack and innerStack, don't implement the interface wholly, but at least it seems clear now that the stack used within low level is the `EvaluationStack`.

If this is a welcomed pr i will make more of them as i find more cases while exploring the code base.

### DIFF
--- a/packages/@glimmer/interfaces/index.d.ts
+++ b/packages/@glimmer/interfaces/index.d.ts
@@ -6,6 +6,7 @@ export * from './lib/module-locators';
 export * from './lib/tier1/symbol-table';
 export * from './lib/template';
 export * from './lib/serialize';
+export * from './lib/stack';
 export { default as ComponentCapabilities } from './lib/component-capabilities';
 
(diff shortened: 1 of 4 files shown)


billybonks:
This makes sense and is a good rational as to when things go into the interfaces folder.

I was trying to enable better code exploration by using typescript features. There are currently a few classes that look like they could be a stack, and there was no quick way of seeing which on is consumed by `LowLevelVM` when using `find all references`

screen shot 2018-10-16 at 11 06 50 pm

After this pr the same ui in visual studio shows

screen shot 2018-10-16 at 11 05 19 pm

locks:
With the rewrites since this PR I don't think it longer applies. Thanks for your contribution!

PR #856, opened 2018-09-27
`Bounds` must be non-empty

chancancode:
💣 contains breaking changes

The `Bounds` object represents a _inclusive_ range. That is, the bounds is referring to the range in DOM _including_ the `firstNode()` and the `lastNode()`. Therefore, it does not ever make sense for those endpoints to be `null`. (Technically, `(parent, null, null)` is a valid way to represent an _empty_ range. However, in practice we do not have a use case for this, and every use of `null` were actual bugs.)

This is technically a breaking change, but then again, if `null` were passed today it's likely a bug, and if the consuming side was checking for `null` it's likely just deadcode that will never execute in practice.

### DIFF
--- a/packages/@glimmer/node/lib/node-dom-helper.ts
+++ b/packages/@glimmer/node/lib/node-dom-helper.ts
@@ -1,6 +1,7 @@
 import * as SimpleDOM from 'simple-dom';
 import { DOMTreeConstruction, Bounds, ConcreteBounds } from '@glimmer/runtime';
 import { Simple } from '@glimmer/interfaces';
+import { Option } from '@glimmer/util';
 
 export default class NodeDOMTreeConstruction extends DOMTreeConstruction {
   protected document!: SimpleDOM.Document; // Hides property on base class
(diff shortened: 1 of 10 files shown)


packages/@glimmer/node/lib/node-dom-helper.ts
```diff
@@ -11,16 +12,10 @@ export default class NodeDOMTreeConstruction extends DOMTreeConstruction {
...
-    let prev = reference ? reference.previousSibling : parent.lastChild;
-
+  insertHTMLBefore(parent: Simple.Element, reference: Option<Simple.Node>, html: string): Bounds {
     let raw = this.document.createRawHTMLSection(html);
     parent.insertBefore(raw, reference);
-
-    let first = prev ? prev.nextSibling : parent.firstChild;
-    let last = reference ? reference.previousSibling : parent.lastChild;
-
-    return new ConcreteBounds(parent, first, last);
```
chancancode: This is a bug (similar in nature to #852) in FastBoot. Since you don't normally run updates there I don't know how much this matters (i.e. should I split this into a BUGFIX PR and release that separately).

packages/@glimmer/runtime/lib/compat/svg-inner-html-fix.ts
```diff
@@ -30,41 +31,51 @@ export function applySVGInnerHTMLFix(
...
+      if (html === '') {
+        return super.insertHTMLBefore(parent, nextSibling, html);
+      }
+
       if (parent.namespaceURI !== svgNamespace) {
         return super.insertHTMLBefore(parent, nextSibling, html);
       }
 
-      return fixSVG(parent, div, html, nextSibling);
+      // TODO: why are these casts okay???
```
chancancode: @chadhietala @wycats

packages/@glimmer/runtime/lib/dom/helper.ts
```diff
@@ -134,10 +139,44 @@ export class DOMOperations {
...
     html: string
   ): Bounds {
-    return insertHTMLBefore(this.uselessElement, _parent, nextSibling, html);
+    if (html === '') {
+      let comment = this.createComment('');
+      _parent.insertBefore(comment, _nextSibling);
+      return new ConcreteBounds(_parent, comment, comment);
+    }
+
+    // TODO why are these casts okay???
```
chancancode: @chadhietala @wycats

chancancode:
I don’t think so. I also removed some exports too, but I don’t know of any consumers

chancancode:
On the other hand, it doesn’t cost anything to bump the minor

packages/@glimmer/runtime/lib/compat/svg-inner-html-fix.ts
chadhietala: The mental model is that `DOMChanges` is dealing with live elements. Updating doesn't mean anything in SSR case.

PR #862, opened 2018-11-06
[DOCS] Adds runtime documentation

pzuraq:
Been trying to understand the Glimmer VM at a higher level, and how it integrates with Ember and Glimmer.js. I figured it would be good to document my findings as I go through this.

If any of this is incorrect let me know! Still trying to learn and figure out how it all works, both at a high level and a low-level, and would love to get feedback on any parts that could be better 😄  

### Notes

* The Glimmer templating languages has diverged a lot from Handlebars, is it still accurate to refer to them as Handlebars templates throughout the docs?
* Unsure if the Application Lifecycle section should be in the introduction or in the runtime section. Really going for a high level lifecycle, which definitely glosses over large parts of the runtime, but I think is more helpful from the perspective of _consumers_ of Glimmer-VM to understand this lifecycle.
* I think the `Compiler` section in the intro should be merged with the precompiler page, seems like it gets too detailed too quickly. Most details mentioned there are implementation details which _probably_ aren't too valuable to consumers of Glimmer-VM (though I may be wrong).

### External Sources

* [Gavin and Yehuda's deep dive (2016)](https://www.youtube.com/watch?v=vL8sCi1Bv6E)
* [Gavin's 2017 Emberconf talk](https://www.youtube.com/watch?v=A5k2u7AZf6g)

### DIFF
--- a/guides/03-runtime-overview.md
+++ b/guides/03-runtime-overview.md
@@ -2,7 +2,7 @@
 
 1. [Introduction](./01-introduction.md)
 2. [Precompiler Overview](./02-precompiler-overview.md)
-3. [~~Runtime Overview~~](./03-runtime-overview.md)
+3. [Runtime Overview](./03-runtime-overview.md)
 4. [References](./04-references.md)
 5. [Validators](./05-validators.md)
 6. [~~Runtime Compiler~~](./06-runtime-compiler.md)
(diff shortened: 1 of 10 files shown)


guides/03-runtime-overview.md
```diff
@@ -13,7 +13,37 @@
 
 # Runtime Overview
 
-TODO: add content here
+The Glimmer runtime takes the compiled bytecode discussed in the previous section and renders it in the browser. At a high level, the runtime begins by rendering a component tree, beginning with one root component and walking down through its templates. While doing this, it saves an optimized set of bytecodes which can be used later on when rerendering the application. These update codes ignore content which is purely static, such as DOM elements and helpers or components which have static inputs, and adds updating instructions for dynamic content.
```
tomdale: Seems like this is actually describing the compilation step? I.e. "saves an optimized set of bytecodes" is describing the opcode compiler. This frame is also specific to the lazy/runtime compiler mode, and doesn't hold if you're using Glimmer in the eager/bytecode compiler mode.

tomdale: Ahhh, I see. You're describing the building of the updating bytecode, but I was interpreting it as describing the just-in-time template compilation of lazy mode. The language here I think could be interpreted to apply to either.

Let's merge this, but revisit if we can make it clearer that this is specifically referring to the partial evaluation optimization.

PR #863, opened 2018-11-07
Fixed Windows Build Error

mattselden12:
Ran yarn test, got a permissions error attempting to create the symlinks, changed to use junction instead since it was trying to symlink a directory.  Also got error from testem- Node launcher relied on being executed with a shell.

### DIFF
--- a/build/symlink-dependencies/index.js
+++ b/build/symlink-dependencies/index.js
@@ -49,7 +49,8 @@ module.exports = {
         if (isScopedPackage(dep)) { mkdirpScope(dep, modulesPath); }
         fs.symlinkSync(
           path.join('../../../../', dep),
-          path.join(modulesPath, dep)
+          path.join(modulesPath, dep),
+          'junction'
         );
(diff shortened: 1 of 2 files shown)


build/symlink-dependencies/index.js
```diff
@@ -49,7 +49,8 @@ module.exports = {
         if (isScopedPackage(dep)) { mkdirpScope(dep, modulesPath); }
         fs.symlinkSync(
           path.join('../../../../', dep),
-          path.join(modulesPath, dep)
+          path.join(modulesPath, dep),
+          'junction'
```
rwjblue: Is this conditional? In other words, do we have to detect the platform first and do one thing for windows and another for *nix?

mattselden12: I believe it is ignored outside of Windows, and it passed CI so I would assume that makes it okay.

rwjblue:
Thank you!

PR #865, opened 2018-11-19
Destructor redux

wycats:
Update the destructor API to make it more general.

Now, instead of a bunch of separate, bespoke destructor APIs, there is a single API for associating a destructor with a parent object. Instead of just calling `destroy()` on an object to destroy it, you destroy it with an environment, which (1) makes the destruction async, (2) recursively executes any associated destructors.


rwjblue:
@wycats - This seems to include the changes from the named-blocks branch as well as some additional changes, is that intentional? Do you intend to land this PR before or after #864?

PR #867, opened 2018-11-29
[WIP] Attempt more graceful recovery of render-time exceptions

tomdale:
This PR attempts to implement basic graceful recovery of VM state in the event of an exception getting thrown during render. The most important constraint is that subsequent revalidations after a failed revalidation are possible, i.e. a single exception during render doesn't crash the entire application and require a page reload.

I don't feel confident that this approach doesn't introduce its own problems and would like review from folks who are more familiar with this part of the architecture.

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/append.ts
+++ b/packages/@glimmer/runtime/lib/vm/append.ts
@@ -478,12 +478,21 @@ export default class VM<T> implements PublicVM {
     return result.value;
   }
 
+  recover() {
+    this.elementStack.panic();
+  }
+
   next(): IteratorResult<RenderResult> {
     let { env, program, updatingOpcodeStack, elementStack } = this;
     let opcode = this.inner.nextStatement();
     let result: IteratorResult<RenderResult>;
     if (opcode !== null) {
-      this.inner.evaluateOuter(opcode, this);
+      try {
+        this.inner.evaluateOuter(opcode, this);
+      } catch (err) {
+        this.recover();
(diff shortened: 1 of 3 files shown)


krisselden:
looks like sauce labs flaked out on the test of the PR.

packages/@glimmer/runtime/lib/vm/append.ts
```diff
@@ -478,12 +478,21 @@ export default class VM<T> implements PublicVM {
...
+    this.elementStack.panic();
+  }
+
   next(): IteratorResult<RenderResult> {
     let { env, program, updatingOpcodeStack, elementStack } = this;
     let opcode = this.inner.nextStatement();
     let result: IteratorResult<RenderResult>;
     if (opcode !== null) {
-      this.inner.evaluateOuter(opcode, this);
+      try {
```
krisselden: should we use the babel-debug-helpers to make it so DEBUG mode this catch doesn't happen?

tomdale:
@wycats believes this is better handled by the destructor infrastructure put in place in his `destructor-redux` branch. We are trying to decide if it's better to work on landing that rather than temporarily merging this admittedly hacky fix.

tomdale:
The implementation proposed here had too many problematic edge cases that could cause elements to be duplicated and/or inserted into the wrong position in DOM on re-renders. The `destructor-redux` branch has since been merged into master, so our ability to recover from errors should be much better rationalized now.

PR #869, opened 2018-12-14
use href attribute of svg element without quotation marks behaves differently than one without

mhankus:
Due to some problem with ember app, I've filed bug 
https://github.com/emberjs/ember.js/issues/16968

@rwjblue suggested to create test (and help me to do it - thank you) and attach here

Problem can be observed here:
https://ember-twiddle.com/1433f1b137d34c9422dc88eda124c2d0

There are 2 tests, one is passing, one is not. I'm testing value of href attribute (which is of type SVGAnimatedString in browser DOM) - this is the only difference I've noticed.


rwjblue:
Nice! Thank you for submitting these tests! I’ll try to circle around with the team to come up with next steps...

tomdale:
Huh, I thought we had stopped treating these cases differently (i.e. using quotes to force string coercion).

rwjblue:
@tomdale - Ya, I was also pretty surprised that these tests fail. I'm just not sure where to dig in...

mhankus:
This is namespace problem. When you open my fiddle
 https://ember-twiddle.com/1433f1b137d34c9422dc88eda124c2d0 
and remove namespace declaration and prefix so that code looks like

```<use href={{iconLink}}></use>```

everything renders ok. 
Problem is that according to https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use href without xlink namespace  is not supported by Safari - I'm on windows and cannot verify if it is correct information.

**Comment by mhankus** [comment]
Hello again.
 I'was looking around source code, and found suspicious piece of code 

template-compiler.ts lines 179-180
```
else if (action.value.type === 'MustacheStatement') {
        this.opcode([isComponent ? 'componentAttr' : 'dynamicAttr', [name, null]], action);
      } else {
        this.opcode([isComponent ? 'componentAttr' : 'dynamicAttr', [name, namespace]], action);
      }
```
if type is MustacheStatement and it is dynamicAttr namespace is not passed to opcode. I don't know why. 
Changing code to 
```
  this.opcode([isComponent ? 'componentAttr' : 'dynamicAttr', [name, namespace]], action);
```
seems to fix problem with my test cases, but I have no idea what else can break. Can someone take a look at it ?

rwjblue:
Nice job tracking this down! That fix definitely seems correct/fine to me.

Would you mind rebasing to resolve conflicts?

@chancancode / @wycats / @tomdale - would one of y'all mind reviewing?

rwjblue:
Thank you!

PR #871, opened 2018-12-20
Feat: Add option to turn on `ignoreStandalone` in HBS parser

zimmi88:
Adds new config option, `ignoreStandalone`, to `preprocess()` method.
This property is passed on to the handlebars parser (default: false).
When enabled, the handlebars parser preserves more whitespace. This,
in turn, creates ASTs that, when printed back into templates, results in
more developer-friendly code.

Also bumps `@types/handlebars` to latest, as this is necessary for glimmer
to recognize that `handlebars.parse()` does accept options.

### DIFF
--- a/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
+++ b/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
@@ -342,6 +342,7 @@ export interface PreprocessOptions {
   plugins?: {
     ast?: ASTPluginBuilder[];
   };
+  ignoreStandalone?: boolean;
 }
 
 export interface Syntax {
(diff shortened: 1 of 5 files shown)


stefanpenner:
@zimmi88 awesome this looks great, thanks for doing this + good tests. I've ping @tomdale to take a closer look.

zimmi88:
Hi @stefanpenner and @tomdale... happy new year! Was checking in to see if there were any updates/feedback regarding this PR, when convenient. Thanks!

packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
```diff
@@ -361,7 +362,8 @@ const syntax: Syntax = {
 };
 
 export function preprocess(html: string, options?: PreprocessOptions): AST.Program {
-  let ast = typeof html === 'object' ? html : handlebars.parse(html);
+  const ignoreStandalone = options ? options.ignoreStandalone : false;
+  let ast = typeof html === 'object' ? html : handlebars.parse(html, { ignoreStandalone });
```
rwjblue: Are there other options that are allowed for `handlebars.parse`? 

Should we try something like this instead:

```ts
export function preprocess(html: string, options?: PreprocessOptions): AST.Program {
  let parseOptions = options ? options.parseOptions : undefined;
  let ast = typeof html === 'object' ? html : handlebars.parse(html, parseOptions);
  // ...snip...
}
```

tomdale:
@zimmi88 Awesome, thanks for this PR! It sounds useful. I agree with @rwjblue's suggestion that we should probably think about an API for passing parser-specific options through to Handlebars, so we don't have to play whackamole with adding new options as they arise. Other than that the PR looks great.

PR #879, opened 2018-12-21
package.json: Sort contents in documented order

Turbo87:
see https://github.com/keithamus/sort-package-json/


tomdale:
Now that's what I call German engineering.

PR #891, opened 2018-12-23
Switch from Saucelabs to Browserstack

kategengler:
~Waiting on confirmation of Open Source account.~ ~BrowserStack has assured me the Open Source license will be activated when the logo credit to them is in the README (included in this PR).~ BrowserStack account is activated.

Of note: 
- Sadly, there is no way to indicate `latest - 1` as a version in BrowserStack, so these browser versions will need to be periodically updated.


rwjblue:
> Ember.js uses a failure-only reporter to reduce output, if desired we can extract that and use it here as well.

Yeah, I guess that means we should extract to a separate package that we can share as needed, right?

kategengler:
Yes, I can do that if that's desired. (vs. having to click to the plain-text log)

kategengler:
The extracted failure-only reporter works -- just need to figure out publishing it before this is ready.

rwjblue:
Eeck. Just made a yarn.lock conflict here, sorry.

kategengler:
Published `testem-failure-only-reporter` -- I think this PR is now ready?

PR #892, opened 2018-12-23
Failing test for forwarding attributes to yielded components

cibernox:
Extracted from https://github.com/emberjs/ember.js/pull/17146

@chancancode @rwjblue I hope this reproduction helps track this down.

In the test, as it stands right now, the `<Inner ...attributes>` component is not rendered at all. If you remove the `...attributes` it does render.

### DIFF
--- a/packages/@glimmer/test-helpers/lib/suites/components.ts
+++ b/packages/@glimmer/test-helpers/lib/suites/components.ts
@@ -138,26 +138,25 @@ export class BasicComponents extends RenderTest {
   }
 
   @test({
-    kind: 'basic',
+    kind: 'fragment',
   })
   'invoke a component forwarding attributes'() {
     this.registerComponent(
       'Glimmer',
       'InnerComponent',
-      '<p data-foo="target" ...attributes>Inner Component</p>'
+      '<p data-foo="definition" data-bar="definition" ...attributes>Inner Component</p>'
     );
     this.render(
       {
-        name: 'OuterComponent',
-        layout: '{{#with (component "InnerComponent") as |Inner|}}<Inner ...attributes />{{/with}}',
-        template: 'hello!',
-        args: {},
+        layout:
+          '{{#with (component "InnerComponent") as |Inner|}}<Inner data-foo="invocation" ...attributes />{{/with}}',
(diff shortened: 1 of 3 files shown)


packages/@glimmer/test-helpers/lib/suites/components.ts
```diff
@@ -137,6 +137,30 @@ export class BasicComponents extends RenderTest {
...
+      {
+        name: 'OuterComponent',
+        layout: '{{#with (component "InnerComponent") as |Inner|}}<Inner ...attributes />{{/with}}',
+        template: 'hello!',
+        args: {},
+        attributes: { title: 'Emberconf' },
+      },
+      {}
+    );
+    debugger;
```
cibernox: ooops

cibernox:
@chancancode did you have time to check what might be wrong with this?

rwjblue:
I've just pushed a fix, here is the explanation from the commit message:

```
Prior to the changes here, for every `AttrNode` we processed we
processed its value (handling literals, mustaches, etc). The value we
process is then consumed by whatever operation is represented (e.g.
`staticArg`, `dynamicArg`, `staticAttr`, `trustingAttr`, `dynamicAttr`).
Unfortunately, when we processed the `...attributes` `AttrNode` into an
`attrSplat` we did not also ensure to consume its value (which we don't
really care about). The result was that we had one too many literal
values on the `JavascriptCompiler`'s `values` array, meaning _all_ other
operations that use a value would be getting the _prior_ operations
value (e.g. an "off by one" issue in the values array).

The fix here is to ensure that `attrSplat` consumes the value that is
pushed to ensure things stay balanced.
```

cibernox:
@rwjblue great. Can we merge this? How do updates to the glimmer-vm roll out to ember versions? Could Ember 3.7.1 get this?

rwjblue:
Will need to be backported for inclusion in Ember 3.4 LTS, 3.7, and 3.8+ into the following branches:

* release-0-35 (this is used by ember-source@3.4)
* make a release-0-36 branch (for Ember 3.7)
* make a release-0-37 (for Ember 3.8+)

PR #893, opened 2018-12-24
yarn.lock: Deduplicate dependency versions

Turbo87:
see https://github.com/atlassian/yarn-deduplicate

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -97,14 +97,6 @@ abbrev@1:
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"
   integrity sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=
 
-accepts@~1.3.3:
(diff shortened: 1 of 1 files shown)


Turbo87:
I guess it's not a good sign that this just freezed on `ember build` now........ 😱

rwjblue:
No, definitely not great...

rwjblue:
Any insight on why this is getting stuck?

Turbo87:
no, haven't looked into it yet

Turbo87:
I've narrowed it down to `broccoli-babel-transpiler@6.4.0` being the first bad release

/cc @stefanpenner

Turbo87:
I've excluded the Babel dedupes from this PR for now to make it work, but this remains a problem...

rwjblue:
Also might try adding `JOBS=1` to ensure we aren't just doing parallelization when we didn't mean too...

Turbo87:
@rwjblue it might also be Babel itself updating because the package depending on Babel is updated 🤷‍♂️

PR #894, opened 2018-12-24
Update `broccoli-typescript-compiler` to v4.1.0 

Turbo87:
This PR attempts to update the `broccoli-typescript-compiler` dependency to v4.1.0, which internally updates TypeScript from v2.8.3 to v3.2.2. This results in a few different types of errors, most of which got resolved by #889 already. There are two types of errors remaining:

- use of `foreignobject` instead of `foreignObject` as a tag name string (fixed in this PR)
- broken type declarations in `@glimmer/syntax/lib/types/nodes.ts` (I've spent 2-3 hours trying to get this to work but still fail to understand how this is supposed to work)

I've given up trying to fix the `nodes.ts` issue,  which is why I opened this PR as WIP.

@krisselden according to git you wrote the type declarations in that file. would be great if you could have a look.

This PR will be needed to unblock https://github.com/emberjs/ember.js/pull/17298

**UPDATE:** After another 2 hours, I seem to have found a solution to the `nodes.ts` issues. Afterward, I also discovered other issues which are fixed by #896, #895 and two more commits in this PR.

/cc @tomdale @rwjblue

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -2,7 +2,7 @@
 
 const merge = require('broccoli-merge-trees');
 const funnel = require('broccoli-funnel');
-const { typescript } = require('broccoli-typescript-compiler');
+const typescript = require('broccoli-typescript-compiler').default;
 
 const buildTests = require('./build/broccoli/build-tests');
 const buildPackages = require('./build/broccoli/build-packages.js');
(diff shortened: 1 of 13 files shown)


rwjblue:
Needs a rebase and is also blocked on addressing my comment in https://github.com/glimmerjs/glimmer-vm/pull/895...

rwjblue:
Ok, I think we only need one last rebase?

rwjblue:
Added breaking label (since updating TS should be considered a breaking change).

Thom4sGoh:
https://github.com/glimmerjs/glimmer-vm/pull/894#event-2043168198

PR #883, opened 2018-12-22
Update Ember CLI to v3.6.0

Turbo87:
supports Broccoli 2 and other nice things 😉

PR #898, opened 2019-01-04
Update `ember-cli` to v3.6.1

Turbo87:
similar to https://github.com/glimmerjs/glimmer-vm/pull/883, but hopefully this time it passes, now that https://github.com/glimmerjs/glimmer-vm/pull/895 was merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "broccoli-source": "^1.1.0",
     "broccoli-typescript-compiler": "^4.1.0",
     "dag-map": "^2.0.2",
-    "ember-cli": "~3.4.4",
+    "ember-cli": "~3.6.1",
     "ember-cli-sauce": "^2.2.0",
     "emberjs-build": "^0.19.0",
     "eslint": "^5.10.0",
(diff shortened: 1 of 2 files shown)


Turbo87:
nope... still a problem 🙈

Turbo87:
looks like it's caused by
```
    at AssetPrinterSize.findFiles (/Users/tbieniek/Code/glimmer-vm/node_modules/ember-cli/lib/models/asset-size-printer.js:84:14)
```

Turbo87:
most likely caused by https://github.com/ember-cli/ember-cli/pull/8086/ as `walk-sync` does not handle circular symlinks in any way.

/cc @stefanpenner @rwjblue

rwjblue:
Interesting! Good catch @Turbo87!

rwjblue:
Would you mind reporting the circular symlink issue over on walk-sync?

rwjblue:
I think you can also work around here for now by suppressing the asset size printer output:

```
ember build -prod --suppress-sizes
```

PR #900, opened 2019-01-15
Ensure positionals are passed when prepareArgs is false

rwjblue:
Without this change, it is not possible to have a custom component manager that provides positional args. Specifically when `prepareArgs` is set we always bail out of `invokeStaticComponent` and use `invokeComponent` instead, but when `prepareArgs` is `false` we _always_ compile args with positional of `null`.

TODO:

- [x] add tests....


rwjblue:
Updated with a test that failed against master (8a1c8bc70), and now passes.

rwjblue:
Retargeting to release-0-38 (will bring back onto master separately).

PR #904, opened 2019-01-19
Internals improvements

wycats:
This is a fairly significant, based-off-master and all-tests-passing restructuring of the internals to bring them more up-to-date with modern practices.

- updates to the latest version of TypeScript and various dependencies
- changes the internals to be more suitable for embeddings with fewer hacks
- incorporates common behavior between Ember and Glimmer.js into the VM
- eliminate Environment superclass in favor of much smaller delegate APIs
- moves auto-track primitives into @glimmer/reference
- eliminates dependency of `UpdatableReference` on @glimmer/object-reference (an abandoned implementation of legacy Ember functionality)
- eliminates the test-helpers cycle
- generates opcode metadata and stack assertions from opcodes.toml


wycats:
It's probably reasonably to hold off on merging this into master until Glimmer.js is integrated.

There is also some dependence on symbols in this implementation, which aren't available in IE11. This could be eliminated if necessary.

PR #910, opened 2019-02-04
Expose meta data to AST plugins environment

wondersloth:
- Spread `ModuleLocator` attributes into `meta` property in `PreprocessOptions`.
  This is similar to `PrecompileOptions` existing use of `meta` attribute.
- Modify interface to allow for optional attribute to extend correctly.
- Wrote test to validate passing of meta data, that gets exposed in the env
  for AST plugins.


wondersloth:
Alternatively we can expose the `locator:ModuleLocator` on the `meta` object, though I was intending to align to existing convention.

PR #911, opened 2019-02-06
Update Guides to include more complete steps

SinS3i:
### Why
The embedding guides were a great start to explaining how to setup and run a simple embedding environment, but they still required a bit of manual digging to determine where certain things were coming from, or small corrections to elaborate on helper functions that were in the glimmerjs/glimmer-embedding-tutorial repository.

### What
Added information I found useful in ramping up on the aforementioned APIs (mostly imports, a few helper methods)


tomdale:
Thank you!

PR #918, opened 2019-02-11
[FEATURE | BREAKING] Change semantics of in-element

chadhietala:
This implements the [clearing semantics](https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md#small-proposed-changes) of `in-element` as it pertains to https://github.com/emberjs/rfcs/pull/287. Instead of appending we need to clear the contents of the destination element unless `insertBefore` was explicitly set to `null`.

This is required work for https://github.com/emberjs/rfcs/pull/418

### DIFF
--- a/packages/@glimmer/node/lib/serialize-builder.ts
+++ b/packages/@glimmer/node/lib/serialize-builder.ts
@@ -96,13 +96,13 @@ class SerializeBuilder extends NewElementBuilder implements ElementBuilder {
   pushRemoteElement(
     element: SimpleElement,
     cursorId: string,
-    nextSibling: Option<SimpleNode> = null
+    insertBefore: Option<null> = null
   ): Option<RemoteLiveBlock> {
     let { dom } = this;
     let script = dom.createElement('script');
     script.setAttribute('glmr', cursorId);
-    dom.insertBefore(element, script, nextSibling);
-    return super.pushRemoteElement(element, cursorId, nextSibling);
+    dom.insertBefore(element, script, insertBefore);
+    return super.pushRemoteElement(element, cursorId, insertBefore);
   }
 }
 
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
@@ -7,9 +7,8 @@ import {
   isConst,
   isConstTag,
 } from '@glimmer/reference';
-import { Option } from '@glimmer/util';
-import { check, CheckString, CheckElement, CheckNode, CheckOption } from '@glimmer/debug';
-import { Op } from '@glimmer/interfaces';
+import { check, CheckString, CheckElement } from '@glimmer/debug';
+import { Op, Option } from '@glimmer/interfaces';
 import { $t0 } from '@glimmer/vm';
 import {
   ModifierDefinition,
@@ -22,7 +21,7 @@ import { Assert } from './vm';
 import { DynamicAttribute } from '../../vm/attributes/dynamic';
 import { CheckReference, CheckArguments, CheckOperations } from './-debug-strip';
 import { CONSTANTS } from '../../symbols';
-import { SimpleElement, SimpleNode } from '@simple-dom/interface';
+import { SimpleElement } from '@simple-dom/interface';
 
 APPEND_OPCODES.add(Op.Text, (vm, { op1: text }) => {
   vm.elements().appendText(vm[CONSTANTS].getString(text));
@@ -43,11 +42,10 @@ APPEND_OPCODES.add(Op.OpenDynamicElement, vm => {
 
 APPEND_OPCODES.add(Op.PushRemoteElement, vm => {
   let elementRef = check(vm.stack.pop(), CheckReference);
-  let nextSiblingRef = check(vm.stack.pop(), CheckReference);
+  let insertBeforeRef = check(vm.stack.pop(), CheckReference);
   let guidRef = check(vm.stack.pop(), CheckReference);
 
   let element: SimpleElement;
-  let nextSibling: Option<SimpleNode>;
   let guid = guidRef.value() as string;
 
   if (isConst(elementRef)) {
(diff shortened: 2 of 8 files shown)


packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -43,11 +42,10 @@ APPEND_OPCODES.add(Op.OpenDynamicElement, vm => {
 
 APPEND_OPCODES.add(Op.PushRemoteElement, vm => {
   let elementRef = check(vm.stack.pop(), CheckReference);
-  let nextSiblingRef = check(vm.stack.pop(), CheckReference);
+  let _insertBefore = check(vm.stack.pop(), CheckReference);
```
rwjblue:
```suggestion
  let insertBeforeRef = check(vm.stack.pop(), CheckReference);
```

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -58,15 +56,9 @@ APPEND_OPCODES.add(Op.PushRemoteElement, vm => {
...
   }
 
-  if (isConst(nextSiblingRef)) {
-    nextSibling = check(nextSiblingRef.value(), CheckOption(CheckNode));
-  } else {
-    let cache = new ReferenceCache(nextSiblingRef as Reference<Option<SimpleNode>>);
-    nextSibling = check(cache.peek(), CheckOption(CheckNode));
-    vm.updateWith(new Assert(cache));
-  }
+  let insertBefore = _insertBefore.value() as Option<null>;
```
rwjblue:
```suggestion
  let insertBefore = insertBeforeRef.value() as Option<null>;
```

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -379,15 +379,21 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
...
+        while (element.lastChild !== marker) {
+          element.removeChild(element.lastChild!);
+        }
+      }
+
       let currentCursor = this.currentCursor;
       let candidate = currentCursor!.candidate;
 
-      this.pushElement(element, nextSibling);
+      this.pushElement(element, null);
```
rwjblue: I think we should pass through `insertBefore` even though we know _currently_ its always `null`...

```suggestion
      this.pushElement(element, insertBefore);
```

packages/@glimmer/node/lib/serialize-builder.ts
```diff
@@ -96,13 +96,13 @@ class SerializeBuilder extends NewElementBuilder implements ElementBuilder {
   pushRemoteElement(
     element: SimpleElement,
     cursorId: string,
-    nextSibling: Option<SimpleNode> = null
+    _insertBefore: Option<null>
```
rwjblue: I'd prefer to continue passing `insertBefore` around (even though we assert on values other than `null` or `false`).

```suggestion
    insertBefore: Option<null>
```

packages/@glimmer/node/lib/serialize-builder.ts
```diff
@@ -96,13 +96,13 @@ class SerializeBuilder extends NewElementBuilder implements ElementBuilder {
...
   ): Option<RemoteLiveBlock> {
     let { dom } = this;
     let script = dom.createElement('script');
     script.setAttribute('glmr', cursorId);
-    dom.insertBefore(element, script, nextSibling);
-    return super.pushRemoteElement(element, cursorId, nextSibling);
+    dom.insertBefore(element, script, null);
```
rwjblue:
```suggestion
    dom.insertBefore(element, script, insertBefore);
```

packages/@glimmer/node/lib/serialize-builder.ts
```diff
@@ -96,13 +96,13 @@ class SerializeBuilder extends NewElementBuilder implements ElementBuilder {
...
+    return super.pushRemoteElement(element, cursorId, null);
```
rwjblue:
```suggestion
    return super.pushRemoteElement(element, cursorId, insertBefore);
```

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -379,15 +379,21 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
   __pushRemoteElement(
     element: SimpleElement,
     cursorId: string,
-    nextSibling: Option<SimpleNode> = null
+    insertBefore: Option<null>
   ): Option<RemoteLiveBlock> {
     let marker = this.getMarker(element as HTMLElement, cursorId);
 
     if (marker.parentNode === element) {
+      if (insertBefore !== null) {
```
rwjblue: It's my understanding that we want `null` to mean append, `undefined` to mean replace, and any other value to mean "before this thing" (though this option isn't implemented yet). If thats the case, the current logic here is going to make fixing that third scenario pretty hard. How about changing the logic of the conditional to something like:

```suggestion
      if (insertBefore === undefined) {
```

packages/@glimmer/runtime/lib/vm/element-builder.ts
```diff
@@ -205,18 +205,26 @@ export class NewElementBuilder implements ElementBuilder {
...
   __pushRemoteElement(
     element: SimpleElement,
     _guid: string,
-    nextSibling: Option<SimpleNode>
+    insertBefore: Option<null>
   ): Option<RemoteLiveBlock> {
-    this.pushElement(element, nextSibling);
+    this.pushElement(element, insertBefore);
+
+    if (insertBefore !== null) {
```
rwjblue:
```suggestion
    if (insertBefore === undefined) {
```

chancancode:
I think I have a few recommendations:

* Need tests for the "replace" case (i.e. when `insertBefore` is not specified).
* `Option<null>` doesn't make sense, since it just expands into `null | null`.
* I think it would make sense to keep the internal `pushRemoteElement` API taking an element and a `nextSibling`. This is basically the "DOM cursor" concept that is used throughout the element builder code. Changing this would make it inconsistent with e.g. `pushElement`.
* Instead, in the clearing case, I think it would make more sense for the caller to clear the target element (may require adding APIs to the element build or somewhere else, not sure), then call `pushRemoteElement` with the cleared element and `null` nextSibling`.

chadhietala:
I added the test. We can do a refactor in the future to break apart the clearing out the target. I had some questions about the Enum map for operations since it has dups in it.

PR #919, opened 2019-02-15
fix: security audit handlebars update

knownasilya:
https://www.npmjs.com/advisories/755


knownasilya:
@rwjblue thanks added

knownasilya:
Having trouble with yarn in syntax package:

```
error An unexpected error occurred: "https://registry.yarnpkg.com/@glimmer%2flocal-debug-flags: Not found".
```

rwjblue:
Thank you @knownasilya!

PR #921, opened 2019-02-24
ComponentManager types - fix typeParam shadowing

mike-north:
defining a local type parameter `ComponentInstanceState` shadows the imported symbol of the same name, resulting in a self-referential default type. Looks like this has been fixed in several places, but this one was missed.


stefanpenner:
The test failure was unrelated, specifically it was a browser timeout. Giving it a kick, let's see if it self resolves.

PR #922, opened 2019-02-27
Use type generic for `scope()` function

pzuraq:
We're getting errors from TS downstream when trying to use `@glimmer/component` in Ember apps with TS, this change fixes them. I _believe_ it makes sense, based on my understanding of TS and generics, for scope to be the generic defined in the abstract class, but if I'm wrong I'm happy to make the correct change to fix (and learn more about how TS works 😄)


josemarluedke:
I have seen TS errors related to this. Here are they: https://gist.github.com/josemarluedke/195071140e82ce028ce6abb8745a9ca3

Not sure if this change will fix it all for my case.

PR #923, opened 2019-03-10
RFC #435: Components forward element modifiers with splattributes

cibernox:


cibernox:
@chancancode I don't think I can push this forward without some assistant. The internals of the glimmer-vm are a bit daunting without guidance. Will you have some cycles to help me put this on track?

chancancode:
@cibernox yes! are you able to add the tests? does it work if you just remove these lines? https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/compiler/lib/javascript-compiler.ts#L320-L323

cibernox:
@chancancode if you check this PR closely you can see I had already removed those lines.

chancancode:
@cibernox did it not work? looking at the test, it seems like it's testing the wrong thing (the invocation passed a positional argument, but the test is looking for a named argument)

cibernox:
@chancancode you where right, the test was wrong, but in either case the `didInsertElement` of the modifier is being called at all.

chancancode:
:+1: thanks for checking

Can you try to dump the wire format JSON output between these on your branch?

```
<div {{foo "bar" baz=this.bat}}>element</div>
```

```
<Woot {{foo "bar" baz=this.bat}}>component</Woot>
```

cibernox:
`<div {{foo "bar" baz=this.bat}}>element</div>` generates: 
```json
{
  "parsedLayout": {
    "id": "client-0",
    "block": {
      "symbols": [],
      "statements": [
        [7, "div"],
        [9],
        [0, "element"],
        [3, "foo", ["bar"], [["baz"], [[24, 0, ["bat"]]]]],
        [10]
      ],
      "hasEval": false
    },
    "referrer": {
      "kind": "unknown",
      "meta": {},
      "module": "some/template",
      "name": "default"
    }
  },
  "layout": null,
  "partial": null,
  "wrappedLayout": null,
  "symbols": [],
  "hasEval": false,
  "referrer": {
    "kind": "unknown",
    "meta": {},
    "module": "some/template",
    "name": "default"
  },
  "id": "client-0"
}
```

`<Woot {{foo "bar" baz=this.bat}}>component</Woot>` generates:
```json
{
  "parsedLayout": {
    "id": "client-0",
    "block": {
      "symbols": [],
      "statements": [
        [
          5,
          "Woot",
          [],
          [
            [],
            []
          ],
          [
            [
              "default"
            ],
            [
              {
                "statements": [
                  [
                    0,
                    "component"
                  ]
                ],
                "parameters": []
              }
            ]
          ]
        ]
      ],
      "hasEval": false
    },
    "referrer": {
      "kind": "unknown",
      "meta": {},
      "module": "some/template",
      "name": "default"
    }
  },
  "layout": null,
  "partial": null,
  "wrappedLayout": null,
  "symbols": [],
  "hasEval": false,
  "referrer": {
    "kind": "unknown",
    "meta": {},
    "module": "some/template",
    "name": "default"
  },
  "id": "client-0"
}
```

As you can see the modifiers information is lost along the way. I debugged this a few weeks ago.

Seems that the information of the opcode generated in https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/compiler/lib/javascript-compiler.ts#L324 does not include any reference to the modifiers. The opcodes are completely different. I attempted to refactor this to make `endComponent` return the modifiers among the `attrs`, but it found myself out of my depth.

chancancode:
@cibernox I see.

The general architecture here is...

```
Parser -> [[AST]] -> TemplateCompiler -> [[Intermediate]] -> JavaScriptCompiler -> [[WireFormat]]
```

We know that the modifiers made it to the AST. (That's the one you get on http://astexplorer.net or deal with when writing AST transforms.) But we also know that it didn't make it into the WireFormat.

I can see there are a few problems:

1. The `TemplateCompiler` drops the modifiers for components: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/compiler/lib/template-compiler.ts#L134-L149

    This needs to be fixed by hosting the for loop above the component closing opcodes.

2. The second problem is the `ComponentBlock` (a helper class in the `JavaScriptCompiler`) does not handle modifiers: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/compiler/lib/javascript-compiler.ts#L100

    This needs to be fixed by adding a branch that pushes modifier statements into the `attributes` array (and changing the types to allow it)

Once you fixed those things, I _think_ you will start seeing the modifiers in the wire format output. At this point, maybe things will Just Work™, or maybe it will require some more minor adjustments in the runtime side, but it'll be a great start.

cibernox:
That's is more or less what I tried last time, but I'll try again and share any progress

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -137,13 +137,14 @@ export default class TemplateCompiler {
...
     } else if (isComponent(action)) {
-      this.opcode(['closeComponent', action], action);
-    } else if (action.modifiers.length > 0) {
       for (let i = 0; i < action.modifiers.length; i++) {
         this.modifier([action.modifiers[i]]);
       }
-      this.opcode(['closeElement', action], action);
+      this.opcode(['closeComponent', action], action);
     } else {
+      for (let i = 0; i < action.modifiers.length; i++) {
```
cibernox: @chancancode I did this, but it wasn't clear to me if the right approach was to add a modifier opcode before closing the component (like it happens when closing regular Elements) or the modifiers when applied to components should just be part of the arguments of the `openComponent`.

chancancode: @cibernox This should apply to both dynamic (`<this.foo ... />`) and regular components (`<Foo />`). Basically everything other than named blocks. So it probably makes sense to structure it like this:

```js
  closeElement([action]: [AST.ElementNode]) {
    if (isNamedBlock(action)) {
      // assert no attributes and modifiers here probably, unless that's already handled elsewhere
      this.opcode(['closeNamedBlock', action]);
    } else {
      for (let i = 0; i < action.modifiers.length; i++) { ... }

      if (isDynamicComponent(action)) {
        ...
      } else if (...) {
      } else if (...) {
      } else {
        unreachable();
      }
    }
  }
```

chancancode: as far as the factoring, the idea is that `TemplateCompiler` takes the AST (nested tree-structure), "flattens" it into an array of operations that needs to happen, and `JavaScriptCompiler` (terrible legacy name at this point) takes the falt list of operations and emit a the wire-format (mostly flat, but some "nesting" from having blocks).

So yes, this should emit the modifier opcodes before the close opcode, and the next stage can resemble them into the attributes block.

cibernox:
Thanks, I'll continue with this in a few hours

cibernox:
@chancancode I manager to add the modifiers before the `flushElement` instruction, so they are added to the list of attributes to be forwarded to the inner div.

The test seemed to pass but then I realized that inside the modifier, `this.element` was not the div I expected it to be. Can you take a look at this?

packages/@glimmer/compiler/lib/javascript-compiler.ts
```diff
@@ -243,8 +241,11 @@ export default class JavaScriptCompiler
   modifier(name: string) {
     let params = this.popValue<Params>();
     let hash = this.popValue<Hash>();
-
-    this.push([SexpOpcodes.Modifier, name, params, hash]);
+    if (this.currentComponent) {
+      this.currentComponent.push([SexpOpcodes.Modifier, name, params, hash]);
+    } else {
+      this.push([SexpOpcodes.Modifier, name, params, hash]);
+    }
```
chancancode: This shouldn't be necessary: `this.push` goes through `this.currentBlock.push`, which should be the same as `this.currentComponent`. However, I think [this list](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/compiler/lib/javascript-compiler.ts#L91-L101) need to be updated to account for modifiers (push it into `this.attributes`).

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -127,24 +127,25 @@ export default class TemplateCompiler {
         this.attribute([typeAttr], hasSplat || actionIsComponent);
       }
 
+      for (let i = 0; i < action.modifiers.length; i++) {
+        this.modifier([action.modifiers[i]]);
+      }
```
chancancode: :+1:

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -127,24 +127,25 @@ export default class TemplateCompiler {
...
-    } else if (isComponent(action)) {
-      this.opcode(['closeComponent', action], action);
-    } else if (action.modifiers.length > 0) {
-      for (let i = 0; i < action.modifiers.length; i++) {
-        this.modifier([action.modifiers[i]]);
-      }
-      this.opcode(['closeElement', action], action);
     } else {
-      this.opcode(['closeElement', action], action);
+      if (isDynamicComponent(action)) {
```
chancancode: This can be restored to the `} else if(...) {` form, now that we aren't sharing any code between these branches

packages/@glimmer/wire-format/index.ts
```diff
@@ -15,7 +15,8 @@ export function isAttribute(val: Statement): val is Statements.Attribute {
     val[0] === SexpOpcodes.StaticAttr ||
     val[0] === SexpOpcodes.DynamicAttr ||
     val[0] === SexpOpcodes.TrustingDynamicAttr ||
-    val[0] === SexpOpcodes.ComponentAttr
+    val[0] === SexpOpcodes.ComponentAttr ||
+    val[0] === SexpOpcodes.Modifier
```
chancancode: This list need to be updated: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/interfaces/lib/compile/wire-format.d.ts#L209-L214, otherwise it will cause a type error when you try to push it.

chancancode: By the way, I think you should just include the check for `AttrSplat` in here and and delete the `isAttrSplat` function. I don't know why it is handled separately, and as far as I can tell, it doesn't have to be, and probably shouldn't.

cibernox: Done!

packages/@glimmer/compiler/lib/template-compiler.ts
cibernox: Done

packages/@glimmer/compiler/lib/javascript-compiler.ts
cibernox: Reverted to a regular push

packages/@glimmer/wire-format/index.ts
chancancode: @cibernox still need to update the wire-format types 😄

cibernox: I just pushed that

chancancode:
@cibernox looks great so far.

Assuming the wire format JSON looks correct to you, the next step is to fix the runtime side of things.

This is where having TypeScript working for you locally would be quite helpful. After changing the `Attribute` type to include modifiers, try running a full typescript build to see if it flagged any new type errors – these are the spots we will need to update to handle modifiers.

Separately, since we changed modifiers attachment timing, from _after_ to _during_ construction (i.e. from after to before flushing the element), we will at least need to update [here](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts#L96) to use the [`constructing`](https://github.com/glimmerjs/glimmer-vm/blob/a15a7c86ae50a882f3f395a394cdaff6211ade30/packages/%40glimmer/interfaces/lib/dom/attributes.d.ts#L89) property instead of the [`element`](https://github.com/glimmerjs/glimmer-vm/blob/a15a7c86ae50a882f3f395a394cdaff6211ade30/packages/%40glimmer/interfaces/lib/dom/attributes.d.ts#LL90) property.

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -93,9 +93,15 @@ APPEND_OPCODES.add(Op.Modifier, (vm, { op1: handle }) => {
   let { manager, state } = vm.runtime.resolver.resolve<ModifierDefinition>(handle);
   let stack = vm.stack;
   let args = check(stack.pop(), CheckArguments);
-  let { element, updateOperations } = vm.elements();
+  let { constructing, updateOperations } = vm.elements();
   let dynamicScope = vm.dynamicScope();
-  let modifier = manager.create(element, state, args, dynamicScope, updateOperations);
+  let modifier = manager.create(
+    <SimpleElement>constructing,
```
chancancode: use `expect` (https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/util/lib/platform-utils.ts#L15-L18) instead

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,27 @@ class ModifierTests extends RenderTest {
...
-
-    assert.throws(() => {
-      this.render('<Foo (bar foo="foo") />');
-    }, 'Compile Error: Element modifiers are not allowed in components');
+  'work on component invocations'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: unknown[], namedArgs: Dict<string>) {
```
chancancode: should probably be `Dict<unknown>`

chancancode:
For the ordering tests (currently failing), I think you can just change the ordering in the tests. We will figure out how to preserve Ember's public API ordering when we integrate.

chancancode:
Seems like other than the ordering tests, everything is passing. Let's make sure we have test for the following cases:

1. Passing "bound" things to modifiers (to test the update hook)
2. Passing `this`/and or things derived from `this` to modifiers (`<div {{foo this this.bar}} />`) to make sure we are calling the attributes block with the right context
3. Passing local variables to modifiers (`{{#let ... |foo|}}<div {{bar foo}} />{{/let}}`) to make sure we are calling the attributes block with the right scope
4. Calling `...attributes` on multiple elements (it should instantiate multiple modifiers)
5. Deeply nested versions (multiple let scopes in between, multiple components/...attribute layers etc) of the above combinations, to make sure the scope/context is right

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,27 @@ class ModifierTests extends RenderTest {
...
-
-    assert.throws(() => {
-      this.render('<Foo (bar foo="foo") />');
-    }, 'Compile Error: Element modifiers are not allowed in components');
+  'work on component invocations'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: Dict<unknown>, namedArgs: Dict<string>) {
```
chancancode: sorry, I should be more clear, I meant `params: unknown[], namedArgs: Dict<unknown>`

cibernox:
@chancancode I added tests for all the use cases, including components that receive modifiers and forward them to inner components with `...attributes` so they are applied several levers deep.

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+  'modifiers on components are forwarded to all the elements receiving the splattributes'(
+    assert: Assert
+  ) {
+    let elements: SimpleElement[] = [];
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: unknown[], namedArgs: Dict<unknown>) {
+        assert.deepEqual(params, ['something']);
+        assert.deepEqual(namedArgs, { foo: 'else' });
+        if (this.element) {
+          elements.push(this.element);
```
chancancode: Let's push the element's id here...

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    }
+    this.registerComponent(
+      'Glimmer',
+      'TheFoo',
+      '<div id="inner-one" ...attributes>Foo</div><div id="inner-two" ...attributes>Bar</div>'
+    );
+    this.registerModifier('bar', Bar);
+    this.render('<TheFoo {{bar "something" foo="else"}}/>');
+    assert.equal(
+      elements.length,
```
chancancode: ...then deepEqual `['inner-one', 'inner-two']` here

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+  'modifiers on components accept bound arguments'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: unknown[], namedArgs: Dict<unknown>) {
+        modifierParams = params;
+        modifierNamedArgs = namedArgs;
+        modifiedElement = this.element;
+      }
```
chancancode: Let's implement `didUpdate` and test that the values change

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+  ) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: unknown[], namedArgs: Dict<unknown>) {
+        modifierParams = params;
+        modifierNamedArgs = namedArgs;
+        modifiedElement = this.element;
+      }
```
chancancode: Let's implement `didUpdate` and check that the values change

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+  'modifiers on components can be received and forwarded to inner components'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
+    class Bar extends AbstractInsertable {
+      didInsertElement(params: unknown[], namedArgs: Dict<unknown>) {
+        modifierParams = params;
+        modifierNamedArgs = namedArgs;
+        modifiedElement = this.element;
+      }
```
chancancode: Let's implement `didUpdate` and check that the values change

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    }
+    this.registerComponent(
+      'Glimmer',
+      'TheInner',
+      '<div id="inner-div" ...attributes>{{yield}}</div>'
+    );
+    this.registerComponent('Glimmer', 'TheFoo', '<TheInner ...attributes>Hello</TheInner>');
```
chancancode: Let's add another element in this middle layer and splat it there too

packages/@glimmer/compiler/lib/javascript-compiler.ts
```diff
@@ -2,7 +2,7 @@ import { assert } from '@glimmer/util';
 import { Stack, DictSet, Option, expect } from '@glimmer/util';
 import { AST } from '@glimmer/syntax';
 import { CompileOptions } from './template-compiler';
-import { isFlushElement, isArgument, isAttribute, isAttrSplat } from '@glimmer/wire-format';
+import { isFlushElement, isArgument, isAttribute } from '@glimmer/wire-format';
```
chancancode: while you are here, let's alphabetize the imports on this line

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
-      this.render('<Foo {{bar foo="foo"}} />');
-    }, 'Compile Error: Element modifiers are not allowed in components');
-
-    assert.throws(() => {
-      this.render('<Foo (bar foo="foo") />');
-    }, 'Compile Error: Element modifiers are not allowed in components');
+  'modifiers on components are forwarded to a single element receiving the splattributes'(
+    assert: Assert
+  ) {
+    let modifierParams = null;
```
chancancode: `let modifierParams: Option<unknown[]> = null;`

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    let modifierNamedArgs = null;
```
chancancode: `let modifierNamedArgs: Option<Dict <unknown>> = null;`

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    let modifiedElement: SimpleElement | undefined;
```
chancancode: `let modifiedElement: Option<SimpleElement> = null;`

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+      2,
+      'The modifier has been instantiated twice, once for each element with splattributes'
+    );
+  }
+
+  @test
+  'modifiers on components accept bound arguments'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
```
chancancode: types: see above

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    );
+  }
+
+  @test
+  'modifiers on components accept `this` in both positional params and named arguments'(
+    assert: Assert
+  ) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
```
chancancode: types: see above

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+    );
+  }
+
+  @test
+  'modifiers on components accept local variables in both positional params and named arguments'(
+    assert: Assert
+  ) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
```
chancancode: types: see above

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +82,180 @@ class ModifierTests extends RenderTest {
...
+      'inner-div',
+      'Modifier is called on the element receiving the splattributes'
+    );
+  }
+
+  @test
+  'modifiers on components can be received and forwarded to inner components'(assert: Assert) {
+    let modifierParams = null;
+    let modifierNamedArgs = null;
+    let modifiedElement: SimpleElement | undefined;
```
chancancode: types: see above

cibernox:
@chancancode I applied all your feedback except `let modifiedElement: Option<SimpleElement> = null;` because then I couldn't find a way to get the ID of the element without typescript complaining about `Property 'getAttribute' does not exist on type 'never'`

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -82,16 +83,223 @@ class ModifierTests extends RenderTest {
...
+      'inner-div',
+      'Modifier is called on the element receiving the splattributes'
+    );
+  }
+
+  @test
+  'modifiers on components are forwarded to all the elements receiving the splattributes'(
+    assert: Assert
+  ) {
+    let elementIds: any[] = [];
```
chancancode: This can now be `string[]` (or maybe `Option<string>[]` – `getAttribute` returns `Option<string>`, which we can either just push into the array like you are doing now, or check for null first)

packages/@glimmer/integration-tests/test/modifiers-test.ts
chancancode: Looks like this one is still missing

chancancode:
@cibernox 👍let's just leave that one as-is then! one more minor type suggestion and one missing update test, otherwise LGTM!

chancancode:
By the way, since ember is still on v0.37.1, the quickest way to get this feature into ember is probably backporting it into the `release-0-37` branch (after this is merged into master/0.39). Hopefully that should be pretty straightforward, since this end up being a pretty small diff.

Optionally, it maybe a good idea to backport it to `release-0-38` as well. I don't know of anyone that uses it, but if it is easy, it may avoid regression if we end up having to first upgrade ember to 0.38 for some reason.

chancancode:
@cibernox I looked at the commits and figured it shouldn't be that hard; so I went ahead and upgraded ember to 0.38. If https://github.com/emberjs/ember.js/pull/17824 is merged, then we can backport it to release-0-38 only.

cibernox:
@chancancode updated and squashed, ready to merge!

chancancode:
@cibernox thank you so much 🙌🏼

PR #924, opened 2019-03-25
Add ability to pass in dynamic scope to render component apis

chiragpat:
Currently only the renderMain API's have the ability to take in a DynamicScope as an argument. This PR extends that ability to the renderComponent API's as well.


tomdale:
@chiragpat Looks good, thank you! I made a small change to your original PR, to make sure that the signatures for the `initial` and `empty` static methods were consistent in taking the `dynamicScope` in the options hash.

PR #931, opened 2019-04-10
Modifying in-element semantics to support non-null insertBefore elements

chiragpat:
- This is a slight modification of the changes introduced in PR #918, to continue to allow passing non null values to the insertBefore param of the in-element helper

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/in-element.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/in-element.ts
@@ -131,17 +131,38 @@ export class InElementSuite extends RenderTest {
   }
 
   @test
-  '`insertBefore` can only be null'() {
+  'With insertBefore'() {
     let externalElement = this.delegate.createElement('div');
-    let before = this.delegate.createElement('div');
+    replaceHTML(externalElement, '<b>Hello</b><em>there!</em>');
 
-    this.assert.throws(() => {
-      this.render('{{#in-element externalElement insertBefore=before}}[{{foo}}]{{/in-element}}', {
-        externalElement,
-        before,
-        foo: 'Yippie!',
-      });
-    }, /insertBefore only takes `null` as an argument/);
+    this.render(
+      stripTight`{{#in-element externalElement insertBefore=insertBefore}}[{{foo}}]{{/in-element}}`,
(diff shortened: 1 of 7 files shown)


rwjblue:
For context, the idea here is to allow Glimmer.js to continue using the `insertBefore` value in its [root template](https://github.com/glimmerjs/glimmer.js/blob/10f44093781b9fd65dd996484429809cd6ba0681/packages/%40glimmer/application/src/templates/main.hbs#L2). When Ember updates, it will add an AST transform to assert on any value other than `NullLiteral` for `insertBefore` (and therefore still satisfy the original intent of the RFC).

chancancode:
We generally don't mix `undefined` and `null` – most (if not all) of the `Maybe` in here should be `Option`

chiragpat:
Sure I can swap it back to Option. The main reason I did it was in dom.ts I don't initialize the insertBefore Option if the underlying ref value is undefined so that we can support replacing all the html in the container if no insertBefore was passed into the helper. 

Any recommendation on fixing the type error I get because of that?

rwjblue:
> We generally don't mix undefined and null – most (if not all) of the Maybe in here should be Option

@chancancode - I agree in general, but the specific feature we are implementing intentionally allows `null` and `undefined` as distinct values with different behaviors.

chancancode:
How was the code managing that before? Or was this introducing two features (setting it to undefined, and setting it to a child node)? If so, does it need another test?

Very generally speaking, I would prefer to handle that difference relatively early on, clear the element, and just pass null from there on, but that may be incidental difficult to do. I think we mentioned it last time too but I think we decided it’s a non-blocker.

chancancode:
Anyway, I am traveling tomorrow so feel free to merge it as you see fit 😬

chiragpat:
> How was the code managing that before? Or was this introducing two features (setting it to undefined, and setting it to a child node)? If so, does it need another test?

Before the code was casting it to an Option<null> even if it was undefined, later down in the pushRemoteElement method it would actually check if this option was undefined. Setting it to undefined was introduced in the original PR that I have linked to. There is a test already for null, undefined and an actual value. The undefined test has existed from before.

chancancode:
Oh uh, that’s just a bug, thank you for fixing it. It seems good to fix the types to reflect reality (ie Maybe 👍). We can do the refactor I suggested in another pass.

PR #933, opened 2019-04-18
glimmer-syntax: failing test for printing if/block statements and leading/trailing whitespace

fivetanley:
The printer does not seem to preserve whitespace whenever an `{{#if}}` / block statement is involved. For example:

Expected:

```handlebars
<div>
  {{#if foo}}
    <div>hi</div>
  {{/if}}
</div>
```

Got:

```handlebars
<div>
{{#if foo}}  <div>hi</div>
{{/if}}</div>
```


fivetanley:
went with using https://github.com/ember-template-lint/ember-template-recast#print instead!

PR #938, opened 2019-04-30
Add `codemod` mode for @glimmer/syntax preparse.

rwjblue:
Currently, setting `mode` to `'codemod'` will:

* Disable `entity` parsing in simple-html-tokenizer (ensures that parsing + printing is not lossy WRT entities)
* Enable `ignoreStandalone` mode in handlebars parser (ensures that standalone whitespace after a block opening or closing is not stripped).

In the future we can do other things to make the codemod world even better, but this is a really good first step (IMHO).

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/print.ts
+++ b/packages/@glimmer/syntax/lib/generation/print.ts
@@ -7,10 +7,63 @@ function unreachable(): never {
   throw new Error('unreachable');
 }
 
-export default function build(ast: AST.Node): string {
+interface PrinterOptions {
+  entityEncoding: 'transformed' | 'raw';
+}
+
+export default function build(
+  ast: AST.Node,
+  options: PrinterOptions = { entityEncoding: 'transformed' }
+): string {
   if (!ast) {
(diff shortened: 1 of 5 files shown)


Turbo87:
looks like #941 caused a conflict here, sorry :(

Turbo87:
I rebased it and fixed the conflict :)

Turbo87:
restarted CI due to the browserstack timeout. it's 🍏 now.

nabeelz6:
@rwjblue I commented on the prettier repo but it seems that HTML entity numbers are still being converted: https://github.com/jgwhite/prettier/issues/1#issuecomment-749608482

Is this something @glimmer/syntax could address or we'd have to update prettier for?

jpike88:
@nabeelz6 it's definitely a problem with the glimmer parser

nabeelz6:
> @nabeelz6 it's definitely a problem with the glimmer parser

@jpike88 Thanks for fixing in https://github.com/glimmerjs/glimmer-vm/pull/1284

PR #939, opened 2019-04-30
Update deps in interfaces and util

### LINKED ISSUE #935: Missing dependencies when relying on @glimmer/syntax
Downstream issue: https://github.com/ember-template-lint/ember-template-recast/issues/20

Somewhere between 0.38.1 and 0.39.3, `@glimmer/interfaces` both...

- Dropped its dependencies on `@glimmer/wire-format` and `@simple-dom/interface` and
- Started relying on `@glimmer/reference` in its code, without adding `@glimmer/reference` as a dependency.

It also looks like `@glimmer/utils` references `@simple-dom/interface` in its code, but has it listed as a devDependency.

These issues collectively cause build errors when using a library that depends on `@glimmer/syntax` specifically (such as `ember-template-recast`).

Proposed fix is to update the `package.json` for these packages to rely on these dependencies. If these dependencies were intentionally dropped, or if there is an alternate recommended solution for end users, please let me know. Thanks!

zimmi88:
- Fix issue with missing deps for packages that rely on @glimmer/syntax.

Fixes #935

### DIFF
--- a/packages/@glimmer/interfaces/package.json
+++ b/packages/@glimmer/interfaces/package.json
@@ -2,6 +2,11 @@
   "name": "@glimmer/interfaces",
   "version": "0.40.1",
   "repository": "https://github.com/glimmerjs/glimmer-vm/tree/master/packages/@glimmer/interfaces",
+  "dependencies": {
+    "@glimmer/reference": "^0.40.1",
+    "@glimmer/runtime": "^0.40.1",
+    "@simple-dom/interface": "^1.4.0"
+  },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


zimmi88:
Welp, this doesn't work. Looks like this introduces a circular dependency. =(

Should there be peer dependencies declaring that these deps should exist? Or just leave it up to consuming packages to run build and add in whatever the build complains about?

Turbo87:
closing in favor of https://github.com/glimmerjs/glimmer-vm/pull/1049 and the previous PRs. thanks for your contribution though 🙏

PR #941, opened 2019-05-07
syntax/print: Fix self-closing angle bracket component invocations

Turbo87:
Previously `<Foo />` would have been printed as `<Foo></Foo>`. After this PR the printer uses the `selfClosing` property correctly and prints out `<Foo />`.

/cc @rwjblue @tylerturdenpants


locks:
Thanks for the fix 😎

PR #942, opened 2019-05-11
syntax: Add support for `chained` inverse blocks

Turbo87:
It seems that some code was added lately to support the `chained` property of the Handlebars AST in the printer, but the parser was not correctly hooked up yet. This PR adds the missing pieces so that `else if` is correctly represented in the Glimmer AST now.

Since the `b.blockItself()` signature was changed this should be considered a breaking change.

Closes https://github.com/glimmerjs/glimmer-vm/pull/366

/cc @rwjblue @wycats

### DIFF
--- a/packages/@glimmer/syntax/lib/builders.ts
+++ b/packages/@glimmer/syntax/lib/builders.ts
@@ -457,12 +457,14 @@ function buildProgram(
 function buildBlockItself(
   body?: AST.Statement[],
   blockParams?: string[],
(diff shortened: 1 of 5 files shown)


rwjblue:
Can you rebase (conflicted by https://github.com/glimmerjs/glimmer-vm/pull/938), I think the main issue will be the updated print-test.ts format...

Turbo87:
@rwjblue done

PR #951, opened 2019-06-29
Preserve mustache whitespace stripping information

CvX:
Fixes https://github.com/rajasegar/ember-angle-brackets-codemod/issues/46 (a part of the prettier quest: https://github.com/jgwhite/prettier/issues/1)

This PR most likely requires some work (e.g. not sure how I feel about those `{ open: true, close: false }` all over the place), so all feedback is welcome! 🙂 

cc: @GavinJoyce @rwjblue @jgwhite

### DIFF
--- a/packages/@glimmer/syntax/test/generation/print-test.ts
+++ b/packages/@glimmer/syntax/test/generation/print-test.ts
@@ -101,8 +101,19 @@ QUnit.module('[glimmer-syntax] Code generation - source -> source', function() {
 
     // newlines after opening block
     '{{#each}}\n  <li> foo </li>\n{{/each}}',
-
-    // TODO: fix whitespace control in codemod mode
-    // '\n{{~#foo-bar~}} {{~/foo-bar~}}  ',
   ].forEach(buildTest);
+
+  test('whitespace control is preserved', function(assert) {
+    let before = '\n{{~var~}}  ';
+    let after = '{{~var~}}';
+
+    assert.equal(printTransform(before), after);
+  });
+
+  test('block whitespace control is preserved', function(assert) {
+    let before = '\n{{~#foo-bar~}} {{~else if x~}} {{~else~}} {{~/foo-bar~}}  ';
+    let after = '{{~#foo-bar~}}{{~else if x~}}{{~else~}}{{~/foo-bar~}}';
+
+    assert.equal(printTransform(before), after);
+  });
 });
--- a/packages/@glimmer/syntax/lib/builders.ts
+++ b/packages/@glimmer/syntax/lib/builders.ts
@@ -14,7 +14,8 @@ function buildMustache(
   params?: AST.Expression[],
   hash?: AST.Hash,
   raw?: boolean,
-  loc?: AST.SourceLocation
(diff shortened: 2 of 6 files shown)


CvX:
It’s the Edge tests that are randomly failing: https://travis-ci.org/glimmerjs/glimmer-vm/builds

I’ve rebased and force-pushed to trigger Travis. Gotta spin the wheel. 😉

rwjblue:
@wycats / @chancancode - Any objections to the AST changes made here? In practical terms, this is quite important to any codemod situations (e.g. prettier, ember-template-recast, angle-bracket-codemod, etc)...

packages/@glimmer/syntax/test/generation/print-test.ts
```diff
@@ -101,8 +101,19 @@ QUnit.module('[glimmer-syntax] Code generation - source -> source', function() {
...
+  test('whitespace control is preserved', function(assert) {
+    let before = '\n{{~var~}}  ';
+    let after = '{{~var~}}';
+
+    assert.equal(printTransform(before), after);
+  });
+
+  test('block whitespace control is preserved', function(assert) {
+    let before = '\n{{~#foo-bar~}} {{~else if x~}} {{~else~}} {{~/foo-bar~}}  ';
+    let after = '{{~#foo-bar~}}{{~else if x~}}{{~else~}}{{~/foo-bar~}}';
```
rwjblue: Hmm, why is the newline removed? We want to ensure that it isn't removed...

rwjblue: Would you mind working through this in a follow up? I'd like to merge now...

CvX: @rwjblue The newline is removed by handlebars.js parser:

[handlebars/compiler/base.js#L22-L23](https://github.com/wycats/handlebars.js/blob/a510d8225222c4cb9c5f99cab9652a5dde060273/lib/handlebars/compiler/base.js#L22-L23)
[handlebars/compiler/whitespace-control.js#L30-L35](https://github.com/wycats/handlebars.js/blob/a510d8225222c4cb9c5f99cab9652a5dde060273/lib/handlebars/compiler/whitespace-control.js#L30-L35)

We could add a `preserveWhitespace` option that would allow to return the AST with all whitespace information left intact.

Currently the only `WhitespaceControl` option is [`ignoreStandalone`](https://github.com/wycats/handlebars.js/blob/a510d8225222c4cb9c5f99cab9652a5dde060273/lib/handlebars/compiler/whitespace-control.js#L9), which we use in [the `codemod` mode](https://github.com/glimmerjs/glimmer-vm/blob/83a0c2fdfa8f22fd353dcab32ea41c4bcacaa379/packages/%40glimmer/syntax/lib/parser/tokenizer-event-handlers.ts#L389-L391).

rwjblue: Ya, that seems good to me. In the meantime, I think we can work around by adding back the whitespace that is stripped (at least in "codemod" mode).

PR #952, opened 2019-07-07
feat: builder aliases

lifeart:


packages/@glimmer/syntax/lib/builders.ts
```diff
@@ -572,6 +569,51 @@ export default {
   },
 };
 
+const locationBuilders = {
+  loc: buildLoc,
+  pos: buildPosition,
+};
+
+const modernBuilders = {
+  MustacheStatement: legacyBuilders.mustache,
```
rwjblue: These should be camel case with initial lower case character. That is what recast and jscodeshift use (by way of [ast-types](https://github.com/benjamn/ast-types)).

For example, this one would be `b.mustacheStatement(...args...)`.

lifeart: I'm confused with babel builders

https://astexplorer.net/#/gist/6fa4e6380a5d53d6385efb7716012ba0/390e735ebaea048f40b7ecbc058c544cc58302df

Looks like they support both cases

lifeart: @rwjblue c builder aliases added (amel case with initial lower case character), should I remove capitalized?

rwjblue: Hmm. Who uses initial capital case vs who supports initial lower? I’d really prefer not to have three different builder functions for the same thing...

lifeart: @rwjblue how about `Program/program` builder, should we have it? (it's return "Template" type)

rwjblue: I'd prefer to only create _new_ aliases for the "go forward" AST types.

packages/@glimmer/syntax/lib/builders.ts
```diff
@@ -572,6 +569,85 @@ export default {
...
+  attrNode: legacyBuilders.attr,
+  textNode: legacyBuilders.text,
+  subExpression: legacyBuilders.sexpr,
+  pathExpression: legacyBuilders.path,
+  concatStatement: legacyBuilders.concat,
+  hash: legacyBuilders.hash,
+  hashPair: legacyBuilders.pair,
+  literal: legacyBuilders.literal,
+  program: legacyBuilders.program, // returning type - Template
+  block: legacyBuilders.blockItself,
```
rwjblue: When merged, this will conflict with the much older `builders.block`

https://github.com/glimmerjs/glimmer-vm/blob/8b92c2c1a71cb11f1b81e5bbce3f4397d06fbfe6/packages/%40glimmer/syntax/lib/builders.ts#L544

Not sure that we can get away with the merging behaviors because of this 🤔

lifeart: we can apply `Capitalized` case or crate named export for updated builders, instead of 'default'

packages/@glimmer/syntax/lib/builders.ts
Turbo87: @rwjblue the usual convention is that UpperCase is for class names and lowerCase for functions. so a factory function for `FooBar` would be called `fooBar()` and use `new FooBar()` internally.

tl;dr I agree that we should only use camelCase function names

PR #953, opened 2019-07-08
Failing test showing incorrect ...attributes location info.

rwjblue:
This is a failing test PR (for now) showing that when `...attributes` is not the "last" attribute for a given `ElementNode`, the location information is incorrect.


rwjblue:
The underlying issue is in simple-html-tokenizer. We allow spaces to separate an attribute name from the equals sign, in order to support this we "consume" all whitespace we encounter between the end of the "valueless" attribute and the beginning of the next one (or we hit `>`).

PR #955, opened 2019-07-10
possible bug fix in `pair` function

bekzod:
`pair` function probably should return non-constant tag when one of tags is constant, not sure how to test it


wycats:
The easiest way to test this would be to create one constant tag and one DirtyableTag, combine them using `pair`, bump the dirtyable tag and confirm that the pair tag returns false for validate(originalSnapshot)

PR #960, opened 2019-07-22
Update simple-html-tokenizer to v0.5.8

josemarluedke:
Upgrade `simple-html-tokenizer` to get changes from https://github.com/tildeio/simple-html-tokenizer/pull/69. 

Unfortunately, this breaks a test. 
Screen Shot 2019-07-22 at 4 27 47 PM

https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/syntax/test/parser-node-test.ts#L53-L70

Should the changes in that PR affect SVG title? Happy to update the test if it makes sense.

Interestingly, MDN docs say that SVG title can be "Any elements or character data". 
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title#Usage_notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "babel-plugin-debug-macros": "^0.1.11",
     "ember-cli-browserstack": "^0.0.7",
     "handlebars": "^4.0.13",
-    "simple-html-tokenizer": "^0.5.7"
+    "simple-html-tokenizer": "^0.5.8"
   },
   "devDependencies": {
     "@types/qunit": "^2.0.31",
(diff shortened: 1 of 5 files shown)


CvX:
I'm looking into it. 🙂 (On my `fix-rehydration` branch there's also a second failing test)

You'd probably want to also bump the version in `packages/@glimmer/integration-tests` and `packages/@glimmer/syntax`. 🙂

CvX:
> Should the changes in that PR affect SVG title? Happy to update the test if it makes sense.
> 
> Interestingly, MDN docs say that SVG title can be "Any elements or character data".
> https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title#Usage_notes

Yeah, the test should be updated. A quick test in a browser shows that html tags and entities don't get parsed in svg title.

josemarluedke:
@CvX Thanks for the responses. I have updated again and tests should pass now. 

@rwjblue Should I cherry-pick this into `release-0-38-alpha` branch?

rwjblue:
Yes please!

PR #963, opened 2019-07-31
[BACKPORT] Performance Tuning for Ember 3.13

pzuraq:
This PR represents the Glimmer portion of https://github.com/emberjs/ember.js/pull/18223

### DIFF
--- a/packages/@glimmer/reference/lib/reference.ts
+++ b/packages/@glimmer/reference/lib/reference.ts
@@ -1,4 +1,5 @@
-import { Opaque } from '@glimmer/util';
+import { Opaque, Option } from '@glimmer/util';
+import { Revision, Tag, Tagged, value, validate } from './validators';
 
 export interface Reference<T> {
   value(): T;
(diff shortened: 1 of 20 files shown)


packages/@glimmer/reference/lib/validators.ts
```diff
@@ -1,398 +1,331 @@
...
 
-  while (node !== null) {
-    let tag = node.tag;
+  private isUpdating = false;
+  private subtag: Tag | null = null;
+  private subtags: Tag[] | null = null;
 
-    if (tag === VOLATILE_TAG) return VOLATILE_TAG;
-    if (tag !== CONSTANT_TAG) optimized.push(tag);
+  [TYPE]: MonomorphicTagType;
```
chadhietala: I may just be not understanding what the actual `MonomorphicTag` type is, but why do you have to use the brands here instead of the const enum?

pzuraq: It's because `TYPE` is the intersection of all of the types on all of the classes that make up the `MonomorphicTag` type (it being the intersection of all the subtypes), which is slightly different than the actual enum. This is part of what gets us the ability to magically do `new MonomorhpicTagImpl(TagType.Dirtyable)` and have that type to the `DirtyableTag` interface.

@krisselden may be able to explain better, this was his dark type magic that I'm still trying to wrap my head around 😄

rwjblue:
@pzuraq - Where are we at this this?

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -9,3 +10,128 @@ export default Reference;
...
+  private lastValue: Option<T> = null;
+  private lastRevision: Option<Revision> = null;
+  private initialized = false;
+
+  constructor(reference: VersionedReference<T>) {
+    this.tag = reference.tag;
+    this.reference = reference;
+  }
+
+  peek(): T {
```
krisselden: why does peek() force it compute? that smells a bit.

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -9,3 +10,128 @@ export default Reference;
...
+}
+
+export type Validation<T> = T | NotModified;
+
+export type NotModified = 'adb3b78e-3d22-4e4b-877a-6317c2c5c145';
+
+const NOT_MODIFIED: NotModified = 'adb3b78e-3d22-4e4b-877a-6317c2c5c145';
+
+export function isModified<T>(value: Validation<T>): value is T {
+  return value !== NOT_MODIFIED;
```
krisselden: String equality is a little weird here.  why not make NOT_MODIFIED a Symbol('NOT_MODIFIED')?

If we are trying to avoid symbols, you are only using === for a unique value, you can just make an {} and then type it as something unique.

```
interface NotModified { __brand__: 'adb3b78e-3d22-4e4b-877a-6317c2c5c145' }
const NotModified: NotModified = {} as any;

export function isModified<T>(value: T | NotModified): value is T {
    return value !== NotModified;
}
```

pzuraq: To be fair, this code is just moved from the validators file. I think it’d make sense to update it in a follow up 👍

packages/@glimmer/reference/lib/reference.ts
wycats: @pzuraq What's up here?

packages/@glimmer/reference/lib/reference.ts
wycats: Generally speaking, using a string as a fake symbol allows you to both use the value as a key in an object, as well as cheaply/easily comparing it. I would generally prefer to use symbols as unique tokens, but until we can, I prefer long strings containing UUIDs.

wycats:
I'm pressing the merge button but could use a response to Kris' question.

packages/@glimmer/reference/lib/reference.ts
pzuraq: This code was also ported over without changes from the `validators.ts` file. I agree it probably doesn't make sense to force a compute from a method called `peek`, but I do think that's outside of the scope of this PR/change, happy to address it in a followup though when we start cleaning some stuff up

rwjblue:
@pzuraq - Can you make a PR for this against master? I don't want us to loose these optimizations when we do finally update Ember to master of glimmer-vm...

Issue #946, opened 2019-05-29
Type errors when using TypeScript v3.5.1


josemarluedke:
I started having the following TypeScript type errors when upgraded to v3.5.1.

Note: This is an Octane app (Ember + Glimmer components).

```
node_modules/@glimmer/runtime/dist/types/lib/component/curried-component.d.ts:6:87 - error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type 'CurriedComponentDefinition' is not assignable to type 'Maybe<Dict<unknown>>'.
    Type 'CurriedComponentDefinition' is not assignable to type 'Dict<unknown>'.
      Index signature is missing in type 'CurriedComponentDefinition'.

6 export declare function isComponentDefinition(definition: Maybe<Dict>): definition is CurriedComponentDefinition;
                                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@glimmer/util/dist/types/lib/destroy.d.ts:3:69 - error TS2677: A type predicate's type must be assignable to its parameter's type.
  Type 'SymbolDestroyable' is not assignable to type 'Maybe<Dict<unknown>>'.
    Type 'SymbolDestroyable' is not assignable to type 'Dict<unknown>'.
      Index signature is missing in type 'SymbolDestroyable'.

3 export declare function isDestroyable(value: Maybe<Dict>): value is SymbolDestroyable;
```

chancancode:
Hmm, why is your typescript building glimmer-vm?

josemarluedke:
My app is built using TypeScript (ember-cli-typescript). 

Here is my tsconfig.json for context.

Note that `path` on glimmer/component is to solve issues with types (https://github.com/glimmerjs/glimmer.js/issues/179). 

```json
{
  "compilerOptions": {
    "target": "ES2017",
    "allowJs": true,
    "allowSyntheticDefaultImports": true,
    "alwaysStrict": true,
    "baseUrl": ".",
    "experimentalDecorators": true,
    "inlineSourceMap": true,
    "inlineSources": true,
    "module": "ESNext",
    "moduleResolution": "node",
    "noEmit": true,
    "noEmitOnError": false,
    "noFallthroughCasesInSwitch": true,
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "strictNullChecks": true,
    "strictPropertyInitialization": true,
[...]

PR #964, opened 2019-08-13
[FEAT] Perf Tuning + Functional API

pzuraq:
This adds the perf tuning and functional API changes made in the 0.38
release to master, from #963

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,10 @@
     "tslint-config-prettier": "^1.15.0",
     "tslint-plugin-prettier": "^2.0.1"
   },
+  "resolutions": {
+    "typescript": "3.3.3",
+    "broccoli-typescript-transpiler/typescript": "3.3.3"
+  },
   "changelog": {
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -87,6 +87,10 @@
     "tslint-config-prettier": "^1.15.0",
     "tslint-plugin-prettier": "^2.0.1"
   },
+  "resolutions": {
```
rwjblue: Why do we need these resolutions? When can we remove them (maybe make an issue to track that)?

pzuraq: TypeScript doesn’t allow us to cast `any` to type `unique symbol` until more recently, and its the only way we can polyfill Symbol and type it correctly. We should likely update to the latest TS when possible (brocolli-typescript-transpiler needs to update). There are some other breaking changes on latest though

pzuraq: I think https://github.com/glimmerjs/glimmer-vm/issues/946 kind of tracks this

pzuraq: Opened #965

PR #968, opened 2019-09-10
Run ESLint on *all* JS files in the project

Turbo87:
as promised in #967 

and thanks to GitHub for showing the commits in the wrong order 😂


Turbo87:
@rwjblue any thoughts?

PR #971, opened 2019-09-22
Handle missing remote element marker gracefully.

melmerp:
This is a continuation of pairing with @tomdale.

Prior to this change, a missing marker for a remote element would cause rehydration to error which caused the app to completely cease working.

After this change, a missing marker for a remote element is gracefully ignored and the remote element is rendered on the client.

Context:
We have a use case when using FastBoot APIs to SSR where we would like to early flush without waiting for the visitPromise to resolve with the head content for non-seo bot users. When using the 'ember-cli-head' addon, this causes problems as rehydration expected the 'in-element' from 'ember-cli-head' to SSR with remote element markers (<script glmr="cursorId"></script>). This PR attempts to gracefully handle the missing remote element marker case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   },
   "devDependencies": {
     "@types/qunit": "^2.0.31",
-    "amd-name-resolver": "^1.0.0",
+    "amd-name-resolver": "^1.3.1",
     "auto-dist-tag": "^1.0.0",
     "babel-plugin-nukable-import": "^0.4.2",
     "babel-plugin-strip-glimmer-utils": "^0.1.1",
--- a/packages/@glimmer/integration-tests/lib/modes/rehydration/builder.ts
+++ b/packages/@glimmer/integration-tests/lib/modes/rehydration/builder.ts
@@ -12,7 +12,7 @@ export class DebugRehydrationBuilder extends RehydrateBuilder {
     if (node.nodeType !== 8) {
       if (el.nodeType === 1) {
         // don't stat serialized cursor positions
-        if (el.tagName !== 'SCRIPT' && !el.getAttribute('gmlr')) {
+        if (el.tagName !== 'SCRIPT' || !el.getAttribute('glmr')) {
           this.clearedNodes.push(node);
         }
       } else {
--- a/packages/@glimmer/integration-tests/test/initial-render-test.ts
+++ b/packages/@glimmer/integration-tests/test/initial-render-test.ts
@@ -244,9 +244,17 @@ class Rehydration extends AbstractRehydrationTests {
 
   @test
   'in-element can rehydrate'() {
-    let template = '<outer>{{#in-element remote}}<inner>Wat Wat</inner>{{/in-element}}</outer>';
+    let template = strip`
+      <outer><prefix></prefix>
+      {{#in-element remote}}<inner>Wat Wat</inner>{{/in-element}}
+      <suffix></suffix></outer>
+      `;
     let doc = this.delegate.serverDoc;
     let remote = doc.createElement('remote');
+    let prefix = doc.createElement('prefix');
+    let suffix = doc.createElement('suffix');
+    remote.appendChild(prefix);
+    remote.appendChild(suffix);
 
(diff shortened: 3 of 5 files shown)


rwjblue:
@melmerp / @tomdale - I see that this is still marked as a draft, what other changes are needed? Mind updating the description to add a checklist?

package.json
```diff
@@ -46,7 +46,7 @@
   },
   "devDependencies": {
     "@types/qunit": "^2.0.31",
-    "amd-name-resolver": "^1.0.0",
+    "amd-name-resolver": "^1.3.1",
```
chadhietala: Do we need to update this?

melmerp: On a fresh checkout, running yarn produces the error below. This was how Tom and I resolved it. Is it not correct?

`mpascual@mpascual-mn2.linkedin.biz ~/P/tmp> git clone git@github.com:glimmerjs/glimmer-vm.git
Cloning into 'glimmer-vm'...
remote: Enumerating objects: 38, done.
remote: Counting objects: 100% (38/38), done.
remote: Compressing objects: 100% (37/37), done.
remote: Total 43688 (delta 0), reused 14 (delta 0), pack-reused 43650
Receiving objects: 100% (43688/43688), 10.59 MiB | 7.32 MiB/s, done.
Resolving deltas: 100% (28556/28556), done.
mpascual@mpascual-mn2.linkedin.biz ~/P/tmp> cd glimmer-vm
mpascual@mpascual-mn2.linkedin.biz ~/P/t/glimmer-vm (master)> yarn
yarn install v1.17.3
[1/4] 🔍  Resolving packages...
warning Resolution field "typescript@3.3.3" is incompatible with requested version "typescript@~3.2.1"
[2/4] 🚚  Fetching packages...
error https://github.com/ember-cli/amd-name-resolver.git: Extracting tar content of undefined failed, the file appears to be corrupt: "Invalid tar header. Maybe the tar is corrupted or it needs to be gunzipped?"
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
mpascual@mpascual-mn2.linkedin.biz ~/P/t/glimmer-vm (master)>                                                                                                                                                  1 ↵
`

packages/@glimmer/integration-tests/lib/modes/rehydration/builder.ts
```diff
@@ -12,7 +12,7 @@ export class DebugRehydrationBuilder extends RehydrateBuilder {
     if (node.nodeType !== 8) {
       if (el.nodeType === 1) {
         // don't stat serialized cursor positions
-        if (el.tagName !== 'SCRIPT' && !el.getAttribute('gmlr')) {
+        if (el.tagName !== 'SCRIPT' || !el.getAttribute('glmr')) {
```
tomdale: Doesn’t this need to be `&&` still? If I’m reading this correctly, this will now match *any* script element, not just those with the `glmr` attribute. (Great catch on the typo!)

melmerp: I think it needs to be `||`. Please correct me if I'm wrong, but I believe that the point of this code is to track the removal of nodes that are NOT an element like the remote element marker. If I'm wrong, the following is totally wrong...

If I'm right, then this code can be written more like:
```
if (!(el.tagName === 'SCRIPT' && el.getAttribute('glmr'))) { // if not a remote element marker
  this.clearNodes.push(node);
}
```
By [De Morgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) (reaching back into my digital design days), these two are equivalent: 
```
const A = el.tagName === 'SCRIPT';
const B = el.getAttribute('glmr');
!(A && B) === (!A || !B) // these are equivalent
```
If you want me to switch to the other form, I'm totally in favor of it as I believe it's probably more readable but I thought the original author was intending to apply the transformation.

tomdale: This is what I get for doing code reviews past my bedtime… thanks for the thorough explanation!

PR #974, opened 2019-09-25
maybe perf tuning | strict ifs

lifeart:
@pzuraq is it make sence?

// https://github.com/glimmerjs/glimmer-vm/pull/964


lifeart:
possible rule - https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md

krisselden:
in general for hotspots the explicit checks are better than falsy coercions, especially if you have several types that falsy coerce and one of them happens occasionally.

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -59,22 +59,22 @@ export class ReferenceCache<T> implements Tagged {
...
-    if (!this.initialized) {
+    if (this.initialized === false) {
       return this.initialize();
     }
 
     let { reference, lastRevision } = this;
     let tag = reference.tag;
 
-    if (validate(tag, lastRevision as number)) return NOT_MODIFIED;
+    if (validate(tag, lastRevision as number) === true) return NOT_MODIFIED;
```
krisselden: does validate not always return a boolean? if validate always returns a boolean then there is no benefit from this

packages/@glimmer/runtime/lib/vm/stack.ts
```diff
@@ -40,8 +40,8 @@ export class InnerStack {
   }
 
   write(pos: number, value: unknown): void {
-    if (isImmediate(value)) {
-      this.writeRaw(pos, encodeImmediate(value));
+    if (isImmediate(value) === true) {
```
krisselden: if isImmediate always returns a boolean then there is no coercion you are avoiding by doing this.

pzuraq:
@lifeart can you add that eslint rule in this PR? It should allow the cases that Kris pointed out, where we know the value is always a boolean, but prevent all the other cases where we don't know and end up coercing.

lifeart:
Don't know why GH showing conflict for `tslint.json`

lifeart:
@krisselden, @pzuraq done.

packages/@glimmer/syntax/lib/builders.ts
```diff
@@ -9,6 +9,10 @@ import { StringLiteral, BooleanLiteral, NumberLiteral } from './types/handlebars
 export type BuilderPath = string | AST.PathExpression;
 export type TagDescriptor = string | { name: string; selfClosing: boolean };
 
+function withFallback<T, U>(originalValue: T | null | undefined, fallbackValue: U): T | U {
```
pzuraq: We should consider changing this to nullish coalescing once TS 3.7 is released

PR #976, opened 2019-09-26
Replace TSLint with ESLint

Turbo87:
TSLint is [deprecated](https://medium.com/palantir/tslint-in-2019-1a144c2317a9) and replaced by using ESLint with a TypeScript-aware parser instead.

PR #977, opened 2019-09-26
CI: Use Node 8

Turbo87:
Node 6 is no longer officially supported so we should update to at least Node 8 on our CI system

Luckily, since we use TypeScript for everything here this does not have to be considered a breaking change 🎉 

This should hopefully unblock https://github.com/glimmerjs/glimmer-vm/pull/976

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ branches:
 
 language: node_js
 node_js:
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks @Turbo87!

PR #979, opened 2019-10-11
Fix AST loc info for `ConcatStatement`s containing `TextNode`s

rwjblue:
Prior to this change, the following snippet would show the `static ` `TextNode` portion of the `ConcatStatement` as starting at column 13, but it actually starts at 12.

```hbs
<div class="static {{dynamic}}"></div>
```

**All** `TextNode`'s that were part of a `ConcatStatement` were off by one in the same way.

The main issue here is that when `simple-html-tokenizer` calls `appendToAttributeValue` the `this.tokenizer.line`/`this.tokenizer.column` counters have already been incremented, and since we are using those as the starting location information for the new `TextNode` they were wrong (in an "off by one" way). We had already identified that this was an issue when the `TextNode` started with `\n`, but failed to realize that it **also** applies to non-newline cases as well.

### DIFF
--- a/packages/@glimmer/syntax/test/loc-node-test.ts
+++ b/packages/@glimmer/syntax/test/loc-node-test.ts
@@ -326,20 +326,24 @@ foo"
   data-derp="foo
 {{concat ''}}
     huzzah"
+  data-qux="{{zomg}} static"
     ></div>
   `);
 
   let [, div] = ast.body;
   if (assertNodeType(div, 'ElementNode')) {
-    let [dataFoo, dataBar, dataDerp] = div.attributes;
+    let [dataFoo, dataBar, dataDerp, dataQux] = div.attributes;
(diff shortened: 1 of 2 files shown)


packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -349,10 +353,16 @@ foo"
 
     if (assertNodeType(dataDerpValue, 'ConcatStatement')) {
       let [fooStaticText, concat, huzzahStaticText] = dataDerpValue.parts;
-      locEqual(fooStaticText, 9, 14, 10, 0);
+      locEqual(fooStaticText, 9, 13, 10, 0);
```
rwjblue: The prior assertion was wrong :sob:, and this test should have _already_ been failing

PR #981, opened 2019-10-21
Update ember-cli-browserstack so that BrowserStack tests will continue to work

kategengler:
Move ember-cli-browserstack to devDependencies; not sure it was ever
correct in dependencies


rwjblue:
Thank you!

Issue #978, opened 2019-10-09
[@glimmer/syntax] disallow modifier usage in closing tag


lifeart:
for now - this is valid syntax
```hbs
<div></ {{mod-name}} div >
```
and it's parsed like
![image](https://user-images.githubusercontent.com/1360552/66519691-75becd00-eaf0-11e9-870b-e142878275b2.png)

lifeart:
Strange node parsing behavour:

```hbs
<~div~></~div~ > -> ElementNode.tag: 'div~'
<"div"></"div" > -> ElementNode.tag: 'div"'
<\div></\div>    -> ElementNode.tag: 'div'
<#div></div>     -> ElementNode.tag: 'div'
<div></#div>     -> ElementNode.tag: 'div'
<?div?></?div?>  -> ElementNode.tag: 'div?'
<.div></div>     -> ElementNode.tag: 'div'
<.div></.div>    -> ElementNode.tag: 'div'
```

rwjblue:
The issue reported (having anything in the closing tag) should be disallowed. The other examples are unrelated (and possibly their own bugs).

PR #982, opened 2019-10-23
Throw error if closing tag contains Muststache Statement

camerondubas:
Made an attempt at resolving https://github.com/glimmerjs/glimmer-vm/issues/978

I wrote a helper method that checks whether or not a closing tag contains any `MustacheStatements`. A `SyntaxError` is thrown if it does.

---

I'm quite open to feedback as I feel like there's a cleaner way to do this.

The major issue I ran into was that `EndTag` itself doesn't contain any details about whether or not contains any `Nodes`/`Statements`. This meant that I had to parse the `element.children` to see if there were any `children` within the closing tag. 

Currently I explicitly only check for `MustacheStatement`s, since the original issue (https://github.com/glimmerjs/glimmer-vm/issues/978) was specific to modifiers, but this could be expanded to check for any children within the `EndTag`.

### DIFF
--- a/packages/@glimmer/syntax/test/parser-node-test.ts
+++ b/packages/@glimmer/syntax/test/parser-node-test.ts
@@ -610,6 +610,20 @@ test('disallowed mustaches in the tagName space', function(assert) {
   }, /Cannot use mustaches in an elements tagname: `{{bar` at L1:C6/);
 });
 
+test('disallowed mustaches in closing tag', function(assert) {
+  assert.throws(() => {
+    parse('\nbefore <div></ {{some-modifier}} div>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
+
+  assert.throws(() => {
+    parse('\nbefore <div></div {{some-modifier}}>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
+
+  assert.throws(() => {
+    parse('\nbefore <div><span></span {{some-modifier}}></div>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
+});
+
 test('mustache immediately followed by self closing tag does not error', function() {
   let ast = parse('<FooBar data-foo={{blah}}/>');
   let element = b.element('FooBar/', ['attrs', ['data-foo', b.mustache('blah')]]);
--- a/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
+++ b/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
@@ -302,6 +302,9 @@ function validateEndTag(
     // <input> or <br />, so we need to check for that here. Otherwise, we would
     // throw an error for those cases.
     error = 'Invalid end tag ' + formatEndTagInfo(tag) + ' (void elements cannot have end tags).';
+  } else if (mustacheInClosingTag(tag, element)) {
+    error =
+      'Invalid end tag ' + formatEndTagInfo(tag) + '. (closing tags cannot contain modifiers).';
   } else if (element.tag === undefined) {
     error = 'Closing tag ' + formatEndTagInfo(tag) + ' without an open tag.';
   } else if (element.tag !== tag.name) {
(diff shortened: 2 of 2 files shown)


packages/@glimmer/syntax/test/parser-node-test.ts
```diff
@@ -610,6 +610,20 @@ test('disallowed mustaches in the tagName space', function(assert) {
...
+    parse('\nbefore <div></ {{some-modifier}} div>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
+
+  assert.throws(() => {
+    parse('\nbefore <div></div {{some-modifier}}>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
+
+  assert.throws(() => {
+    parse('\nbefore <div><span></span {{some-modifier}}></div>');
+  }, 'Error: Invalid end tag `div` (on line 2). (closing tags cannot contain modifiers).');
```
lifeart: should error point into `span` instead of `div`?

camerondubas: @lifeart Thanks for this! You're absolutely right. This comment made me realize my tests weren't actually testing correctly, which lead me down the path the determining how to solve this whole issue much more elegantly.

packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -148,6 +148,15 @@ export abstract class HandlebarsNodeVisitors extends Parser {
         );
 
       case TokenizerState.beforeAttributeName:
+        if (this.currentNode && this.currentNode.type === 'EndTag') {
```
rwjblue: I'm surprised we enter this state during a close tag 🤔

camerondubas: I was surprised as well. This state is hit in this scenario: `<div></div {{some-modifier}}>`, and it was actually already throwing this error`"Error: expected start tag"` which is triggered in the `this.currentStartTag` call on line 160, but doesn't really describe the error accurately. 

`currentStartTag` snippet:
```js
  get currentStartTag(): Tag<'StartTag'> {
    let node = this.currentNode;
    assert(node && node.type === 'StartTag', 'expected start tag');
    return node as Tag<'StartTag'>;
  }
```

If you comment out this `if` check, two of the tests I added will fail, throwing the `expected start tag` error instead of the one added in this PR.

rwjblue: OK, gotcha, thank you for explaining.

I _believe_ what is going on is that the HTML spec requires the tagname **immediately** after the `</`.

See the spec info here:

https://html.spec.whatwg.org/multipage/parsing.html#end-tag-open-state

And our implementation here:

https://github.com/tildeio/simple-html-tokenizer/blob/691662c9f00aec719d0eec8e622ed99935119c55/src/evented-tokenizer.ts#L482-L491

I think _essentially_ we want an `else` to the `if` above that throws a useful error (RE: invalid end tag). I think that would look something like:

```js
    endTagOpen() {
      let char = this.consume();

      if (char === '@' || char === ':' || isAlpha(char)) {
        this.transitionTo(TokenizerState.endTagName);
        this.tagNameBuffer = '';
        this.delegate.beginEndTag();
        this.appendToTagName(char);
      } else {
        this.delegate.reportSyntaxError('invalid closing tag');
      }
    }
```

Would you be interested in working on a fix over in simple-html-tokenizer also?

rwjblue: I also think this is a bit wrong:

https://github.com/tildeio/simple-html-tokenizer/blob/691662c9f00aec719d0eec8e622ed99935119c55/src/evented-tokenizer.ts#L266-L282

We have created the `endTagName` state, and made it _essentially_ the same as the `tagName` state (the spec actually says to use the `tagName` state ([see here](https://html.spec.whatwg.org/multipage/parsing.html#end-tag-open-state)), but thats a tad bit difficult for us). But honestly, I don't think we actually are allowed to enter those other states (`beforeAttributeName` and `selfClosingStartTag`). We should probably add errors there too.

camerondubas: I'll make an attempt at fixing those issues in Simple HTML Tokenizer this weekend. I'll open a PR with the fixes, or an issue, if I'm unable to come up with solutions

Thanks for looking into this!

camerondubas: Made a PR here https://github.com/tildeio/simple-html-tokenizer/pull/73 would love feedback 😄

PR #983, opened 2019-10-28
Ensure codemod mode ignores whitespace control.

rwjblue:
`handlebars.parse` automatically applies whitespace control to the AST before it is returned. For example, parsing the following template ([feel free to review the AST on ASTExplorer](https://astexplorer.net/#/gist/f850c5457b808756107c0bc5ad226989/b0366df64be5c01b5f9eadd93e795b3b9d65d2c7)):

```hbs
 {{#foo}}
   {{~bar~}} {{baz~}}
 {{/foo}}
```

Using `Handlebars.parse`, the AST returned would have truncated the following whitespace:

* The whitespace prior to the `{{#foo}}`
* The newline following `{{#foo}}`
* The leading whitespace before `{{~bar~}}`
* The whitespace between `{{~bar~}}` and `{{baz~}}`
* The newline after `{{baz~}}`
* The whitespace prior to the `{{/foo}}`

---

The inline `parseWithoutProcessing` method can be removed when we update to a version of `handlebars` that includes https://github.com/wycats/handlebars.js/pull/1584 (likely to be 4.5.0).

### DIFF
--- a/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
+++ b/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
@@ -379,20 +379,26 @@ const syntax: Syntax = {
   Walker,
 };
 
+// emulate parseWithoutProcessing (from https://github.com/wycats/handlebars.js/pull/1584)
+// can be removed when we update to Handlebars 4.5.0
+function codemodParse(html: string, parseOptions?: HandlebarsParseOptions) {
+  // this sets up parser.yy and parser.yy.locInfo
+  handlebars.parse(html, parseOptions);
+
+  // casting to any here, because Parser is not listed in the types
+  return (handlebars as any).Parser.parse(html);
+}
+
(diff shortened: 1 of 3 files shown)


rwjblue:
Labeling this as "breaking" since the impact to tooling currently using `codemod` mode is that they have to remove work arounds for the previously stripped whitespace.

packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
```diff
@@ -379,20 +379,26 @@ const syntax: Syntax = {
...
   let ast: HBS.Program;
   if (typeof html === 'object') {
     ast = html;
+  } else if (mode === 'codemod') {
+    ast = codemodParse(html, options.parseOptions);
   } else {
-    let parseOptions = options.parseOptions || {};
-
-    if (mode === 'codemod') {
-      parseOptions.ignoreStandalone = true;
```
rwjblue: FYI - `ignoreStandalone` is not used in `parseWithoutProcessing` mode so no need to preserve passing it here.

PR #985, opened 2019-10-29
syntax: Add optional `path` argument to visitor functions

Turbo87:
This PR adds path tracking to the `traverse()` function. The node visitor methods now receive a second argument, which is a `Path` object with `node` and `parent` keys. The `parent` key can be another `Path` or `null` if we've reached the root element of the traversal (usually the `Template` node).

This type of path tracking is roughly similar to ones used by e.g. ESLint and recast and can help in the development of template-lint rules and codemods.

### DIFF
--- a/packages/@glimmer/syntax/index.ts
+++ b/packages/@glimmer/syntax/index.ts
@@ -18,6 +18,7 @@ export {
 } from './lib/traversal/errors';
 export { default as traverse } from './lib/traversal/traverse';
 export * from './lib/traversal/visitor';
+export { default as Path } from './lib/traversal/path';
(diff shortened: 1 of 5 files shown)


Turbo87:
@rwjblue I've added the additional tests. I agree that a `key` property on the `Path` might be a nice idea, but I think we can probably retrofit that in a second step. Do you think this is mergeable as-is?

Turbo87:
I've updated this PR to use a class for `Path` instead and include a `parentKey` property on it :)

rwjblue:
Thank you @Turbo87!

PR #986, opened 2019-10-29
Update to handlebars@4.5.1.

rwjblue:
handlebars@4.5.0+ has a new method `handlebars.parseWithoutProcessing`, which we were emulating.

This PR updates the minimum Handlebars version to ensure that we have the new API available (from https://github.com/wycats/handlebars.js/pull/1584), and removes the work around added in https://github.com/glimmerjs/glimmer-vm/pull/983.


rwjblue:
#983 introduce the same change (and was deemed "breaking" to existing `mode: 'codemod'` users), since this is just removing our internal work around in favor of public APIs it is not breaking (and not terribly important 😛 ).

PR #987, opened 2019-11-05
[REFACTOR] Create `@glimmer/validator`

pzuraq:
This PR extracts the tag and autotracking portions of the
`@glimmer/reference` package, which are fairly isolated from the rest
of the VM, and removes any additional dependencies they have on other VM
packages. This keeps them isolated, which makes it less likely that
type changes or issues could leak and cause problems in downstream
packages which have public types, such as `@glimmer/tracking`.

Also updates the implementation of autotracking in the VM to match Ember's directly, so that we can add this package once Ember is able to update its VM.


packages/@glimmer/runtime/lib/tags.ts
```diff
@@ -0,0 +1,30 @@
+import { Slice, LinkedListNode } from '@glimmer/util';
+import { Tag, Tagged, createCombinatorTag, CONSTANT_TAG } from '@glimmer/tag';
+
+export function combineTagged(tagged: ReadonlyArray<Tagged>): Tag {
```
rwjblue: Why is this in runtime and not tag?

packages/@glimmer/tag/lib/tracking.ts
```diff
@@ -0,0 +1,107 @@
+import { Tag, combine, CONSTANT_TAG } from './tags';
+import { createTag, dirty } from './tags';
```
rwjblue: Seems vaguely odd for tag package to include tracked. Maybe “tag” isn’t the right name? Any other options?

packages/@glimmer/runtime/lib/tags.ts
pzuraq: I wanted to avoid pulling extra types into `@glimmer/tag` for things that are essentially unrelated. These combinators are really optimizations for a couple of narrow use cases in the VM itself, and while I wasn't comfortable removing them/simplifying them without perf tests, I also don't think they really belong with the rest of the tracking infra, conceptually.

If perf wasn't a concern, or tests showed it wasn't an issue, I probably would have made them use the `combine()` function like everyone else. If perf is an issue, I think the internal `createCombinatorTag` API is good for solving the problem.

rwjblue:
@pzuraq - Would you mind leaving a little info in this PR description about how you expect Glimmer.js and Ember.js to absorb the breakage (might just be update import paths, but I dunno)? Just something new I'm trying to do to aid the upgrade...

PR #988, opened 2019-11-05
Refactor RehydrateBuilder to be more robust

rwjblue:
Given this template:

```hbs
<p>{{#if true}}<div></div>{{/if}}</p>
```

The browser auto-corrects it to:

```html
<p></p><div></div><p></p>
```

When the rehydration builder runs against this, it throws an error (and fails to continue rehydrating) leaving the app in a broken state.

---

This PR refactors a number of things about how `RehydrateBuilder` recovers from errors like the one mentioned above, it also fixes some fundamental issues with rehydration of `{{in-element}}` usage (which was previously broken).

After this refactor, the act of rehydration should **never** error (leaving the app in a horribly broken state). At worst, whatever amount of SSR'ed content has already been processed when a mismatch occurs will remain stable and the remainder of the content will be cleared just before we set up the new / correct DOM from the initial render.

---

Paired with @krisselden on most of the fixes / refactors in this PR.

TODO:

- [ ] Add failing test case for https://github.com/glimmerjs/glimmer-vm/pull/988#discussion_r344434372

### DIFF
--- a/packages/@glimmer/integration-tests/test/initial-render-test.ts
+++ b/packages/@glimmer/integration-tests/test/initial-render-test.ts
@@ -318,6 +318,81 @@ class Rehydration extends AbstractRehydrationTests {
   }
 
   @test
+  'missing attributes'() {
+    let template = '<div data-foo="true"></div>';
+    this.renderServerSide(template, {});
+    this.assertServerOutput('<div data-foo="true"></div>');
+
+    // remove the attribute
+    let element = this.element as Element;
+    let [div] = element.children;
+    div.removeAttribute('data-foo');
+
+    this.renderClientSide(template, {});
+    this.assertRehydrationStats({ nodesRemoved: 0 });
+    this.assertHTML('<div data-foo="true"></div>');
+    this.assertStableRerender();
+  }
+
+  @test
+  'remove extra attributes'() {
+    let template = '<div data-foo="true"></div>';
+    this.renderServerSide(template, {});
+    this.assertServerOutput('<div data-foo="true"></div>');
+
+    // add an extra attribute
+    let element = this.element as Element;
+    let [div] = element.children;
+    div.setAttribute('data-bar', 'oops');
+
+    this.renderClientSide(template, {});
+    this.assertRehydrationStats({ nodesRemoved: 0 });
+    this.assertHTML('<div data-foo="true"></div>');
+    this.assertStableRerender();
+  }
+
+  @test
+  'updates attribute to current value'() {
+    let template = '<div class="always-present show-me"></div>';
+    this.renderServerSide(template, {});
+    this.assertServerOutput('<div class="always-present show-me"></div>');
+
+    // mutate the attribute
+    let element = this.element as Element;
+    let [div] = element.children;
+    div.setAttribute('class', 'zomg');
+
+    this.renderClientSide(template, {});
+    this.assertRehydrationStats({ nodesRemoved: 0 });
+    this.assertHTML('<div class="always-present show-me"></div>');
+    this.assertStableRerender();
+  }
+
+  @test
+  'does not mutate attributes that already match'() {
+    let observer = new MutationObserver(mutationList => {
+      mutationList.forEach(mutation => {
+        let target = mutation.target as Element;
+        this.assert.ok(
+          false,
+          `should not have updated ${mutation.attributeName} on ${target.outerHTML}`
+        );
+      });
+    });
+
+    let template = '<div data-foo="whatever"></div>';
+    this.renderServerSide(template, {});
+    this.assertServerOutput('<div data-foo="whatever"></div>');
+
+    observer.observe(this.element as Element, { attributes: true, subtree: true });
+
+    this.renderClientSide(template, {});
+    this.assertRehydrationStats({ nodesRemoved: 0 });
+    this.assertHTML('<div data-foo="whatever"></div>');
+    this.assertStableRerender();
+  }
+
+  @test
   'Node curlies'() {
     let template = '<div>{{node}}</div>';
 
--- a/packages/@glimmer/integration-tests/test/initial-render-test.ts
+++ b/packages/@glimmer/integration-tests/test/initial-render-test.ts
@@ -197,6 +197,33 @@ class Rehydration extends AbstractRehydrationTests {
   }
 
   @test
+  'missing closing block within multiple text nodes'() {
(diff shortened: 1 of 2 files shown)


rwjblue:
CI is failing on IE11 for bizarro ways because IE11 doesn't insert an end `</p>` the same way modern browsers do.

With IE11, the following `<p>hello <div>world</div></p>` is "corrected" to `<p>hello <div>world!</div><p></p>`, but on modern browsers we get `<p>hello </p><div>world!</div><p></p>`. 🙀

scalvert:
👍

packages/@glimmer/integration-tests/test/initial-render-test.ts
```diff
@@ -516,7 +545,7 @@ class Rehydration extends AbstractRehydrationTests {
       remoteParent: clientRemoteParent,
       remoteChild: clientRemoteChild,
     });
-    this.assertRehydrationStats({ nodesRemoved: 2 });
+    this.assertRehydrationStats({ nodesRemoved: 0 });
```
rwjblue: This PR fixes rehydration for in-element, that is why this goes to expect no removed nodes.

packages/@glimmer/integration-tests/test/initial-render-test.ts
```diff
@@ -277,11 +309,8 @@ class Rehydration extends AbstractRehydrationTests {
     this.element = assertElement(host.firstChild);
 
     this.renderClientSide(template, { remote: clientRemote });
-    this.assertRehydrationStats({ nodesRemoved: 1 });
-    this.assert.equal(
-      toInnerHTML(clientRemote),
-      '<prefix></prefix><suffix></suffix><inner>Wat Wat</inner>'
-    );
+    this.assertRehydrationStats({ nodesRemoved: 2 });
```
rwjblue: Using `{{in-element` **without** passing `insertBefore` as a named argument _always_ means "clear this element".

* https://github.com/emberjs/rfcs/blob/master/text/0287-promote-in-element-to-public-api.md#small-proposed-changes
* https://github.com/glimmerjs/glimmer-vm/pull/918/files

packages/@glimmer/integration-tests/test/initial-render-test.ts
```diff
@@ -29,6 +29,11 @@ import {
 import { expect } from '@glimmer/util';
 import { SimpleElement } from '@simple-dom/interface';
 
+// `window.ActiveXObject` is "falsey" in IE11 (but not `undefined` or `false`)
+// `"ActiveXObject" in window` returns `true` in all IE versions
+// only IE11 will pass _both_ of these conditions
+const isIE11 = !(window as any).ActiveXObject && 'ActiveXObject' in window;
```
chadhietala: It would be good to annotate this stuff so it's easy to cleanup in a #worldwithoutIE11.

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -35,9 +36,7 @@ export class RehydratingCursor extends CursorImpl {
 export class RehydrateBuilder extends NewElementBuilder implements ElementBuilder {
   private unmatchedAttributes: Option<SimpleAttr[]> = null;
   [CURSOR_STACK]!: Stack<RehydratingCursor>; // Hides property on base class
-  private blockDepth = 0;
-
-  // private candidate: Option<SimpleNode> = null;
+  blockDepth = 0;
```
chadhietala: Is the only reason why this has to be `public` is to enforce the type on L101?

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -156,30 +185,45 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
...
-      assert(
-        currentCursor.nextSibling !== null &&
-          isComment(currentCursor.nextSibling) &&
-          getCloseBlockDepth(currentCursor.nextSibling) === openBlockDepth,
-        'expected close block to match rehydrated open block'
-      );
-      currentCursor.candidate = this.remove(currentCursor.nextSibling!);
-      currentCursor.openBlockDepth--;
+
+    if (!isRehydrating) {
```
chadhietala: For clarity, it might be better if this did an equality check

```js
if (isRehydrating === false) {
```

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
rwjblue: Yes, if it is `private` the "special" types that we added for `pushElement` won't work. The real issue is that `NewElementBuilder` (our super class) invokes `this.pushElement` during `RehydrateBuilder`'s `super()` call which means our class fields haven't ran yet (and therefore `this.blockDepth` is `undefined` still). So we tried to make the types such that it was clearer that the `this` within that method was "weird" (to avoid future mistakes).

kratiahuja:
👍

packages/@glimmer/integration-tests/test/initial-render-test.ts
rwjblue: Happy to! How do you think I should annotate?

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
rwjblue: Good idea, done!

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -72,55 +71,85 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
...
       let openBlockDepth = currentCursor.openBlockDepth;
       if (openBlockDepth >= currentCursor.startingBlockDepth) {
-        while (current && !(isComment(current) && getCloseBlockDepth(current) === openBlockDepth)) {
+        while (current) {
+          if (isCloseBlock(current)) {
+            let closeBlockDepth = getBlockDepth(current);
+            if (openBlockDepth >= closeBlockDepth) {
+              // cleared up until the close but we haven't closed the current
+              // block unless we are above
+              currentCursor.openBlockDepth = closeBlockDepth;
```
krisselden: remove this line, and I will describe the unit test tomorrow

rwjblue: Removed, will try to review with you in the morning to figure out the failing test case...

PR #991, opened 2019-11-07
Add a set of "chaos monkey" tests for rehydration.

rwjblue:
Based on an idea that @krisselden and I had while working on #988 this adds tests that run a number of iterations removing a random node, and confirming that the final rehydrated (e.g. client rendered) HTML is correct.

In order to make these tests actionable the infrastructure added here provides a way to replicate _specific_ iterations upon failure. The way this works is based on QUnit's internal `QUnit.config.seed` implementation. When a `QUnit` test suite is ran with the `seed` query param set (with no value e.g. `/tests/index.html?hidepassed&seed`), QUnit will choose a random seed and use that seed to choose a randomized test order. Then if a failure occurs with the randomized and seeded test run, you can run the tests again with **exactly** that seed value by
specifying the `seed` query param _with_ the failing seed value (e.g. `/tests/index.html?hidepassed&seed=THEVALUEHERE`). The implementation here piggie backs on that system to use the QUnit's seed value (along with a copied version of their [hashing function](qunitjs/qunit:src/core/utilities.js@2.9.3#L144-L158)
and [seeded randomizer](qunitjs/qunit:src/core/processing-queue.js@2.9.3#L134-L154)) in addition to the iteration being ran to always select a "known value" to remove. This means that given the correct seed (and iteration number) we can easily replicate **specific** failure scenarios without manually having to track it down.

The way you'd do this is:

1. Run the tests with `/tests/index.html?hidepassed&seed` until a failure occurs.
2. Look at the failed tests assertion, grab the rerun URL printed in the failed assertion.
3. Run the tests again with that set of query params (e.g. `/tests/index.html?hidepassed&seed=SEED_HERE&iteration=ITERATION_HERE`).

---

Currently includes the changes from #988, but will rebase once that lands.

### DIFF
--- a/packages/@glimmer/integration-tests/test/chaos-rehydration-test.ts
+++ b/packages/@glimmer/integration-tests/test/chaos-rehydration-test.ts
@@ -1,0 +1,233 @@
+import { Dict, Option } from '@glimmer/interfaces';
+import {
+  OPEN,
+  CLOSE,
+  equalTokens,
+  suite,
+  RehydrationDelegate,
+  test,
+  Content,
+  content,
+  blockStack,
+  ComponentBlueprint,
+  replaceHTML,
+  RenderTest,
+} from '..';
+import { expect } from '@glimmer/util';
+import { SimpleElement, NodeType } from '@simple-dom/interface';
+
+// `window.ActiveXObject` is "falsey" in IE11 (but not `undefined` or `false`)
+// `"ActiveXObject" in window` returns `true` in all IE versions
+// only IE11 will pass _both_ of these conditions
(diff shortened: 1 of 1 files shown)


rwjblue:
Some ideas that @krisselden just mentioned to me in DM

* have the system randomly insert unexpected nodes (at somewhat arbitrary depths)
* have the system randomly mutate nodeValue of text nodes and comments (including our "special" comments)

krisselden:
is this still a WIP?

rwjblue:
Thanks for the reminder @krisselden, rebased and updated. Should be good to land once green.

PR #992, opened 2019-11-14
[BUGFIX] Removes compile from CompileTimeResolverDelegate

pzuraq:
This PR removes the `compile` function from the
`CompileTimeResolverDelegate` interface. This function was being used by
tests to compile source to templates, specifically because there was no
other convenient way to pass it to component managers since the testing
managers in this codebase don't have an owner/container to refer to.

The solution was to remove it, and to provide the registry directly to
the testing component managers. This mirrors what embedding environments
would be expected to do more closely, and keeps the APIs of the resolver
delegates pretty clean and simple/non-extensible, preventing them from
becoming "god objects".

## Impact on consumers

There shouldn't be a noticeable impact except that `@glimmer/compiler`
will no longer be a dependency of `@glimmer/opcode-compiler`. The
`compile` functions for the delegates were only being used by tests.

### DIFF
--- a/packages/@glimmer/bundle-compiler/lib/lookup.ts
+++ b/packages/@glimmer/bundle-compiler/lib/lookup.ts
@@ -6,11 +6,9 @@ import {
   CompileTimeResolverDelegate,
   ModuleLocator,
   CompileTimeComponent,
-  Template,
 } from '@glimmer/interfaces';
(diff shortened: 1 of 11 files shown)


pzuraq:
Pair reviewed with @wycats today, everything looked good so we went ahead and merged

PR #993, opened 2019-11-16
[REFACTOR] Refactors while integrating with Ember

pzuraq:
### Changelog

- Overall updates
  - Added `@glimmer/env` to the build, usable from Ember as well.
  - Converted to native symbol, with a simple polyfill

- Updates to the Runtime builders (`AotRuntime()`,`JitRuntime()`)
  - Can now accept an options hash as the first argument containing
    either the document or append/update operations.

- Updates to `Environment`
  - Removed `DefaultEnvironment`
  - Simplified environment inheritance tree, flattened down to one class
  - Added `willDestroy` to environment for `WILL_DROP` (legacy sync drops)

- Updates to the `RuntimeEnvironmentDelegate`:
  - Rename to `EnvironmentDelegate`
  - Added `isInteractive`, equivalent of Ember's usage (should update?)
  - Added `extra` field, which is a generic type for the Environment. This 
    allows the embedding environment to pass any extra values they need.
  - Removed the `iterables` API, key for methods are now statically
    generated and entirely a concern of the IteratorImpl
  - Added `toIterator`, which allows the delegate to provide an
    `IteratorDelegate` for a given value. To keep things simpler and
    more performant, this is not yet a true browser delegate. This way,
    it matches the API of the (now internal) `IterationItem`.
      - Note for future: `key` and `memo` names are fairly confusing,
        maybe `memo` -> `index`?
  - Added `getPath`/`setPath` to allow Ember to provide a getter/setter 
    implementation, allowing for PropertyReference to delegate property access,
    along with `keyFor` in iterators.
  - Added `setTemplatePathDebugContext`/`getTemplatePathDebugContext`. These are
    used to add debug context for paths in templates (e.g. `{{this.foo}}` or 
    `{{my-helper}}`) to the DebugRenderTree in Ember. Eventually they should 
    be removed, when the DebugRenderTree has been pulled upstream into Glimmer.

- Updates to `IterableImpl`
  - Added the `IteratorDelegate` interface (could maybe use a better
    name). This is the minimal API to let Ember provide all the
    different types of iterators it needs to (in particular,
    `EmberArray`, objects with `forEach`, and `{{each-in}}` iterators).
  - No more native iterables - we don't support `Symbol` yet, plus it's
    fairly redundant with `IteratorDelegate`.
  - Very simple `IteratorWrapper` class instead, which adds the key
    to the final iteration item internally. No more need to expose
    `keyFor` concerns to Ember, except `getPath`.

- Updates to `@glimmer/runtime`
  - Added `WILL_DROP` to destroyables, and `willDestroy` to component managers

- Updates to `@glimmer/reference`
  - Removed `MapReference`
  - Removed `UpdatableReference`, replaced with `UpdatableRootReference` and
    `IterationItemReference`
  - Refactored the majority of references around two base assumptions.
    - Generally, references should refer to values in the template
    - There are fundamentally two kinds of references in the template: Roots
      and properties. Roots are unique, there are a few different types. 
      Properties are almost all the same.
  - All template references now receive the `env`, which allows them to delegate
    access, and setup debug context.

- Updates to `@glimmer/validator`
  - Added the autotracking transaction from Ember, along with an API for setting
    it up based on the embedding environment.
  - Simplified the `meta` system, now it's just a few WeakMaps.
  - Added DEBUG assertions against incorrect usage of tags, for instance 
    updating a tag with a more recent tag in a way that wouldn't bust the cache.


ember-cli-build.js
```diff
@@ -42,6 +43,8 @@ module.exports = function() {
       });
     }
     jsTree = stripGlimmerUtilities(jsTree);
+  } else {
+    jsTree = addGlimmerEnv(jsTree);
```
rwjblue: I don't think we particularly want to babel these files twice (which is what it seems like is being done here), can we add the plugin to the existing babel instead?

package.json
```diff
@@ -92,7 +93,6 @@
     }
   },
   "volta": {
-    "node": "10.16.3",
-    "yarn": "1.17.3"
```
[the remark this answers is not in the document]
pzuraq: ah, I think I was having an issue with yarn linking locally, will add it back

packages/@glimmer/reference/lib/template.ts
```diff
@@ -0,0 +1,330 @@
...
+        update(valueTag, computeTag);
+        this.computeRevision = value(tag);
+      }
+    }
+  }
+
+  compute() {
+    this.computeTag = track(() => {
+      this.computeValue = this.fn(this.args);
+    }, DEBUG && `${this.env.getTemplatePathDebugContext(this)}`);
```
rwjblue:
```suggestion
    }, DEBUG && this.env.getTemplatePathDebugContext(this));
```

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -244,29 +247,76 @@ class TransactionImpl implements Transaction {
...
-export type ToBool = (value: unknown) => boolean;
+function defaultDelegateFn(delegateFn: Function | undefined, delegateDefault: Function) {
+  let defaultFn = delegateFn !== undefined ? delegateFn : delegateDefault;
 
-function toBool(value: unknown): boolean {
-  return !!value;
+  if (DEBUG) {
+    // Bind to `null` in DEBUG since these methods are assumed to be pure
+    // functions, so we can reassign them.
+    return defaultFn.bind(null);
```
rwjblue: If this becomes an issue we could even wrap this in a proxy that asserts on `this` access, seems overkill at this point, but might eventually be useful.

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -244,29 +247,76 @@ class TransactionImpl implements Transaction {
...
+    }
   }
 
-  abstract iterableFor(reference: Reference, key: unknown): OpaqueIterable;
-  abstract protocolForURL(s: string): string;
+  iterableFor(ref: VersionedPathReference, inputKey: unknown): OpaqueIterable {
+    // TODO: We should add an assertion here to verify that we are passed a
+    // TemplatePathReference, but we can only do that once we remove
+    // or significantly rewrite @glimmer/object-reference
+    let key = inputKey === null ? '@identity' : String(inputKey);
```
rwjblue: Yes, lets please kill `@glimmer/object-reference` 🙏

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -537,13 +677,13 @@ export function JitRuntime(
...
-  document: SimpleDocument,
+export function JitRuntimeFromProgram<R, E>(
+  options: EnvironmentOptions,
   program: RuntimeProgram,
-  resolver: RuntimeResolverDelegate = {},
-  delegate: RuntimeEnvironmentDelegate = {}
-): JitRuntimeContext {
-  let env = new RuntimeEnvironment(document, new RuntimeEnvironmentDelegateImpl(delegate));
+  resolver: RuntimeResolverDelegate<R> = {},
+  delegate: EnvironmentDelegate<E> = {}
```
rwjblue: I think ideally we should try to make it so that the only thing in the system that is aware of the delegate setup that we are using _is_ the environment. I don't particularly think that will make much difference here at the moment, but it does isolate unrelated parts of the system (like this one) from _other future_ changes that we might make to the delegate.

tldr; lets pass the environment here instead of the delegate, but seems fine to do in a follow up...

packages/@glimmer/runtime/lib/lifetime.ts
```diff
@@ -1,18 +1,34 @@
...
 
-  if (linked) {
+  if (linked !== null) {
     env.didDestroy(snapshot(linked));
   }
 }
 
+export function legacySyncDestroy(parent: object, env: Environment) {
+  if (DEBUG) {
+    console.log('legacySyncDestroy', parent, LINKED.get(parent));
```
rwjblue: Is this supposed to be left in? Seems like if we hit this a lot, Ember users would be annoyed... 🤔

rwjblue: Chatted offline with @pzuraq, this `DEBUG` constant here is imported from local debug, so that isn't a large concern. Making another comment RE: changing that variable name...

packages/@glimmer/runtime/lib/lifetime.ts
```diff
@@ -1,18 +1,34 @@
...
   let linked = takeAssociated(parent);
 
-  if (linked) {
+  if (linked !== null) {
     env.didDestroy(snapshot(linked));
   }
 }
 
+export function legacySyncDestroy(parent: object, env: Environment) {
+  if (DEBUG) {
```
rwjblue: Mind changing the `DEBUG` variable to `LOCAL_DEBUG`?

packages/@glimmer/util/lib/lifetimes.ts
```diff
@@ -35,6 +38,10 @@ export function associateDestructor(parent: object, child: Drop): void {
   associated.add(child);
 }
 
+export function peekAssociated(parent: object): Option<Set<Drop>> {
+  return LINKED.get(parent) || null;
```
rwjblue: :trollface: 
```js
  return LINKED_IN.get(parent) || null;
```

packages/@glimmer/validator/lib/utils.ts
```diff
@@ -0,0 +1,10 @@
+export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends ((
+  k: infer I
+) => void)
+  ? I
+  : never;
+
+export const symbol =
```
rwjblue: Can you add a comment here about why this is duplicated (because we can't depend on `@glimmer/utils` without borking consumer types in `@glimmer/tracking`)

packages/@glimmer/reference/lib/template.ts
```diff
@@ -0,0 +1,330 @@
...
+      if (computeTag !== null) {
+        // We computed once, so setup the cache state correctly
+        update(valueTag, computeTag);
+        this.computeRevision = value(tag);
+      }
+    }
+  }
+
+  compute() {
+    this.computeTag = track(() => {
```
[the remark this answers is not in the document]
pzuraq: Yes, but that requires a more pluggable concept of Helper managers (which we definitely need to add)

PR #994, opened 2019-12-19
[BUGFIX] Avoid in operator with symbols

pzuraq:
Symbols don't work with the in operator when they are polyfilled.
This prevents us from relying on the in operator for symbols in general.

We should add a test suite in the future that includes the Babel polyfill to test this out.


rwjblue:
Released in v0.45.1

PR #998, opened 2020-01-03
ESLint: Enable `node/no-missing-import` rule

Turbo87:
see https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-missing-import.md

This rule warns us if we're importing a file or module that doesn't exist at all. This does not check for correct dependency declarations yet though.


rwjblue:
@Turbo87 - Needs a rebase.

PR #999, opened 2020-01-03
[BUG] glimmer reports possible XSS when static styles are used with element modifier applied to the same element

mhankus:
Fix for glimmer warning about possible XSS when static styles are used with element modifier applied to the same element.

- style-warnings-test.js - added test demonstrating problem
- template-compiler.ts - extended condition for trusting attributes and removed duplicate branches
- allocate-symbols.ts - my guess is that there was a typo, so I fixed it - without it, tests fail - please double check this change


mhankus:
In template-compiler.ts, I removed one branch because testing for MustacheStatement executes exactly the same code as its else branch.  It is a shame, because I was the one who made them exactly the same:
https://github.com/glimmerjs/glimmer-vm/commit/9185a3afdd54cee9aac2699363e41f2bfdd4d143#diff-c8ff1fb8d6b8ef3608c6c679c8b6282c

So now, it is only fix for my previous omission.

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -178,18 +178,13 @@ export default class TemplateCompiler implements Processor<InputOps> {
         this.opcode(['attrSplat'], action);
       } else if (isStatic && !isComponent) {
         this.opcode(['staticAttr', [name, namespace]], action);
-      } else if (isTrusting) {
+      } else if (isStatic || isTrusting) {
```
rwjblue: I don't quite think this is the same as before. Specifically, I don't think that we would inherently trust static attributes and that instead when `isStatic && isComponent` we should use `componentAttr` (vs `trustingComponentAttr`).

mhankus: Problem is that static element with modifier is treated as component (template-compiler.ts:102), so ```isStatic && isComponent``` means also static element + static styles + element modifier. So glimmer will complain about potential XSS in styles (test will fail, because static styles will not be trusted)

I can change condition, so it will be specific for styles only 
```if ((isStatic && name === 'style') || isTrusting)```
so other static attributes with isComponent will use ```componentAttr```, and static styles will use "trustingComponentAttr".

What do you think about it ?

File: package.json [hunk 2]
```diff
@@ -100,4 +100,4 @@
     "node": "10.16.3",
     "yarn": "1.17.3"
   }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
mhankus: ok - I've committed it by mistake ( .editorconfig has ```insert_final_newline = true```, and my editor just done that - I'will remove it)

mhankus:
I messed up with rebase - it is too late for me - I must go to sleep - Tomorrow I'll try to fix or close this one and open new pull request.

mhankus:
I've created new pull request https://github.com/glimmerjs/glimmer-vm/pull/1008 
It will be cleaner to understand. 
I'm closing this one. Sorry for confusion

PR #1001, opened 2020-01-03
Remove outdated `typescript` dev dependencies

Turbo87:
We're already bringing in the `typescript` dependency via `broccoli-typescript-compiler`, and the `typescript` dependencies in the packages are also on the wrong version 😅


rwjblue:
Will need a rebase now

PR #1007, opened 2020-01-08
Add a failing test for preprocess(`<TITLE></TITLE>`)

dcyriller:
Parsing `<TITLE>{{aVariable}} a title</TITLE>` with `@glimmer-vm/syntax` fails with the following error:
```
SyntaxError: Unclosed element `TITLE`
```

The error is thrown on this line: https://github.com/glimmerjs/glimmer-vm/blob/v0.45.3/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts#L59

I get the same error for `SCRIPT` and `STYLE`. Is this something expected?

### DIFF
--- a/packages/@glimmer/integration-tests/test/initial-render-test.ts
+++ b/packages/@glimmer/integration-tests/test/initial-render-test.ts
@@ -744,6 +744,28 @@ class Rehydration extends AbstractRehydrationTests {
   }
 
   @test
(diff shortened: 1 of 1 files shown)


dcyriller:
Would be fixed by https://github.com/tildeio/simple-html-tokenizer/pull/81

PR #1008, opened 2020-01-08
[BUGFIX] glimmer reports possible XSS when static styles are used with element modifier applied to the same element

mhankus:
Fix for glimmer warning about possible XSS when static styles are used with element modifier applied to the same element.

- style-warnings-test.js - added test demonstrating problem
- template-compiler.ts - extended condition for trusting static styles attributes and removed duplicate branches
- allocate-symbols.ts - my guess is that there was a typo, so I fixed it - without it, tests fail - please double check this change

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -178,16 +178,11 @@ export default class TemplateCompiler implements Processor<InputOps> {
         this.opcode(['attrSplat'], action);
       } else if (isStatic && !isComponent) {
         this.opcode(['staticAttr', [name, namespace]], action);
-      } else if (isTrusting) {
+      } else if ((isStatic && name === 'style') || isTrusting) {
         this.opcode(
           isComponent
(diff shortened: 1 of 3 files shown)


mhankus:
@rwjblue  - I created new pull request. Sorry for that, but I have no idea what I did yesterday with old pull request. 
Condition was changed, so that trustingComponentAttr is generated only for static styles attribute, for all the other attributes componentAttr is generated.
 
What do you think ?

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -178,18 +178,13 @@ export default class TemplateCompiler implements Processor<InputOps> {
         this.opcode(['attrSplat'], action);
       } else if (isStatic && !isComponent) {
```
krisselden: @rwjblue I think we should just remove the isComponent here, static attributes don't have XSS even on a component

rwjblue:
Thank you for working on this! We used these tests and implemented a slightly different fix over in https://github.com/glimmerjs/glimmer-vm/pull/1035.

PR #1009, opened 2020-01-16
[BUGFIX] Don't serialize template metas in the JIT constant pool

pzuraq:
Currently the JIT mode constant pool serializes template metas. This is
necessary for AOT mode, but is not needed in JIT mode, since templates
are compiled at runtime. Aside from being extra work, it also introduced
an issue in Ember where we could no longer correctly associate templates
with their owners.

This updates the `Constants` class to be `JitConstants` to reflect where
it should be used, and adds a pool specifically for metas.

### DIFF
--- a/packages/@glimmer/program/lib/constants.ts
+++ b/packages/@glimmer/program/lib/constants.ts
@@ -138,7 +138,9 @@ export class RuntimeConstantsImpl implements RuntimeConstants {
   }
 }
 
-export class Constants extends WriteOnlyConstants implements RuntimeConstants {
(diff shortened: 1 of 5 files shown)


packages/@glimmer/program/lib/constants.ts
```diff
@@ -153,6 +155,16 @@ export class Constants extends WriteOnlyConstants implements RuntimeConstants {
     this.others = [];
   }
 
+  templateMeta(meta: unknown): number {
+    debugger
```
rwjblue: 🏹

rwjblue:
Marking as breaking due to the export rename.

PR #1015, opened 2020-01-28
Ensure tracked initializers have access to `this`

chancancode:
See https://github.com/emberjs/ember.js/issues/18690


pzuraq:
Supercedes #1014, I think this is good though, with better types

chancancode:
Oops!! Sorry! Didn’t notice that!

PR #1022, opened 2020-02-05
[BUGFIX] Ensure Iterables work with dictionaries

pzuraq:
In the recent VM upgrade, a general API for adding debug information to
references was added. This included debug information for iteration
items, which have dynamic path segments. There were two issues with the
implementation:

1. We were using the `key` instead of the `memo` property for displaying
   which item in the collection was being used. Counterintuitively,
   `memo` is the actual key for the value in the collection, whereas
   `key` is the memoization key used to determine if the item should be
   updated. This has been updated, so that `memo` is used to show users
   where they are in the collection.

2. Since keys can be arbitrary objects, we can't always safely coerce
   them to be strings, so we have to add some debug-only logic for
   coercion.

### DIFF
--- a/packages/@glimmer/util/lib/debug-to-string.ts
+++ b/packages/@glimmer/util/lib/debug-to-string.ts
@@ -1,0 +1,25 @@
+import { DEBUG } from '@glimmer/env';
+
+let debugToString: undefined | ((value: unknown) => string);
+
+if (DEBUG) {
+  debugToString = (value: unknown) => {
+    let string;
+
+    if (typeof value === 'function') {
+      string = value.name;
+    } else if (typeof value === 'object' && value !== null) {
+      let className = (value.constructor && value.constructor.name) || '(unknown class)';
+
+      string = `(an instance of ${className})`;
+    } else if (value === undefined) {
+      string = '(an unknown tag)';
+    } else {
+      string = String(value);
+    }
+
+    return string;
+  };
+}
+
+export default debugToString;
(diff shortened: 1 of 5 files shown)


mansona:
it looks like this is passing and it's just linting that is causing CI to fail 🤔 does that mean that the bug is fixed!? 🎉

packages/@glimmer/util/lib/debug-to-string.ts
```diff
@@ -0,0 +1,25 @@
...
+      string = `(an instance of ${className})`;
+    } else if (value === undefined) {
+      string = '(an unknown tag)';
+    } else {
+      string = String(value);
+    }
+
+    return string;
+  };
+}
```
chancancode: Should this just be a no-op function in prod? The module can't be removed anyway so I don't know if it's worth making every caller care about it. If it matters (probably not), we should just make it a macro.

packages/@glimmer/util/lib/debug-to-string.ts
```diff
@@ -0,0 +1,25 @@
...
+    let string;
+
+    if (typeof value === 'function') {
+      string = value.name;
+    } else if (typeof value === 'object' && value !== null) {
+      let className = (value.constructor && value.constructor.name) || '(unknown class)';
+
+      string = `(an instance of ${className})`;
+    } else if (value === undefined) {
+      string = '(an unknown tag)';
```
chancancode: Why isn't this `"undefined"`?

packages/@glimmer/util/lib/debug-to-string.ts
```diff
@@ -0,0 +1,25 @@
...
+    } else {
+      string = String(value);
```
chancancode: try/catch?

pzuraq:
Yup, exactly. I've updated the function to be the same as Ember's implementation.

PR #1024, opened 2020-02-07
syntax/traversal: Add `parents()` generator method to the `Path` class

Turbo87:
This makes it easier to iterate over all of the parent nodes of a node

Extracted from https://github.com/ember-template-lint/ember-template-lint/pull/1106

/cc @melsumner

### DIFF
--- a/packages/@glimmer/syntax/lib/traversal/path.ts
+++ b/packages/@glimmer/syntax/lib/traversal/path.ts
@@ -15,11 +15,28 @@ export default class Path<N extends Node> {
     return this.parent ? this.parent.node : null;
   }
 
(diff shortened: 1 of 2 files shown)


Turbo87:
hmm... looks like some Babel transpilation is causing the generator method to be transpiled and now needing the regenerator runtime. do we expect this compiler code to run in the browsers too? does this prevent us from being able to use an iterable here? 🤔

Turbo87:
I may have found a way to make it work with `Symbol.iterator` instead. Let's see what CI says 😅

Turbo87:
ugh, and now IE11 doesn't know about `Symbol`... 😩

PR #1026, opened 2020-02-10
Upgrade async disk cache

chancancode:
Also ensures a trailing new line is added when rewriting package.json using `bin/publish.js`.


chancancode:
This missing trailing new lines causes a diff in the package.json file(s) after running a `yarn` command.

PR #1028, opened 2020-02-12
[CLEANUP][BREAKING] Simplifies JitRuntime, JitContext, Environment setup

pzuraq:
Currently there are three JitRuntime creation methods:

- `CustomJitRuntime`
- `JitRuntime`
- `JitRuntimeFromProgram`

With the recent refactors to Ember and Glimmer.js, we can consolidate
these to a single method. In addition, this PR updates that method to
take the `JitContext()` output directly, rather than requiring users
create a program themselves.

As part of this work, I also simplified tests to use the actual
`JitContext` method rather than making a custom one.

Finally, this PR makes `updateOperations` optional on the environment
delegate. This allows environments that do not want to update, such as
streaming SSR, to opt out of updates altogether.


rwjblue:
@pzuraq needs a rebase

PR #1034, opened 2020-02-13
[BUGFIX] Fixes {{has-block}} when used with curry components as a variable

pzuraq:
When invoked directly as a variable, curry components go through a
different codepath than they normally would. This leads to blocks and
such not being setup correctly. Since blocks aren't used in this case,
this isn't problematic in and of itself. But, the `HasBlock` opcode only
checked for truthiness of a value, and the in this case it would return
`UNDEFINED_REFERENCE` instead of `null`, thus returning truthy.

The temporary fix is to check for `UNDEFINED_REFERENCE` directly. Long
term, we should try to ensure that blocks are always setup correctly, no
matter what codepath a component is invoked by.

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/has-block.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/has-block.ts
@@ -262,4 +262,70 @@ export class HasBlockSuite extends RenderTest {
     this.assertComponent('', { 'data-has-block': 'false' });
     this.assertStableRerender();
   }
+
+  @test({ kind: 'glimmer' })
+  'has-block works within a yielded curried component invoked within mustaches'() {
+    this.registerComponent(
+      'Glimmer',
+      'ComponentWithHasBlock',
+      `<div data-has-block="{{has-block}}" ...attributes></div>`
+    );
+
+    this.registerComponent('Glimmer', 'Yielder', `{{yield (component 'ComponentWithHasBlock')}}`);
+
+    this.registerComponent(
+      'Glimmer',
+      'TestComponent',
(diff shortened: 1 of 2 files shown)


wycats:
lgtm

Thanks so much for the detailed comment!

packages/@glimmer/integration-tests/lib/suites/has-block.ts
```diff
@@ -262,4 +262,48 @@ export class HasBlockSuite extends RenderTest {
     this.assertComponent('', { 'data-has-block': 'false' });
     this.assertStableRerender();
   }
+
+  @test({ kind: 'glimmer' })
+  'has-block works within a yielded curried component invoked as variable'() {
```
rwjblue: "invoked as a variable" doesn't really make sense to me in this context. Maybe:

```suggestion
  'has-block works within a yielded curried component invoked within mustaches'() {
```

packages/@glimmer/integration-tests/lib/suites/has-block.ts
```diff
@@ -262,4 +262,48 @@ export class HasBlockSuite extends RenderTest {
...
+    );
+
+    this.render(`<TestComponent />`);
+
+    this.assertComponent('', { 'data-has-block': 'false' });
+    this.assertStableRerender();
+  }
+
+  @test({ kind: 'glimmer' })
+  'has-block works within a yielded curried component invoked using standard syntax'() {
```
rwjblue: What is "standard syntax"?!

```suggestion
  'has-block works within a yielded curried component invoked via self-closing angle bracket invocation'() {
```

rwjblue:
CI is fixed by #1036

PR #1036, opened 2020-02-13
IE11 doesn't support Number.MAX_SAFE_INTEGER / Number.MIN_SAFE_INTEGER

rwjblue:


chancancode:
Hmm.. the tests in my Ember PR also used these constants. Why didn't the test fail there?

rwjblue:
@chancancode not sure, will dig to find out what is going on over there (but this fix is definitely fixing the CI failures here)

chancancode:
ahhh, I think due to the way my code was written, they are just not getting run on IE11:

https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/tests/integration/content-test.js#L43-L55

`while (i > 1)` === `while (undefined > 1)` === `while (false)`.

Better than an infinite loop, but definitely not what I intended!

PR #1041, opened 2020-02-14
runtime: Add missing `@glimmer/opcode-compiler` dependency

Turbo87:
We are importing from `@glimmer/opcode-compiler` inside the `lib` folder of the `@glimmer/runtime` package, so we need to add this as a dependency.

### DIFF
--- a/packages/@glimmer/runtime/package.json
+++ b/packages/@glimmer/runtime/package.json
@@ -7,6 +7,7 @@
   "dependencies": {
     "@glimmer/env": "0.1.7",
     "@glimmer/low-level": "^0.47.9",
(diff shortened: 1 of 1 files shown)


lifeart:
https://github.com/glimmerjs/glimmer.js/pull/234

rwjblue:
@lifeart - that PR is about compiler + opcode-compiler, not runtime. Also, this PR doesn’t actually change anything about the compiled output. It just makes the dependency relationship that already exists (the code in @glimmer/runtime’s lib/**/*.ts already directly requires from @glimmer/opcode-compiler) explicit.

krisselden:
yes but the imports are only types/or stripped out debug stuff, it isn't intentional

PR #1042, opened 2020-02-14
 ESLint: Enable `node/no-unpublished-import` rule 

Turbo87:
see https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-import.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -227,7 +227,6 @@ module.exports = {
       },
       rules: {
         'node/no-extraneous-import': 'off',
-        'node/no-unpublished-import': 'off',
         'node/no-unsupported-features/es-syntax': 'off',
         'node/no-unsupported-features/node-builtins': 'off',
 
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   plugins: ['@typescript-eslint', 'prettier', 'qunit'],
   settings: {
     node: {
+      allowModules: ['@glimmer/debug', '@glimmer/local-debug-flags'],
```
rwjblue: FWIW, these _are_ actual modules that can be resolved. Why do we have to explicitly whitelist them?

rwjblue: Is it because those packages aren’t published (and therefore not listed in the respective package.json dependencies)?

Turbo87: it's imported in a few lib folders, so it would have to be a non-dev dependency, but since it's a private package that won't work. best solution I could come up with was this one.

rwjblue: Ya, makes sense. We avoided this with `@glimmer/env` by actually publishing...

PR #1043, opened 2020-02-14
Fix @glimmer/interfaces dependency on @glimmer/runtime

krisselden:


rwjblue:
@Turbo87 - this helps reduce more of those circular deps you were looking into

Turbo87:
nice, thanks!

PR #1046, opened 2020-02-17
[CLEANUP] Fix compiler typing

krisselden:
Improved compiler types and fixes performance issues with type checking opcode-compiler statements and expressions that get even worse if you upgrade past typescript 3.6.

The Compiler implementation is shared since it is the same and only varies by generics now (as I think was originally intended).

The mapping types are now derived from Statement and TupleExpression, so they are related and stay in sync.

Also, move const ATTR_BLOCK out of compiler since it wasn't used there nor seemed related since it is only used in the component helper and WrappedComponent.   Moved to component helper which uses it and WrappedComponent uses it but already was dependent on the component helper.

### DIFF
--- a/packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
+++ b/packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
@@ -49,43 +49,17 @@ export const enum SexpOpcodes {
   Concat = 32,
 }
 
-export interface SexpOpcodeMap {
-  [index: number]: TupleSyntax;
+export type StatementSexpOpcode = Statement[0];
+export type StatementSexpOpcodeMap = {
+  [TSexpOpcode in Statement[0]]: Extract<Statement, { 0: TSexpOpcode }>;
+};
+export type ExpressionSexpOpcode = TupleExpression[0];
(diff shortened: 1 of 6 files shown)


packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
```diff
@@ -103,8 +77,7 @@ export const enum ExpressionContext {
   // An `Append` is a single identifier that is contained inside a curly (either in a
   // content curly or an attribute curly)
   AppendSingleId = 'AppendSingleId',
-  // An `Expression` is evaluated into a value (e.g. `person.name` in `(call person.name)`
-  // or `person.name` in `@name={{person.name}}`). This represents a syntactic position
+  // An `Expression` is evaluated into a value (e.g. `person.name` in `(call person.name)` // or `person.name` in `@name={{person.name}}`). This represents a syntactic position
```
rwjblue: This is a bit odd, to have two sets of `//` on a line

krisselden: this was done with format on save.

PR #1048, opened 2020-02-18
[CLEANUP] Cleanup build deps

krisselden:


rwjblue:
Wow, surprised we still had this hanging around!

PR #1049, opened 2020-02-19
Enable eslint-plugin-node's no-extraneous-import rule (mostly).

rwjblue:
This enables eslint-plugin-node's `no-extraneous-import` rule (which is part of its recommended set) for all packages _except_ `@glimmer/interfaces`.

I've excluded `@glimmer/interfaces` specifically because IMHO it **should not** have any external dependencies, and the imports here need to be refactored to avoid needing to add the dependency.

This change however helps us ensure we don't _accidentally_ regress for all of the other packages...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -226,7 +226,6 @@ module.exports = {
         tsconfigRootDir: __dirname,
       },
       rules: {
-        'node/no-extraneous-import': 'off',
         'node/no-unsupported-features/es-syntax': 'off',
         'node/no-unsupported-features/node-builtins': 'off',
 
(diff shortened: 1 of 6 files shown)


packages/@glimmer/program/package.json
```diff
@@ -6,5 +6,8 @@
     "@glimmer/util": "^0.47.9",
     "@glimmer/interfaces": "^0.47.9",
     "@glimmer/encoder": "^0.47.9"
+  },
+  "devDependencies": {
+    "@glimmer/opcode-compiler": "^0.47.9"
```
rwjblue: Must remove this one, or we have cyclic errors during `yarn build`.

rwjblue:
CI is green, but the GH status isn't updated, merging....

PR #1050, opened 2020-02-20
Remove unused and update eslint and prettier.

krisselden:


package.json
```diff
@@ -24,6 +24,7 @@
     "test:types": "node bin/run-types-tests.js",
     "test:smoke": "SMOKE_TESTS=true ember test",
     "eslint": "eslint . --cache --ext .js,.ts",
+    "eslint:fix": "eslint . --cache --ext .js,.ts --fix",
```
rwjblue: Seems fine, but I personally prefer to rely on `yarn eslint --fix` (which works already)

PR #1053, opened 2020-02-21
Upgrade TypeScript version to 3.7.5

krisselden:


package.json
```diff
@@ -90,7 +90,7 @@
     "semver": "^6.3.0",
     "testem-failure-only-reporter": "^0.0.1",
     "toml": "^3.0.0",
-    "typescript": "3.5.3"
+    "typescript": "3.7.5"
```
rwjblue: Should we use `~3.7.5`?

ef4:
Does this mean https://github.com/glimmerjs/glimmer-vm/issues/946 is fixed?

PR #1054, opened 2020-03-04
[FEAT] Adds Owner API

pzuraq:
This PR upstreams the Owner API from Ember.js, for use in embedding
environments. The public component/modifier managers, and future
managers, that both Ember.js and Glimmer.js currently provide all have a
notion of an "owner" which is passed when creating the manager, so it is
already part both embedding environments.

Conceptually, the owner is an opaque value that the embedding
environment decides on how to handle. But for there to be interop
_between_ embedding environments, especially for _libraries_ like
`@glimmer/component`, etc. It's important to have a shared way to assign
and access the owner, even if the value is opaque to the rest of the
framework. These methods provide that.


pzuraq:
This has been superceded

PR #1057, opened 2020-03-12
[PERF] Improves the autotracking transaction by creating fewer Errors

pzuraq:
Currently we create an Error every time a value is consumed, regardless
of whether or not it is being consumed for the first time (and thus we
need the Error for its stack) or if we've already consumed it before,
and don't need an Error at all. This PR changes this so we do it only
when necessary.

This helps to alleviate some dev-mode performance issues we've been seeing in Ember: https://github.com/emberjs/ember.js/issues/18798. Creating the error at all seems to be causing slow down, but this makes the issues noticeably better.

### DIFF
--- a/packages/@glimmer/validator/lib/debug.ts
+++ b/packages/@glimmer/validator/lib/debug.ts
@@ -22,7 +22,7 @@ export let assertTagNotConsumed:
   | undefined
   | (<T>(tag: Tag, obj?: T, keyName?: keyof T | string | symbol, forceHardError?: boolean) => void);
 
-export let markTagAsConsumed: undefined | ((_tag: Tag, sourceError: Error) => void);
+export let markTagAsConsumed: undefined | ((_tag: Tag, sourceError?: Error) => void);
 
 if (DEBUG) {
   let DEPRECATE_IN_AUTOTRACKING_TRANSACTION = false;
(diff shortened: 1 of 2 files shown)


pzuraq:
@rwjblue the algorithm is recursive, so when it begins calling itself again it passes the same error down to prevent creating more errors (with deeper and deeper call stacks)

rwjblue:
> The idea is that if you have used it in multiple locations, once you fix the first one, you'll see the second, and so on.

Yeah, but this sort of whack-a-mole is pretty frustrating, especially when the the framework _could have_ told you all of the locations the first time.

rwjblue:
> Showing more than 2 error stacks seems like it would be more confusing, IMO

I think we can fix this instead of memoizing the consumption error.

pzuraq:
Added a note about the performance impact to the motivation

rwjblue:
We chatted a bit about this offline yesterday. I think we should guard the error creation completely behind a flag that can be enabled via `config/environment.js` (e.g. the `EmberENV` config there). This _likely_ means threading something through via the host, but hopefully that isn't too bad.

I'd prefer to not loose the ability to capture stacks completely (I'd like to land a PR making it host configurable instead of removing completely), but if that is not tenable we can consider removing all together first then bringing back as an opt-in.

PR #1059, opened 2020-03-22
[REFACTOR] Removes legacy test patterns and packages

pzuraq:
This refactor primarily targets a few legacy tests that have not been
updated in some time, and don't use the same test harness that the rest
of the tests do. This enables us to completely drop the alternative test
harness, and prepares us for future refactors by having a more unified
test system.

In addition, this PR drops the `@glimmer/object` and
`@glimmer/object-reference` packages. Most of the last usages were
removed, and these packages are outdated and not actually used in
practice, so it makes sense to remove them now.


pzuraq:
Thanks for that! 😄

PR #1060, opened 2020-03-22
 [FEAT] Adds `memoizeTracked`

pzuraq:
Adds `memoizeTracked`, which is meant to clean up autotracking patterns
in general by simplifying them. Also introduces beginTrackFrame and
endTrackFrame for usage within the VM.

`memoizeTracked` has the following signature:

```ts
function memoizeTracked<T, Args>((...args: Args) => T): (...args: Args) => T;
```

It essentially receives a function, and returns the same function but memoized based on tracking. If you consume a tag within that function, it will memoize based on that tag, and will only update if that tag is dirtied. Otherwise, it returns the previous result.

```ts
let count = 0;

let tag = createTag();

let counter = memoizeTracked(() => {
  count++;
  consumeTag(tag);
});

counter(); // count is 1
counter(); // count still 1
dirtyTag(tag);
counter(); // count is 2
```

Depends on #1063

### DIFF
--- a/packages/@glimmer/integration-tests/lib/components/emberish-curly.ts
+++ b/packages/@glimmer/integration-tests/lib/components/emberish-curly.ts
@@ -22,7 +22,7 @@ import {
 } from '@glimmer/interfaces';
 import { Attrs, AttrsDiff } from './emberish-glimmer';
 import { VersionedPathReference, PathReference } from '@glimmer/reference';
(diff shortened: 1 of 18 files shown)


rwjblue:
Can you update the PR description to explain what the interface is and how you expect it to be used (generally speaking)?

pzuraq:
Ah, sorry about that, will definitely break things up a bit more next time!

PR #1061, opened 2020-03-22
[FEAT] Adds Effect Queue

pzuraq:
This PR prepares the VM for a refactor toward autotracking by extracting
modifiers and wrapping them in "effects". This is short for the term
[side-effect](en.wikipedia.org/wiki/Side_effect_(computer_science)), which describes an action run within a function or computation that
doesn't technically have anything to do with the _output_ of that
computation. Modifiers are a form of side-effect, because they don't
directly affect the output of the VM - they schedule a change that
occurs _later_.

This also presented a problem for autotracking, which is why I started
with this PR. Autotracking operates on the premise that it is watching
state used within a computation - i.e. _synchronously_. Async actions,
like modifiers, violate the basic premise of autotracking. We were able
to model it with Tags using UpdatableTag, which does allow this type of
lazy structure, but updatable tags have been a major source of
performance issues and bugs in general.

Rather than having the VM be responsible for updating these async
actions, the way to handle this with autotracking would be to have
something _else_ handle it, specifically the place where the actual
computations _occur_. This is the Effect Queue.

## How it works

The VM handles registration and destruction of effects, but the
environment is now responsible for updating them. Effects are scheduled
to run in the same position modifiers were previously, and all effects
in the queue are revalidated after each render.

Ordering is also still guaranteed, in that child effects will always be
triggered before parents. This is because new nodes are prepended to the
tree, and thus any new children are guaranteed to run before their older
parents. Due to the nature of the queue, it's possible that sibling
update order could change and be unpredictable.

## Other options

We could use a tree that mirrors the VM tree instead of a queue, but
traversing that tree could end up being fairly expensive, especially if
there are few modifiers. I opted for the simpler solution for the time
being, and figured we could benchmark to see if there is a performance
impact currently, and if so what solutions are better.

Another option would be to make a n-ary tree that simply divides the
effects up evenly in memoization chunks. This might allow us to get some
wins in the general case, when most modifiers have not changed.

Depends on #1060


rwjblue:
CI failure is linting related:

```
$ eslint . --cache --ext .js,.ts

/home/runner/work/glimmer-vm/glimmer-vm/packages/@glimmer/integration-tests/test/modifiers-test.ts
  1:9  error  Replace `·RenderTest,·jitSuite,·test,·Count,·tracked,·EmberishGlimmerComponent,·EmberishGlimmerArgs·` with `⏎··RenderTest,⏎··jitSuite,⏎··test,⏎··Count,⏎··tracked,⏎··EmberishGlimmerComponent,⏎··EmberishGlimmerArgs,⏎`  prettier/prettier
```

packages/@glimmer/runtime/lib/effects.ts
```diff
@@ -0,0 +1,122 @@
...
+  effects: LinkedList<ListNode<Effect>> = new LinkedList();
+
+  /**
+   * Tracker for the current head of the queue. This is used to coordinate
+   * adding effects to the queue. In a given render pass, all effects added
+   * should be added in the order they were received, but they should be
+   * _prepended_ to any pre-existing effects. For instance, let's say that the
+   * queue started off in this state after our first render pass:
+   *
+   *   A, B, C
```
pzuraq, line 51: Add an ASCII graph for the tree of nodes

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -44,137 +39,19 @@ import {
...
+import { ConditionalReference } from './references';
 import { DynamicAttribute, dynamicAttribute } from './vm/attributes/dynamic';
 import { RuntimeProgramImpl } from '@glimmer/program';
-
-export function isScopeReference(s: ScopeSlot): s is VersionedPathReference {
-  if (s === null || Array.isArray(s)) return false;
-  return true;
-}
-
-export class ScopeImpl<C extends JitOrAotBlock> implements PartialScope<C> {
```
pzuraq, line 56: This was moved to `scope.ts` since the environment.ts file was getting large and covering too many concerns.

packages/@glimmer/integration-tests/test/modifiers-test.ts
```diff
@@ -302,6 +302,72 @@ class ModifierTests extends RenderTest {
...
+
+    this.render('<Foo/>');
+
+    assert.deepEqual(inserts, ['outer'], 'outer modifier insert called');
+    assert.deepEqual(updates, [], 'no updates called');
+
+    component!.showing = true;
+    this.rerender();
+
+    assert.deepEqual(inserts, ['outer', 'inner'], 'inner modifier insert called');
```
pzuraq, line 361: Should reset inserts between each assertion

packages/@glimmer/runtime/lib/effects.ts
```diff
@@ -0,0 +1,122 @@
...
+   * their parents. By keeping track of the current head at the beginning of a
+   * transaction, we can insert new effects in the proper order during the
+   * transaction.
+   */
+  currentHead: Option<ListNode<Effect>> = null;
+
+  revalidate = () => this.effects.forEachNode(n => n.value.createOrUpdate());
+}
+
+export class EffectManager {
```
pzuraq, line 70: Rename to EffectsManager

packages/@glimmer/runtime/lib/effects.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (DEBUG) {
+      this.inTransaction = false;
+    }
+
+    let { queues, scheduleEffects } = this;
+
+    for (let phase of effectPhases) {
+      let queue = queues[phase];
+
+      scheduleEffects(phase, queue.revalidate);
```
pzuraq, line 117: Add a unit test to ensure that `scheduleEffects` can be called asynchronously and the rest of the EffectManager continues to work properly.

packages/@glimmer/runtime/lib/environment.ts
pzuraq: Will break this out into a separate PR

packages/@glimmer/runtime/lib/effects.ts
```diff
@@ -0,0 +1,122 @@
...
+  update(): void;
+  teardown(): void;
+}
+
+export class EffectImpl implements Effect {
+  constructor(private hooks: EffectHooks) {}
+
+  private didSetup = false;
+
+  createOrUpdate = memo(() => {
```
pzuraq, line 20: Update this to `run`, and update ModifierManager to just include `run`

pzuraq:
Superceded by #1310

PR #1062, opened 2020-03-22
[REFACTOR] Refactors the VM to use autotracking

pzuraq:
This PR refactors the VM to use autotracking internally, rather than
manually coalescing tags. This effectively means removing the
`Tagged`, `VersionedReference`, and `VersionedPathReference` interfaces,
and reorienting the VM around autotracking instead.

Autotracking frames have been wrapped around memoization points in the
VM, and new updating opcodes have been creating to start and end track
frames. Tags are still used, but more rarely now, since they aren't as
necessary when using autotracking instead.

Depends on #1061


packages/@glimmer/reference/lib/iterable.ts
```diff
@@ -234,7 +217,7 @@ export interface IteratorSynchronizerDelegate<Env> {
     before: Option<unknown>
   ): void;
   delete(env: Env, key: unknown): void;
-  done(env: Env): void;
+  // done(env: Env): void;
```
rwjblue: What is going on here? Do we need to keep this comment?

packages/@glimmer/reference/lib/iterable.ts
```diff
@@ -278,7 +261,7 @@ export class IteratorSynchronizer<Env> {
           phase = this.nextPrune();
           break;
         case Phase.Done:
-          this.nextDone();
+          // this.nextDone();
```
rwjblue: Still keeping this, why?

guides/05-validators.md
```diff
@@ -481,15 +481,15 @@ reference chain.
 For example, this is the updated implementation of `UppercaseReference`:
 
 ```typescript
-interface VersionedReference<T> extends Reference<T> {
+interface Reference<T> extends Reference<T> {
```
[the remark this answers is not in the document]
pzuraq: mm, this was a find and replace so definitely missed it

PR #1064, opened 2020-03-24
[CHORE] Fixes local linking

pzuraq:
Fixes local linking by symlinking the `dist` folder to itself in a fake `node_modules` folder. Also fixes the dev dependencies of various packages.


locks:
Rebased to resolve conflicts.

PR #1066, opened 2020-03-25
[PERF] Removes the Errors created for autotracking transactions

pzuraq:
These errors are expensive to generate, and have become problematic for
development mode in general. The plan is to remove them for now, and try
to bring them back some time in the future in a opt-in way that is more
conventional and comprehensive (e.g. show all possible consumption
locations, not just the first one).

Related: https://github.com/emberjs/ember.js/issues/18834
Related: https://github.com/emberjs/ember.js/issues/18798

### DIFF
--- a/packages/@glimmer/validator/lib/debug.ts
+++ b/packages/@glimmer/validator/lib/debug.ts
@@ -3,7 +3,6 @@ import { DEBUG } from '@glimmer/env';
 
 interface AutotrackingTransactionSourceData {
   context?: string;
-  error: Error;
 }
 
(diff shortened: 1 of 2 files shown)


bendemboski:
🎉 porting some stuff to tracked has totally murdered my performance in tests, and I've verified that this is the cause. I have a test module that takes 110s to run, or 2s to run with the error construction removed. Any idea when this will be released? Trying to decide whether or not to hold up my porting to tracked until this is released so dev cycle, CI time, etc., doesn't skyrocket.

PR #1069, opened 2020-04-11
[BUGFIX] Fix untrack logic

pzuraq:
The current logic for `untrack` can cause the tracking frames to become
lopsided. This updates it to ensure that we count untracks and empty
track frames on the autotracking frame stack.


pzuraq:
Yes, I ran them without the fix and they failed

rwjblue:
Can you cross link the issues this addresses in Ember? IIRC, there was something reported about this.

PR #1070, opened 2020-04-14
Failing test for destroy not called on items in each

scottmessinger:
If I understand the problem correctly, I think this is a failing test for https://github.com/emberjs/ember.js/issues/18855

Let me know if there's anything I can do to help!

### DIFF
--- a/packages/@glimmer/integration-tests/test/updating-test.ts
+++ b/packages/@glimmer/integration-tests/test/updating-test.ts
@@ -1893,8 +1893,11 @@ class UpdatingTest extends RenderTest {
     this.rerender({ list: ['initial', 'update'] });
     this.assertHTML(`<div>initial</div><div>update</div>`);
 
+    this.rerender({ list: ['initial'] });
+    assert.equal(destroyCount, 1, 'destroy was called for removed item');
+
     this.rerender({ list: [] });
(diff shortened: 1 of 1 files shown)


packages/@glimmer/integration-tests/test/updating-test.ts
```diff
@@ -1893,8 +1893,11 @@ class UpdatingTest extends RenderTest {
     this.rerender({ list: ['initial', 'update'] });
     this.assertHTML(`<div>initial</div><div>update</div>`);
 
+    this.rerender({ list: ['initial'] });
+    assert.equal(destroyCount, 1, 'destroy was called for removed item');
+
     this.rerender({ list: [] });
-    assert.equal(destroyCount, 2, 'both list items were correctly destroyed');
+    assert.equal(destroyCount, 1, 'remaining list item was correctly destroyed');
```
rwjblue, line 1900: I believe this should still be 2:

```suggestion
    assert.equal(destroyCount, 2, 'remaining list item was correctly destroyed');
```

scottmessinger: You're totally right. Fixed and rebased.

rwjblue:
OK, I think this should pass now if you rebase...

scottmessinger:
@rwjblue Rebased and tests now pass!

PR #1071, opened 2020-04-15
@glimmer/syntax: Bump minimum handlebars.js version

dcyriller:
..from 4.5.1 to 4.7.4

In v4.7.4, handlebars.js dropped `optimist` dependency. This `optimist`
dep had a dependency on an outdated version of `minimist` affected by
CVE-2020-7598.

This will remove a Github security alert in packages depending on
@glimmer/syntax.

For the sake of consistency, I bumped the main the version in the main
package.json as well. Not sure if it is legit though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@simple-dom/void-map": "^1.4.0",
     "babel-plugin-debug-macros": "^0.3.3",
     "fs-extra": "^9.0.0",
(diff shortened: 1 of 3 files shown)


rwjblue:
Thank you @dcyriller!

rwjblue:
Published as v0.50.3.

PR #1075, opened 2020-04-16
Update main key in package json to reference es2017 bundles

chiragpat:
**Why**
Currently when we try to use glimmer-vm in a node env it pulls in the es5 bundles instead of the es2017 bundle. [node > 10 has 100% support for es2017](https://node.green/#ES2017) so we should be able to just pull that in and don't have to run any of the polyfilled features which could have runtime overahead.

**How**
- Update path in write-package-json build script


chadhietala:
@rwjblue I think this is going to be fine because Ember is going to rollup and transpile anyways but want to confirm before merging. Should we rev the minor in the release?

rwjblue:
Yep, seems good to me but labeling as "breaking".

rwjblue:
Published as v0.51.0.

PR #1080, opened 2020-04-29
Ensure destructors run for items added to an `{{#each` after initial render

pzuraq:
Currently we don't properly setup the `DESTRUCTOR_STACK` on the VM
before beginning the update when we add a new list item. This refactor
renames the `iterate` function to replace `enterItem`, since that's
where the majority of setup code exists, and renames `enterItem` to be
`registerItem`, since that's the only part of the code that really
differs between append and update. It also adds a test, which I ensured
did fail before this update.

### DIFF
--- a/packages/@glimmer/integration-tests/test/updating-test.ts
+++ b/packages/@glimmer/integration-tests/test/updating-test.ts
@@ -1863,7 +1863,7 @@ class UpdatingTest extends RenderTest {
   }
 
   @test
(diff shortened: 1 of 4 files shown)


pzuraq:
Original Ember issue: https://github.com/emberjs/ember.js/issues/18855#issuecomment-607470176

pzuraq:
Looks like it was a fluke, originally it passed and linting failed, I reran and it passed.

PR #1081, opened 2020-04-29
Fix key uniqueness implementation

pzuraq:
Bugfix for the iteration's key uniqueness function, was passing the
value in instead of the key derived from the value. Added a test that
fails if keys are not correctly unique, originally we weren't testing
this because IDENTITY returns the value it was passed in all cases
except `null`.


pzuraq:
Original Ember issue: https://github.com/emberjs/ember.js/issues/18864

PR #1083, opened 2020-05-03
Human readible(-ish) wireformat compilation 

chancancode:
Changes the output of `precompile`.

Before:

```js
{"id":null,"meta":{},"block":"{\"symbols\":[],\"hasEval\":false,\"upvars\":[\"obj\"],\"statements\":[[10,\"input\"],[15,2,[34,0]],[12],[13]]}"}
```

After (DEBUG):

```js
{
  id: null,
  meta: {},
  block: () => {
    /*

    Source:

    ```hbs
    <input value={{obj}} />
    ```

    Transformed:

    ```hbs
    <input value={{obj}} />
    ```

    */

    /* Wire Format */

    return {
      symbols: [],
      hasEval: false,
      upvars: [
        "obj"
      ],
      statements: [
        // ["open-element","input"]
          [10,"input"],
        // ["dynamic-attr","value",["get-free-in-append-single-id","obj"],null]
          [15,2,[34,0]],
        // ["flush-element"]
          [12],
        // ["close-element"]
          [13]
      ]
    }
  }
}
```

After (non-DEBUG):

```js
{
  id: null,
  meta: {},
  block: () => {
    /*

    Source:

    ```hbs
    <input value={{obj}} />
    ```

    Transformed:

    ```hbs
    <input value={{obj}} />
    ```

    WireFormat:

    ```js
    {
      symbols: [],
      hasEval: false,
      upvars: [
        "obj"
      ],
      statements: [
        // ["open-element","input"]
          [10,"input"],
        // ["dynamic-attr","value",["get-free-in-append-single-id","obj"],null]
          [15,2,[34,0]],
        // ["flush-element"]
          [12],
        // ["close-element"]
          [13]
      ]
    }
    ```

    */

    return JSON.parse("{\"symbols\":[],\"hasEval\":false,\"upvars\":[\"obj\"],\"statements\":[[10,\"input\"],[15,2,[34,0]],[12],[13]]}");
  }
}
```

After (non-DEBUG, minified):

```js
{"id":null,"meta":{},"block":()=>JSON.parse("{\"symbols\":[],\"hasEval\":false,\"upvars\":[\"obj\"],\"statements\":[[10,\"input\"],[15,2,[34,0]],[12],[13]]}")}
```

Implications and considerations:

* The output of `precompile` is no longer `JSON.parse`-able. However, this is needed for strict mode anyway, and the documentation has always recommended using `new Function` to parse the output.

* Slightly inflated WF size in non-DEBUG + minified: each template now adds an additional 16 ASCII bytes, `()=>JSON.parse(${WF})` in the "blocks" field. I don't imagine this to be a big deal, as it should be pretty GZip-able? If it is a problem, we can pass in `JSON.parse`, so `p=>p(${WF})`?

* Not sure how changing the "block" field from a string to a function would impact parse time. Would we still need the `JSON.parse` trick in non-DEBUG mode due to lazy parsing? Could we just leave the JS syntax as-is inside the function body, like what I did in the DEBUG version?


chancancode:
IE 11 tests are failing because the `precompile` output uses arrow functions which is not supported. Should I switch it to use `function() { ... }`?

rwjblue:
> Slightly inflated WF size in non-DEBUG + minified: each template now adds an additional 16 ASCII bytes, ()=>JSON.parse(${WF}) in the "blocks" field. I don't imagine this to be a big deal, as it should be pretty GZip-able? If it is a problem, we can pass in JSON.parse, so p=>p(${WF})?

Can we make the `JSON.parse` on the caller side (e.g. if a string is returned call JSON.parse)?

rwjblue:
> IE 11 tests are failing because the precompile output uses arrow functions which is not supported. Should I switch it to use function() { ... }?

Yes

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -53,25 +54,200 @@ const defaultOptions: PrecompileOptions = {
...
+
+  let func = new FunctionCompiler(obj.indentation);
+
+  func.push('/*');
+  func.push();
+
+  func.push('Source:');
+  func.push();
+  func.push('```hbs');
+  func.push(source);
```
rwjblue, line 90: This source needs to be escaped, it can **absolutely** contain `*/` in it...

Consider compiling the following template:

```hbs
LOL */ you screwed now dude
```

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -53,25 +54,200 @@ const defaultOptions: PrecompileOptions = {
...
+  func.push('```');
+  func.push();
+
+  func.push('Transformed:');
+  func.push();
+  func.push('```hbs');
+  func.push(print(ast));
+  func.push('```');
+  func.push();
```
rwjblue, line 99: We should probably only emit both if the `print(parse(source)) !== print(ast)` or something 🤔

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -53,25 +54,200 @@ const defaultOptions: PrecompileOptions = {
...
+
+  let stmts = new WireFormatArrayCompiler(wf.indentation);
+
+  blockJSON.statements.forEach(stmt => {
+    stmts.add(stmt, debug.formatOpcode(stmt));
+  });
+
+  wf.setRaw('statements', stmts.compile());
+
+  if (!DEBUG) {
```
rwjblue, line 115: Instead of using the imported DEBUG, can you add an option? This would allow us to compile both production and development templates from the same actual build (and is how I intend to "fix" the AST transforms in Ember to avoid debug only AST transforms from hitting prod).

packages/@glimmer/compiler/lib/compiler.ts
chancancode: I thought about it but it seemed tricky due to imperfect serialization, but I guess it's still worth doing if it eliminates a good portion of them

rwjblue: Well, since we are already printing the transformed output it seems fine to do a quick `source !== transformedOutput` (even if most are different). Also, the `print(parse(source))` bit (vs using original source) was trying to avoid imperfect serialization issues.

PR #1086, opened 2020-05-06
[BUGFIX] Process in-element in compiler not parser

gitKrystan:
This fixes the issue described in https://github.com/DockYard/ember-maybe-in-element/pull/25#issuecomment-622506443

tl;dr If an AST transform introduces the element, it doesn't go through the parser, so it's missing the required `guid` and `insertBefore` normalization. This commit moves the processing to a later stage to avoid this issue.

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -18,8 +18,6 @@ function isTrustedValue(value: any) {
   return value.escaped !== undefined && !value.escaped;
 }
 
-export const THIS = 0;
-
 export default class TemplateCompiler implements Processor<InputOps> {
   static compile(ast: AST.Template, source: string, options?: CompileOptions): Template {
     let templateVisitor = new TemplateVisitor();
@@ -49,6 +47,12 @@ export default class TemplateCompiler implements Processor<InputOps> {
   private locations: Option<SourceLocation>[] = [];
   private includeMeta = true;
 
+  private cursorCount = 0;
+
+  cursor() {
+    return `%cursor:${this.cursorCount++}%`;
+  }
+
   process(
     actions: Action[]
   ): { opcodes: readonly Ops<AllocateSymbolsOps>[]; locations: readonly Option<SourceLocation>[] } {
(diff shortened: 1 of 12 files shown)


packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -62,6 +66,7 @@ export default class TemplateCompiler implements Processor<InputOps> {
   }
 
   startProgram([program]: [AST.Template]) {
+    this.cursorCount = 0;
```
rwjblue, line 69: ~~I don't fully understand why we are resetting this counter for each program. I would have assumed that it was important that the `guid` we used was unique across each template (or is it needing to be more globally unique?), but resetting in `startProgram` here will lead to the same template reusing the same cursorCount (and therefore `this.cursor()` not being unique across the template).~~

D'oh! I just realized (after digging more into the code) that `startProgram` here is super misleading. It _really_ should have been renamed to `startTemplate` when we split the usages of `Program` AST nodes into `Template` and `Block`.

I do still wonder what level of uniqueness we are looking for with the guid, but that is somewhat unrelated.

rwjblue: Somewhat ironically, the _original_ code did suffer from the problem I described since it is operating on the `Handlebars.Program` (which is definitely emitted for each branch of a block). 

See [this AST Explorer](https://astexplorer.net/#/gist/16cff7a9560a43d914d289168c988ed1/) showing that multiple `Program` nodes would have been seen in the `HandlebarsNodeVisitors` and therefore the same cursor would be reused within a template like 

```hbs
{{#if foo}}
  {{#in-element this.el}}{{/in-element}}
{{else}}
  {{#in-element this.el}}{{/in-element}}
{{/if}}
```

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -18,8 +18,6 @@ function isTrustedValue(value: any) {
   return value.escaped !== undefined && !value.escaped;
 }
 
-export const THIS = 0;
```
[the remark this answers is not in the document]
chancancode: Nope, it was dead code. Unless it’s some kind of public API but then since the entire codebase doesn’t do anything with it I don’t know what that would do either.

rwjblue:
Thank you @gitKrystan!

PR #1088, opened 2020-05-10
[REFACTOR] Rename memoizeTracked to memo

pzuraq:
Since this is no longer going to be the public API, rename it to be
an internal API with a shorter name (less bytes, easier to remember


rwjblue:
Marking as breaking, since it renames an export from the main entry point.

PR #1090, opened 2020-05-10
[FEATURE] Adds `cache` api

pzuraq:
Adds the cache API specified in [RFC 615](https://github.com/emberjs/rfcs/blob/master/text/0615-autotracking-memoization.md).

There were a few minor refactors in addition to this:

- Moved `trackedData` into a separate module, since the `tracking`
  module was getting pretty big.

- Renamed `isConst` to `isConstTagged`, since there is a naming conflict
  with the new `isConst` function. Given `isConstTagged` will be removed
  soon and is not generally much used outside of the VM, I thought it
  was better to prioritize the new API getting this name.

- Rewrote the `memoizeTracked` implementation on top of the cache APIs.
  One thing worth noting is that in order to support the API
  `memoizeTracked` had before, `getValue` also needed to be able to
  accept and pass on args. This is something we'll wrap to hide when we
  expose this publicly in Ember/Glimmer for the time being, since that
  was a capability that was not added in the RFC.

### DIFF
--- a/packages/@glimmer/validator/index.ts
+++ b/packages/@glimmer/validator/index.ts
@@ -17,7 +17,7 @@ export {
   EntityTag,
   EntityTagged,
   INITIAL,
-  isConst,
+  isConstTagged,
   isConstTag,
   Revision,
   Tag,
@@ -39,11 +39,16 @@ export {
   consumeTag,
   isTracking,
   track,
-  trackedData,
   memo,
   untrack,
   isConstMemo,
+  Cache,
+  createCache,
+  isConst,
+  getValue,
 } from './lib/tracking';
+
+export { trackedData } from './lib/tracked-data';
 
 export {
   setAutotrackingTransactionEnv,
--- a/packages/@glimmer/validator/lib/tracking.ts
+++ b/packages/@glimmer/validator/lib/tracking.ts
@@ -1,7 +1,16 @@
 import { DEBUG } from '@glimmer/env';
-import { Tag, combine, CONSTANT_TAG, validateTag, Revision, valueForTag } from './validators';
-import { tagFor, dirtyTagFor } from './meta';
-import { markTagAsConsumed, runInAutotrackingTransaction, assertTagNotConsumed } from './debug';
+import {
+  Tag,
+  combine,
+  CONSTANT_TAG,
+  validateTag,
+  Revision,
+  valueForTag,
+  isConstTag,
+} from './validators';
+
+import { markTagAsConsumed, runInAutotrackingTransaction } from './debug';
+import { symbol } from './utils';
 
 type Option<T> = T | null;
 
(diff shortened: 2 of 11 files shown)


rwjblue:
@pzuraq - Looks like this needs a rebase

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -108,64 +118,122 @@ export function memoizeTracked<T>(
...
+  return new Cache(fn);
 }
 
+export const getValue = Cache.getValue;
+
+export const isConst = Cache.isConst;
+
 //////////
 
+// Legacy tracking APIs
```
rwjblue, line 235: Can you define what legacy means in this context? Maybe even make an issue to remove them and link to that from here (that way we know when we can remove the code).

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -108,64 +118,122 @@ export function memoizeTracked<T>(
...
+  constructor(private fn: (...args: unknown[]) => T, debuggingContext?: string | false) {
+    if (DEBUG) {
+      return ({
+        [DEBUG_CACHE]: this,
+        [DEBUG_CACHE_CONTEXT]: debuggingContext,
+      } as any) as Cache<T>;
+    }
+  }
+
+  static getValue<T>(cache: Cache<T>, ...args: unknown[]): T {
```
rwjblue, line 155: Why is this a static function? It isn't clear to me (it doesn't use `this`), and it doesn't seem to coincide with the API from the RFC (which is that `getValue` is a stand alone function).

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -108,64 +118,122 @@ export function memoizeTracked<T>(
...
+}
 
-  return IS_CONST_MAP.get(memoized) === true;
+export function createCache<T>(fn: () => T): Cache<T> {
+  return new Cache(fn);
 }
 
+export const getValue = Cache.getValue;
+
+export const isConst = Cache.isConst;
```
rwjblue, line 231: I think the code would be much easier to understand (and less error prone) if the implementation were moved into these functions (and out of the static functions).

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -108,64 +118,122 @@ export function memoizeTracked<T>(
...
+      cache = debugCache;
+    }
+
+    return isConstTag(cache.tag!);
   }
+}
 
-  return IS_CONST_MAP.get(memoized) === true;
+export function createCache<T>(fn: () => T): Cache<T> {
+  return new Cache(fn);
```
rwjblue, line 226: I'd much prefer something like:

```suggestion
  if (DEBUG) {
    return Object.freeze({
      [DEBUG_CACHE]: this,
      [DEBUG_CACHE_CONTEXT]: debuggingContext,
    });
  }
  return {};
```

packages/@glimmer/validator/lib/tracking.ts
pzuraq: `static` allows us to access the `private` fields of an instance, without marking them as public. It's a nifty trick for us to have friend functions without using Symbols.

packages/@glimmer/validator/lib/tracking.ts
pzuraq: I really prefer using `static` methods over Symbols, but we can convert to symbols if you prefer. There were perf concerns when we were writing MonomorphicTagImpl, if I recall, I think we saw that using Symbols did have some amount of impact, but it was minor.

packages/@glimmer/validator/lib/tracking.ts
pzuraq: Is this meant to replace code in the constructor? I'm not sure how this would work exactly.

FWIW, I put this code in the constructor of `Cache` so that we could create a `new Cache` with the debugging context from `memo()`. This way, we aren't exposing `debuggingContext` to other users of `createCache`. We could do that instead, and provide a wrapper function in Ember like I'm planning on doing with `getValue()`.

packages/@glimmer/validator/lib/tracking.ts
pzuraq: Yeah, I'll add a comment. `track()` shouldn't be necessary at all in the VM once the autotracking refactors are merged, and we should generally be moving away from it. It may be necessary in Ember for a while longer, but I think we'll be able to drop it in favor of `cache` sooner rather than later.

`untrack` is currently mainly used to handle places that were previously not tracked, and that tracking now would cause backtracking rerender assertions. I think once we move everyone forward onto modern APIs, we'll probably be able to remove it, but I'm not sure yet.

packages/@glimmer/validator/lib/tracking.ts
pzuraq: FWIW, this technique also works with true private fields and methods, so I think it would be valuable to use. We could for instance remove the need for the DEBUG logic and wrapper function that way in the future. Or maybe even now if transpilers handle it well enough...

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -96,64 +115,122 @@ export function memo<T, U, V, W, X, Y, Z>(
...
+
+      cache = debugCache;
+    }
+
+    return isConstTag(cache.tag!);
   }
+}
 
-  return IS_CONST_MAP.get(memoized) === true;
+export function createCache<T>(fn: () => T): Cache<T> {
```
krisselden, line 222: the exported function should use a branded interface for Cache and not leak the class Cache

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -96,64 +115,122 @@ export function memo<T, U, V, W, X, Y, Z>(
...
+  return new Cache(fn);
 }
 
+export const getValue = Cache.getValue;
```
krisselden, line 226: recast function to public interface for `Cache.getValue as unknown as <T>(cache: CacheHandle<T>) => T`

krisselden: same with isValue

packages/@glimmer/validator/lib/tracking.ts
rwjblue: I don't understand why the question is symbol vs static method. The static methods don't even use `this` (and are completely disconnected from the `this` binding).

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -96,64 +115,122 @@ export function memo<T, U, V, W, X, Y, Z>(
...
+  private lastValue: T | undefined;
+  private tag: Tag | undefined;
+  private snapshot: Revision = -1;
+
+  constructor(private fn: (...args: unknown[]) => T, debuggingContext?: string | false) {
+    if (DEBUG) {
+      return ({
+        [DEBUG_CACHE]: this,
+        [DEBUG_CACHE_CONTEXT]: debuggingContext,
+      } as any) as Cache<T>;
```
rwjblue, line 148: Seems somewhat bad to return a different thing than `this` in a constructor. I know it works, and why, but seems like a pretty big footgun and I'm not really sure what all this complexity buys us.

packages/@glimmer/validator/lib/tracking.ts
rwjblue: Sure, but:

1. we aren't doing that (no private usage here AFAICT)
2. we only need extra protection in DEBUG builds, where we can already use other techniques (e.g. `WeakMap`/`WeakSet`) to ensure that what we pass back from `createCache` is what we are given here

packages/@glimmer/validator/lib/tracking.ts
rwjblue: Sounds good. I mostly want to make sure the code comments are actionable. This one infers something is going away, but it would be pretty hard to be able to tell if it were ok to remove if you came back and read this again in ~ 6months...

packages/@glimmer/validator/index.ts
```diff
@@ -39,12 +39,17 @@ export {
   consumeTag,
   isTracking,
   track,
-  trackedData,
   memo,
   untrack,
   isConstMemo,
+  Cache,
```
rwjblue, line 45: Should this be exported? I thought the idea was that you'd always have to go through our accessors 🤔

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -96,64 +115,133 @@ export function memo<T, U, V, W, X, Y, Z>(
...
+    cache[DEBUG_LABEL] = debuggingLabel;
   }
 
-  return memoized;
+  return (cache as unknown) as Cache<T>;
 }
 
-export function isConstMemo(memoized: Function) {
-  if (DEBUG && IS_CONST_MAP.has(memoized) && IS_CONST_MAP.get(memoized) === undefined) {
+export function getValue<T>(cache: Cache<T>, ...args: unknown[]): T {
```
rwjblue, line 175: Ideally we don't export types that have `...args` like this, can we tweak `memo` instead to avoid needing this there?

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -96,64 +115,133 @@ export function memo<T, U, V, W, X, Y, Z>(
...
+  let snapshot = cache[SNAPSHOT];
+
+  if (tag === undefined || !validateTag(tag, snapshot)) {
+    beginTrackFrame();
+
+    try {
+      if (DEBUG) {
+        runInAutotrackingTransaction!(
+          () => (cache[LAST_VALUE] = fn(...args)),
+          DEBUG && cache[DEBUG_LABEL]
```
rwjblue, line 195: We are already inside an `if (DEBUG)`, this `DEBUG &&` can probably be removed?

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -72,88 +81,145 @@ export function endTrackFrame(): Tag {
...
     throw new Error(
-      'Attempted to call `isConstMemo` on a memoized function, but the function has not been run at least once yet. You cannot know if a memoized function is constant or not until it has been run at least once. Call the function, then pass it to `isConstMemo`.'
+      `isConst() can only be used on a cache once getValue() has been called at least once. Called with cache function:\n\n${String(
+        cache[FN]
+      )}`
     );
   }
 
-  return IS_CONST_MAP.get(memoized) === true;
+  return isConstTag(cache[TAG]!);
```
krisselden, line 208: you can put the debug error inside a type assertion function like isCache to remove the !

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -72,88 +81,145 @@ export function endTrackFrame(): Tag {
...
+  (memoized as any)[CACHE_KEY] = cache;
+
+  return memoized;
+}
+
+export function isConstMemo(memoized: Function) {
+  let cache = (memoized as any)[CACHE_KEY] as Cache | undefined;
+
+  return cache === undefined ? false : isConst(cache);
+}
```
rwjblue, line 112: p1/5 Can we do the same trick that we ended up with below to ensure that TS errors if you pass a function that was not the result of calling `memo` into `isConstMemo`?

packages/@glimmer/validator/lib/tracking.ts
```diff
@@ -72,88 +81,145 @@ export function endTrackFrame(): Tag {
...
+const TAG: unique symbol = symbol('TAG');
+const SNAPSHOT: unique symbol = symbol('SNAPSHOT');
+const DEBUG_LABEL: unique symbol = symbol('DEBUG_LABEL');
+
+interface InternalCache<T = unknown> {
+  [FN]: (...args: unknown[]) => T;
+  [LAST_VALUE]: T | undefined;
+  [TAG]: Tag | undefined;
+  [SNAPSHOT]: Revision;
+  [DEBUG_LABEL]?: string | false;
```
rwjblue, line 132: Just to make sure, the `false` here is for the `DEBUG && somethingPassedIn` cases?

pzuraq: yep, exactly. We can revisit that in the future, it's just convenient for the time being

PR #1100, opened 2020-05-15
[BUGFIX] Update iterable items only when changed

pzuraq:
Currently, iterable item references always update their value and dirty
their tag, even if the value didn't actually change. This is different
from the baseline behavior in Ember, and a regression.

This PR updates this behavior, and also updates affected tests. These
tests were written prior to the recent VM upgrade, and probably assumed
some differences in behaviors in the testing environment. Now that we're
trying to align the testing environment more closely with Ember, these
tests need to be updated. I added a `@tracked` decorator to the test
helpers, which should allow us to more closely match Ember's behavior.

Fix for https://github.com/emberjs/ember.js/issues/18968

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/each.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/each.ts
@@ -1,5 +1,6 @@
 import { RenderTest } from '../render-test';
 import { test } from '../test-decorator';
+import { tracked } from '../test-helpers/tracked';
 
 export class EachSuite extends RenderTest {
   static suiteName = '#each';
@@ -165,6 +166,38 @@ export class EachSuite extends RenderTest {
   'it renders all items with duplicate key values'() {
     let list = [{ text: 'Hello' }, { text: 'Hello' }, { text: 'Hello' }];
 
+    this.render(`{{#each list key="text" as |item|}}{{item.text}}{{/each}}`, {
+      list,
+    });
+
+    this.assertHTML('HelloHelloHello');
+    this.assertStableRerender();
+
+    list.forEach(item => (item.text = 'Goodbye'));
+
+    this.rerender({ list });
+    this.assertHTML('GoodbyeGoodbyeGoodbye');
+    this.assertStableNodes();
+
+    list = [{ text: 'Hello' }, { text: 'Hello' }, { text: 'Hello' }];
+
+    this.rerender({ list });
+    this.assertHTML('HelloHelloHello');
+    this.assertStableNodes();
+  }
+
+  @test
+  'it updates items if their key has not changed, and the items are tracked'() {
+    class Item {
+      @tracked public text: string;
+
+      constructor(text: string) {
+        this.text = text;
+      }
+    }
+
+    let list = [new Item('Hello'), new Item('Hello'), new Item('Hello')];
+
     this.render(`{{#each list key="@identity" as |item|}}{{item.text}}{{/each}}`, {
       list,
     });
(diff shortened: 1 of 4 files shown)


packages/@glimmer/integration-tests/lib/suites/each.ts
```diff
@@ -165,6 +166,38 @@ export class EachSuite extends RenderTest {
   'it renders all items with duplicate key values'() {
```
pzuraq, line 166: This test was not testing what it says it was, so I updated it to test _that_ properly (which it does) and added a couple more tests

packages/@glimmer/reference/lib/template.ts
```diff
@@ -307,8 +307,12 @@ export class IterationItemReference<T = unknown> implements TemplatePathReferenc
   }
 
   update(value: T) {
-    dirtyTag(this.tag);
-    this.itemValue = value;
+    let { itemValue } = this;
+
+    if (value !== itemValue) {
```
rwjblue, line 312: Can you remind me what `itemValue` actually represents? Is it the value that we find after looking up the `key` path (or `guidFor` if `key="@identity"`)? Or is it literally the item being yielded?

sandydoo: > Or is it literally the item being yielded?

This. If the "iterable" is an array, then for each value in the array you have an `IterationItemReference` where `itemValue` is the value.

rwjblue: Shouldn’t it be the key that we are checking for stability (not the item itself)? I realize that in the default case for Ember there is no difference between the two (since Ember uses `key=“@identity”` by default), but it is possible to specify a key and AFAICT that is the thing that is important to check for stability :thinking:.

For example, imagine the case where the array and its items are always regenerated (e.g. the immutable pattern) so you’ve used `key` to tell `{{#each` that it should care about some stable property on the item (e.g. `id`).

sandydoo: AFAICT we're not checking the stability here. The key is used when [checking the stability of the array itself over here]() – checking which items were added, which were moved, and which stayed in the same place. If the array is touched and an item is considered retained (by comparing keys), then it just always updates the value and the memo (in the case of arrays, the array index). If you don't compare the values at this point, doing anything to the array can become quite an expensive operation O(n). 

There's more though.

If you have an array of objects and modify some property deep inside, then all of this iterator code is skipped and just that nested `PropertyReference` is updated. So the key you provide to `{{each}}` plays no role in that case. But doing anything to array will trigger the iterator synchronization/list revalidation code and call this update method. So we do all this extra work, even though in some cases Ember provides the exact key modified via `dirtyTagFor`.

Also, `{{each-in}}` is possibly even more broken. Since the default key is `@identity`, updating a the value of an existing key won't cause an update, as may be expected by an end user. It will actually destroy and recreate the item. So, again, if you're using render modifier, you'll get a bunch of weird triggers.

TBH I don't understand why we should be comparing values here at all. Aren't there tags for every value anyways? Shouldn't we be checking those instead?

rwjblue: Gotcha, glad you dug in here :smile:.

@pzuraq - Any reason we are checking `===` vs a tag validation?

pzuraq:
each-in does not exist in Glimmer VM, it is a template transform to each provided by Ember. We should test it in Ember

packages/@glimmer/reference/lib/template.ts
pzuraq: It's the opposite in this case - we know that the value has possibly changed, and we need to dirty the tag if so and update it. We don't have another tag to check.

PR #1102, opened 2020-05-18
Add a compile time error when passing arguments to regular HTML elements (e.g. `<a @foo=`)

fivetanley:
Adds an error message in the compile phase to inform the developer that
`@arguments` are not allowed on HTML nodes. This can often come from
copy/pasting refactoring errors.

For example, given the following template:

```handlebars
<a href="https://emberjs.com" @onClick={{action "foo"}}>Ember.js</a>
```

Before, the developer would get a vague error at runtime ([Example taken from
ember.js#18951](https://github.com/emberjs/ember.js/issues/18951)):

```
jquery.js:3827 Uncaught TypeError: func is not a function
    at StatementCompilers.compile (VM32 ember.debug.js:43121)
    at compileStatements (VM32 ember.debug.js:44323)
    at maybeCompile (VM32 ember.debug.js:44312)
    at CompilableTemplateImpl.compile (VM32 ember.debug.js:44292)
    at Object.evaluate (VM32 ember.debug.js:51011)
    at AppendOpcodes.evaluate (VM32 ember.debug.js:49382)
    at LowLevelVM.evaluateSyscall (VM32 ember.debug.js:52284)
    at LowLevelVM.evaluateInner (VM32 ember.debug.js:52240)
    at LowLevelVM.evaluateOuter (VM32 ember.debug.js:52232)
    at JitVM.next (VM32 ember.debug.js:53175)
```

Now, they get an error message that looks like:

```
SyntaxError: ${attribute.name} is not a valid attribute name. @arguments
are only allowed on components, but the tag for this element
(\`${elementNode.tag}\`) is a regular, non-component HTML element. (location
line and column)
```

Given that there's no special behavior glimmer adds for `@arguments` on regular
HTML nodes, and it seems invalid to put `@` in attribute names in regular HTML
(validator.w3c.org gives ` Attribute @onclick is not serializable as XML 1.0`
error message for the example template above), it seems helpful for the glimmer
compiler to fail at build time to prevent copy/paste errors and mistakes.

Fixes https://github.com/emberjs/ember.js/issues/18951

### DIFF
--- a/packages/@glimmer/compiler/test/compiler-test.ts
+++ b/packages/@glimmer/compiler/test/compiler-test.ts
@@ -55,6 +55,19 @@ function test(desc: string, template: string, ...expectedStatements: BuilderStat
 const Append = Builder.Append;
 const Concat = Builder.Concat;
 
+QUnit.test(
+  '@arguments are on regular non-component/regular HTML nodes throws syntax error',
+  function(assert) {
+    const template = strip`
+    <a @onClick={{action "hi"}}>Link</a>
+  `;
+    assert.throws(
+      () => compile(template),
+      /@onClick is not a valid attribute name. @arguments are only allowed on components, but the tag for this element \(`a`\) is a regular, non-component HTML element/
+    );
+  }
+);
+
 test('HTML text content', 'content', s`content`);
 
 test('Text curlies', '<div>{{title}}<span>{{title}}</span></div>', [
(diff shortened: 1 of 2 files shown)


packages/@glimmer/compiler/test/compiler-test.ts
```diff
@@ -55,6 +55,19 @@ function test(desc: string, template: string, ...expectedStatements: BuilderStat
 const Append = Builder.Append;
 const Concat = Builder.Concat;
 
+QUnit.test(
```
fivetanley, line 58: this doesn't match the `test` style in the rest of the file, but i thought it might be more clear to just use `QUnit.test` directly rather than extend the `test` function in this file to allow a callback/expect errors. Totally open to change this to whatever the maintainers prefer, let me know!

PR #1103, opened 2020-05-18
[BUGFIX] Adds an assertion to prevent duplicate validator packages

pzuraq:
Adds an assertion that will throw if duplicate validator packages are detected.
This is because validator depends on a few global values for things to work properly
at the moment. We may be able to relax this restriction in the future, mainly adding
it now to prevent confusion when things aren't working properly.

### DIFF
--- a/packages/@glimmer/validator/index.ts
+++ b/packages/@glimmer/validator/index.ts
@@ -1,3 +1,17 @@
+import { symbolFor, getGlobal } from './lib/utils';
+
+const globalObj = getGlobal();
+
+const GLIMMER_VALIDATOR_REGISTRATION = symbolFor('GLIMMER_VALIDATOR_REGISTRATION');
+
+if (globalObj[GLIMMER_VALIDATOR_REGISTRATION] === true) {
+  throw new Error(
+    'The `@glimmer/validator` library has been included twice in this application. It could be different versions of the package, or the same version included twice by mistake. `@glimmer/validator` depends on having a single copy of the package in use at any time in an application, even if they are the same version. You must dedupe your build to remove the duplicate packages in order to prevent this error.'
+  );
+}
+
+globalObj[GLIMMER_VALIDATOR_REGISTRATION] = true;
+
 export {
   ALLOW_CYCLES,
   bump,
--- a/packages/@glimmer/validator/lib/utils.ts
+++ b/packages/@glimmer/validator/lib/utils.ts
@@ -11,3 +11,18 @@ export const symbol =
   typeof Symbol !== 'undefined'
     ? Symbol
     : (key: string) => `__${key}${Math.floor(Math.random() * Date.now())}__` as any;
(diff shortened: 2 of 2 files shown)


packages/@glimmer/validator/index.ts
```diff
@@ -1,3 +1,17 @@
+import { symbolFor, getGlobal } from './lib/utils';
+
+const globalObj = getGlobal();
+
+const GLIMMER_VALIDATOR_REGISTRATION = symbolFor('GLIMMER_VALIDATOR_REGISTRATION');
+
+if ((globalObj as any)[GLIMMER_VALIDATOR_REGISTRATION] === true) {
```
pzuraq, line 7: This seems to be the best way to do this, we can't actually add it as an interface because `Symbol.for()` does not produce a `unique symbol`, so it can't be used as a key in an interface.

packages/@glimmer/validator/lib/utils.ts
```diff
@@ -11,3 +11,18 @@ export const symbol =
   typeof Symbol !== 'undefined'
     ? Symbol
     : (key: string) => `__${key}${Math.floor(Math.random() * Date.now())}__` as any;
+
+export const symbolFor =
+  typeof Symbol !== 'undefined'
+    ? Symbol.for
+    : (key: string) => (`__GLIMMER_VALIDATOR_SYMBOL_FOR_${key}` as unknown) as symbol;
+
+export function getGlobal() {
```
rwjblue, line 20:
```suggestion
export function getGlobal(): any {
```

This would avoid the extra cast right?

packages/@glimmer/validator/index.ts
```diff
@@ -1,3 +1,17 @@
...
+  throw new Error(
+    'Two versions of the `@glimmer/validator` library have been included in this application. `@glimmer/validator` depends on having a single version of the package in use at any time in an application. You must dedupe your build to remove the duplicate packages in order to prevent this error.'
```
rwjblue, line 9: > Two versions of the `@glimmer/validator` library have been included in this application.

FWIW, this isn't quite correct. 🤔 You could end up with this error if you have the same version in your output twice (which I wouldn't consider "two versions"). I don't really know how to tweak the prose to make that particular quibble moot though.

packages/@glimmer/validator/lib/utils.ts
pzuraq: yeah, that's true, let's do that for now

packages/@glimmer/validator/index.ts
pzuraq: Updated, I think it's a bit better now. Thanks for pointing that out

john-griffin:
We use glimmer for embed code on third party websites we do not control, sometimes our customers embed our code multiple times and this error is hit, is there any way to get around it?

PR #1108, opened 2020-05-29
Remove $REVISION optimization

pzuraq:
Since we added the `subtagBufferCache`, this optimization is no longer necessary. Updating a tag with `updateTag` will no longer cause a parent tag's cache to ever be in an invalid state, it works as a transparent update instead, so the value we return is definitely correct if we call `COMPUTE` directly.

This optimization still existing is confusing and makes it harder to reason about how tags work, so I'm removing it now to make it easier. We should benchmark to make sure that it can be removed and I'm correct, but given the original reason it was added is no longer relevant, I think it should be fine.


pzuraq:
I’m going to, I think I messed it up, there’s a TS error

PR #1110, opened 2020-06-12
[FEAT] Destroyables

pzuraq:
Implements the [destroyables RFC](https://github.com/emberjs/rfcs/blob/master/text/0580-destroyables.md), refactoring the original implementation that existed. Some notes:

- Added a `destroyChildren` API, since that pattern is used in the VM
  currently. It's not ideal, but I think we need to do larger refactors
  to change it.
- Exposes `setScheduleDestroy` and `setScheduleDestroyed` to allow the
  embedding environment to hook into scheduling. These set the callbacks
  globally (modifier scope).
- Replicated the old behavior in the test environment by flushing during the
   end of the transaction. This is not ideal, but given we don't have backburner
   it seemed like the best place to put it.

### Breaking Changes

- `getDestructor` -> `getDestroyables`, and the API surrounding destruction has changed. Embedding environments should now use the destroyables API to setup destruction.
- `setScheduleDestroy` and `setScheduleDestroyed` should be used to schedule destruction.


packages/@glimmer/interfaces/lib/components/component-manager.d.ts
```diff
@@ -80,68 +80,6 @@ export interface WithUpdateHook<ComponentInstanceState = unknown>
   update(state: ComponentInstanceState, dynamicScope: Option<DynamicScope>): void;
 }
 
-// export interface ComponentManager<
```
pzuraq, line 83: I think this was leftover from an earlier PR, removing as part of general cleanup

packages/@glimmer/util/lib/weak-set.ts
```diff
@@ -0,0 +1,18 @@
+export default (typeof WeakSet === 'function'
```
pzuraq, line 1: This is moving upstream from Ember, we'll be sure to remove it there when we integrate these changes

packages/@glimmer/util/lib/list-utils.ts
```diff
@@ -96,20 +94,6 @@ export class LinkedList<T extends LinkedListNode> implements Slice<T>, Drop {
 
     return node;
   }
-
-  [WILL_DROP]() {
```
pzuraq, line 100: I'm not sure we ever actually destroyed a LinkedList before, I couldn't find a place in the code. This seems like it was conceptually part of the original work, but then didn't get integrated in the end. I don't think it makes much sense anymore with the new direction we're going.

packages/@glimmer/runtime/lib/destroyables.ts
```diff
@@ -0,0 +1,187 @@
+import { DEBUG } from '@glimmer/env';
+import { Destroyable, Destructor } from '@glimmer/interfaces';
+import { _WeakSet as WeakSet } from '@glimmer/util';
+
+const EAGER_DESTRUCTORS = new WeakMap<Destroyable, Destructor<Destroyable>[]>();
```
krisselden, line 5: way too many weak collections, use one weak collection to associate the monomorphic destroyable state.  also, optimize if things will often have 1 to 1 maybe consider typing it `Destroyable | Destroyable[]`

packages/@glimmer/runtime/lib/destroyables.ts
```diff
@@ -0,0 +1,187 @@
...
+  let parents = DESTROYABLE_PARENTS.get(child);
+
+  if (children === undefined) {
+    children = [];
+    DESTROYABLE_CHILDREN.set(parent, children);
+  }
+
+  if (parents === undefined) {
+    parents = [];
+    DESTROYABLE_PARENTS.set(parent, parents);
```
krisselden, line 27: things can have multiple parents? is the idea that a child only lives as long as the shortest lived parent? if most things will only have one parent, this should use `Destroyable | Destroyable[]` as the type.

The fact that you always mutate and read parents and children together and you alway check if it has been registered further argues for a state object versus multiple weak collections, you would have one dictionary get or set then fast properties for the rest of the lookups and sets

rwjblue: > is the idea that a child only lives as long as the shortest lived parent? 

Yes

packages/@glimmer/runtime/index.ts
```diff
@@ -78,5 +78,20 @@ export {
...
+export {
+  destroy,
+  destroyChildren,
+  registerDestructor,
+  unregisterDestructor,
+  associateDestroyableChild,
+  isDestroying,
+  isDestroyed,
+  scheduleDestroy,
+  scheduleDestroyed,
```
rwjblue: Can you underscore these?

packages/@glimmer/runtime/index.ts
```diff
@@ -78,5 +78,20 @@ export {
   rehydrationBuilder,
   SERIALIZATION_FIRST_NODE_STRING,
 } from './lib/vm/rehydrate-builder';
+export {
+  destroy,
+  destroyChildren,
```
rwjblue: Can you underscore this?

packages/@glimmer/runtime/lib/destroyables.ts
```diff
@@ -0,0 +1,309 @@
...
+import { Destroyable, Destructor } from '@glimmer/interfaces';
+import { debugToString } from '@glimmer/util';
+
+interface DestroyableMeta<T extends Destroyable> {
+  source?: T;
+  parents: Destroyable | Destroyable[] | null;
+  children: Destroyable | Destroyable[] | null;
+  eagerDestructors: Destructor<T> | Destructor<T>[] | null;
+  destructors: Destructor<T> | Destructor<T>[] | null;
+  state: 0 | 1 | 3;
```
rwjblue: Is this not the DestroyingState below? If so, why does it not match?

PR #1111, opened 2020-06-20
feat: better DEBUG memoization

runspired:
I recently noticed that the effect of our extra `PropertyReference` creations in DEBUG to preserve context for debuggability decreases the level of memoization that occurs. For instance, a component with a property that is a getter will re-run that getter every location that property is used in it's template. For getters that do not return a stable value, this is a footgun.

In DEBUG (without this fix) this change to memoization semantics can lead to errors that would not occur in production, both in application code (should referential integrity matter) and in the form of rendering assertions for an issue that would not exist in production (a value changing during render).

### DIFF
--- a/packages/@glimmer/reference/lib/template.ts
+++ b/packages/@glimmer/reference/lib/template.ts
@@ -192,6 +192,16 @@ export class HelperRootReference<T = unknown> extends RootReference<T> {
   }
 }
 
+/*
+  DEBUG only fields. Keeping this in a separate interface which Typescript then
+  merges with the class definition allow us to have a DEBUG only property without
+  paying the cost of that field being present and initialized (usually to void 0)
+  in production.
+*/
+export interface PropertyReference {
+  ref?: PropertyReference;
+}
+
 /**
  * PropertyReferences represent a property that has been accessed on a root, or
  * another property (or iterable, see below). `some` and `prop` in
@@ -200,10 +210,6 @@ export class HelperRootReference<T = unknown> extends RootReference<T> {
  * recursively calling `get` on the previous reference as a template chain is
  * followed.
  */
-export interface PropertyReference {
-  ref?: PropertyReference;
-}
-
 export class PropertyReference implements TemplatePathReference {
   public tag: Tag;
   private valueTag: UpdatableTag;
--- a/packages/@glimmer/reference/lib/template.ts
+++ b/packages/@glimmer/reference/lib/template.ts
@@ -103,17 +103,17 @@ export abstract class RootReference<T = unknown> implements TemplatePathReferenc
         this.didSetupDebugContext = true;
         this.env.setTemplatePathDebugContext(this, this.debugLogName || 'this', null);
       }
-
-      return new PropertyReference(this, key, this.env);
-    } else {
-      let ref = this.children[key];
-
-      if (ref === undefined) {
-        ref = this.children[key] = new PropertyReference(this, key, this.env);
-      }
-
-      return ref;
     }
+
+    let ref = this.children[key];
+
+    if (ref === undefined) {
+      ref = this.children[key] = new PropertyReference(this, key, this.env);
+    } else if (DEBUG) {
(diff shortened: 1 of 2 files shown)


runspired:
cc @rwjblue @pzuraq for review. I'm a serious n00b with this code so I may have done this all wrong, and I wasn't sure how to test that whatever DEBUG context this was trying to preserve was actually preserved since the existing tests just inquire whether the method was called.

pzuraq (approved):
I think this is definitely a move in the right direction, I wanted to have the same semantics when first implementing but couldn't think of a way to do it at the time, seems obvious now 😄

packages/@glimmer/reference/lib/template.ts
```diff
@@ -200,6 +200,10 @@ export class HelperRootReference<T = unknown> extends RootReference<T> {
  * recursively calling `get` on the previous reference as a template chain is
  * followed.
  */
+export interface PropertyReference {
```
pzuraq, line 203: This interface is used to merge the property into the class definition dynamically, without adding a class field. Can you add a comment to that effect, to document this?

packages/@glimmer/reference/test/template-test.ts
```diff
@@ -62,6 +62,51 @@ module('@glimmer/reference: template', () => {
...
+      );
+
+      let ref1 = ref.get('foo');
+      ref1.value();
+      let ref2 = ref.get('foo');
+      ref2.value();
+      let ref3 = ref.get('foo');
+      ref3.value();
+      assert.ok(true, 'we did not error');
+      assert.notStrictEqual(ref1, ref2, 'second access is a new reference');
```
pzuraq, line 105: This will only be true in `DEBUG`, can you either wrap this part of the test in a `DEBUG` statement or remove the check?

packages/@glimmer/reference/test/template-test.ts
```diff
@@ -62,6 +62,51 @@ module('@glimmer/reference: template', () => {
...
+      let component = {
+        get foo() {
+          pullCount++;
+          return pullCount;
+        },
+      };
+      let ref = new ComponentRootReference(
+        component,
+        new (class extends TestEnv {
+          setTemplatePathDebugContext(_ref: TemplatePathReference, key: string) {
```
pzuraq, line 79: this API is only used in DEBUG, can we wrap this portion of the test in the DEBUG flag, or split it out into a separate test that is?

pzuraq (commented):
Can you add an integration test in `@glimmer/integration-tests`? You can add one in https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/integration-tests/lib/suites/emberish-components.ts, following the style of the tests in there. It should focus on just the number of times the property is calculated, and ignore any of the DEBUG mode stuff.

packages/@glimmer/reference/test/template-test.ts
runspired: @pzuraq is the idea to prep these tests for potentially running in prod?

rwjblue: Ya, I think in general we’d like to be able to (not sure if there are other tests that break in prod atm, but Ember does run its tests in prod builds)

rwjblue (changes requested):
IMHO, this is only missing a high level test for the actual issue. I know this isn’t the syntax used in this repo (it’s what you’d use in an Ember app), but it illustrates the kind of test I’m looking for:


```js
@test
'repeated access to getter only invokes getter once'() {
  this.owner.register('component:whatever', class extends GlimmerComponent {
    get foo() {
      assert.step('updating foo');
      return 'foo';
    }
  });
  this.owner.register('template:components/whatever', hbs`{{this.foo}}{{this.foo}}`);
  render(hbs`<MyComponent />`);
  assert.verifySteps(['updating foo']);

}
```

rwjblue:
@runspired - Have any time to add that test?

PR #1112, opened 2020-06-27
(fix): {{debugger}} on template

abhilashlr:
Fixes https://github.com/emberjs/ember.js/issues/17962

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
@@ -8,7 +8,7 @@ export type DebugGet = (path: string) => unknown;
 
 export type DebugCallback = (context: unknown, get: DebugGet) => void;
 
-function debugCallback(context: unknown, get: DebugGet): void {
+function templateWithBackingObjectDebugCallback(context: unknown, get: DebugGet): void {
   console.info('Use `context`, and `get(<path>)` to debug this template.');
 
   // for example...
@@ -17,6 +17,18 @@ function debugCallback(context: unknown, get: DebugGet): void {
 
   // eslint-disable-next-line no-debugger
   debugger;
+}
+
+function templateOnlyDebugCallback(get: DebugGet): void {
+  console.info('Use `get(<@path>)` to debug this template.');
+
+  // for example... `get('@foo');`
+  // eslint-disable-next-line no-debugger
+  debugger;
+}
+
+function debugCallback(context: unknown, get: DebugGet): void {
+  context ? templateWithBackingObjectDebugCallback(context, get) : templateOnlyDebugCallback(get);
 }
 
 let callback = debugCallback;


packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
```diff
@@ -19,6 +19,22 @@ function debugCallback(context: unknown, get: DebugGet): void {
   debugger;
 }
 
+function templateOnlyDebugCallback(get: DebugGet): void {
+  console.info('Use `get(<@path>)` to debug this template.');
```
lifeart, line 23: Is it possible to get all passed arguments, unsing such path notation? (kinda `this.args`)...

rwjblue: Ya, we should probably do something like this automatically, but the current plumbing for it doesn't exist (AFAICT). Seems good to do in a follow up...

packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
```diff
@@ -19,6 +19,22 @@ function debugCallback(context: unknown, get: DebugGet): void {
...
+
+  // for example... `get('@foo');`
+  // eslint-disable-next-line no-debugger
+  debugger;
+}
```
rwjblue, line 28:
```
$ tsc -p tsconfig.json --noEmit --pretty false
packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts(22,36): error TS6133: 'get' is declared but its value is never read.
```

abhilashlr: @rwjblue yes I notice this error, but I’m not sure what’s the right way to fix this. Could you please help in here?

rwjblue: You can probably disable it manually, but without actually calling `get` it is likely for the JS runtime to optimize it away (making this not work properly).

If you take a look at the other pathway, we do `get(this)` so that the `get` function is used...

abhilashlr: I've added a variable assignment and ignored using `ts-ignore`. Should we do it this way or say `console.log(get('@foo'));`? The console output would output undefined all the time someone tries to use a `debugger` on the template.

packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
```diff
@@ -23,6 +23,9 @@ function templateOnlyDebugCallback(get: DebugGet): void {
   console.info('Use `get(<@path>)` to debug this template.');
 
   // for example... `get('@foo');`
+  // @ts-ignore
```
rwjblue, line 26: Does TS have an issue with this, or eslint? In general, I would prefer to not have `@ts-ignore` in the section of code that folks would see when they hit this debugger...

packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
rwjblue: Ya, I think in general what you have is good, though I wonder if we can silence the TS error either globally for that file or some ohter way (basically I'd prefer the user not see `//@ts-ignore` in the code just above their debugger).

packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
lifeart: `// @ts-expect-error` ? (https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/#ts-expect-error-comments)

NullVoxPopuli:
What all is remaining on this? I consider DX fixes high value <3

NullVoxPopuli:
What all is remaining on this? I consider DX fixes high value <3

PR #945, opened 2019-05-29
feat:  performance improvements for keyed lists changes

lifeart:
this is failing tests related to https://github.com/glimmerjs/glimmer-vm/issues/940

This PR must increase keyed lists swap perf up to 10x! Moon

(See swap rows row)

from this:
image

to this:
image

PR #1114, opened 2020-07-03
[REFACTOR] Remove LinkedList and simplify iteration

pzuraq:
This refactor removes several concepts from the VM, including:

- LinkedList, which is replaced with plain arrays
- IterationArtifacts, IteratorSynchronizers, and the ListRevalidationDelegate, which have been consolidated into a single `IterableReference` and the ListBlockOpcode directly.
- LabelOpcodes, which are no longer necessary since we jump by index through the updating opcodes.

Overall this simplifies the internals a decent amount, and in testing we saw performance improvements in lists (not major improvements, but decent ones.)

### Breaking Changes

- Removes the following exports from `@glimmer/reference`:
  - `Iterable`
  - `OpaqueIterable`
  - `AbstractIterable`
  - `IterationArtifacts`
  - `ReferenceIterator`
  - `IteratorSynchronizer`
  - `IteratorSynchronizerDelegate`
- Renames `IterableImpl` to `IterableReference`
- Removes the following exports from `@glimmer/interfaces`
  - `LinkedList`
  - `LinkedListNode`
  - `ListNode`
  - `Slice`
  - `CloneableListNode`
  - `ListSlice`
- Replaced `LinkedList` usage in interfaces with arrays
- Removes `iterableFor` from the `Environment` class. Consumers don't directly extend the class anymore, but it could still be used in theory.
- Removes the following exports from `@glimmer/util`
  - `CloneableListNode`
  - `EMPTY_SLICE`
  - `LinkedList`
  - `LinkedListNode`
  - `ListNode`
  - `ListSlice`
  - `Slice`


packages/@glimmer/reference/test/utils/iterator.ts
```diff
@@ -1,165 +1,165 @@
...
-    });
-  }
-  return result;
-}
-
-export function sync(target: Target, artifacts: IterationArtifacts) {
-  let synchronizer = new IteratorSynchronizer({ target, artifacts, env: null });
-  synchronizer.sync();
-}
+// import {
```
rwjblue, line 1: Can you delete this file? Seems like it is fully commented out...

pzuraq: It can be as it turns out 😄 updated

packages/@glimmer/runtime/lib/vm/update.ts
```diff
@@ -196,189 +188,207 @@ export class TryOpcode extends BlockOpcode implements ExceptionHandler {
...
+      let { key } = item;
+
+      if (opcode !== undefined && opcode === item.key) {
+        this.retainItem(opcode, item);
+        currentOpcodeIndex++;
+      } else if (itemMap.has(key)) {
+        let itemOpcode = itemMap.get(key)!;
+
+        if (itemOpcode.seen === true) {
+          this.moveItem(itemOpcode, item, opcode);
```
lifeart, line 304: will it support "cheap" 2 items swap? ref: https://github.com/glimmerjs/glimmer-vm/pull/945

pzuraq: It should be exactly the same algorithm, just simplified. If it looks different, please let me know.

packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -287,15 +285,6 @@ export class EnvironmentImpl<Extra> implements Environment<Extra> {
     }
   }
 
-  iterableFor(ref: VersionedPathReference, inputKey: unknown): OpaqueIterable {
```
rwjblue, line 290: Just wanted to note in the description that this is a breaking change as well (but in practice no consumers use it, I checked both Ember and Glimmer.js)

packages/@glimmer/runtime/lib/utils/tags.ts
```diff
@@ -12,7 +11,7 @@ import { Tag, Tagged, createCombinatorTag, CONSTANT_TAG } from '@glimmer/validat
 export function combineTagged(tagged: ReadonlyArray<Tagged>): Tag {
   let optimized: Tag[] = [];
 
-  for (let i = 0, l = tagged.length; i < l; i++) {
```
rwjblue, line 15: I'm curious about this one, why remove the length caching?

pzuraq: @krisselden showed me [this very helpful blog post](https://mrale.ph/blog/2014/12/24/array-length-caching.html) that explained how caching actually ends up forcing the VM to do _more_ work, not less. 

Basically, array access needs to be bounds checked no matter what, so `.length` is always read anyways. If it is read as part of the `for` loop declaration, then the VM is smart about it and caches the value for when the array is accessed in the loop. Otherwise, it has to load it again itself when doing the check.

packages/@glimmer/runtime/lib/vm/update.ts
lifeart: @pzuraq looks like updated tests not counting `moving/retained` nodes count,  but it's quite expensive DOM operation, see https://github.com/glimmerjs/glimmer-vm/pull/1114/files#diff-6a84e01f254cabe1d50370a1616b9797L253
could we have minimal check for it?

packages/@glimmer/reference/test/iterable-test.ts
```diff
@@ -1,530 +1,186 @@
...
-  target.cleanHistory();
-  sync(target, artifacts);
+      arr.pop();
+      arr.unshift(godfrey);
 
-  assert.equal(
-    target.serializeHistory(),
-    'retain:1,move:8,retain:3,retain:4,retain:5,retain:6,retain:7,move:2'
-  );
-  assert.equal(target.historyStats.move, 2, 'moved nodes count');
```
lifeart, line 253: this one ^

pzuraq: I added some new instrumentation to bring back the coverage from these tests. Originally I thought the tests were only testing the test harness, but in reality they were in fact testing the actual behavior (since the `IteratorSynchronizer` was a separate class), and it would have reduced coverage to remove them entirely.

I also caught some issues thanks to the tests, my understanding of the algorithm was flawed because I wasn't taking into account the fact that the list was changing _as it was synchronizing_. I updated the algorithm and compared it using counts, and it appears to be as optimal as the previous one.

We should try to test the outcome of this is Krausest when we land it.

PR #1115, opened 2020-07-13
[REFACTOR] Converts VM to use tracking internally

pzuraq:
Updates the VM to use revision tracking internally. Prior to this change, the VM 
would manually track changes by constructing its own tags and passing them 
around. This was a complicated and error prone process that did not align with
the future of Glimmer, which is now based on automatic revision tracking.

The main shift in this PR is moving away from publicly exposing tag properties
on values (e.g. `VersionedReference`) and toward tracking consumption of tags
in general. At a high level, every component now wraps itself in a tracking 
transaction while the VM in executing. All of the opcodes that run during that
component's execution are tracked, and will cause that component's updating 
opcodes to rerun if any consumed tags change.

For embedding environments, this means that most APIs are now tracked in some 
form or another, and rather than exposing tags via creating/updating them, they
should be consumed at the right point in the lifecycle. The biggest shift here
is that tags must be consumed during _updates_ as well as _creation_, since tags
are no longer reused over and over again.

## Features

- Custom iterables are now autotracked!

## New APIs

- `logTrackingStack` added to tracking debug functions
- `beginUntrackFrame` and `endUntrackFrame` added as a way to begin/end without 
  adding a `try`

## Breaking Changes

- `VersionedReference` and `VersionedPathReference` have been removed.
- `isConstTagged` has been removed
- `getTag` has been removed from component managers.
- `getDebugName` added as a required method to component and modifier managers
- Tracking debug functions renamed (`Autotracking` -> `Tracking`)
- References now need to implement the `isConst()` method
- Removes the `State` and `StableState` APIs
- `getTemplatePathDebugContext` and `setTemplatePathDebugContext` are no longer
  required as part of the environment.
- `TemplateIterator.sync` no longer executes in an env transaction
- Removes the `memo` API, since it ended up not being the abstraction we used
  internally as originally thought


rwjblue:
Will need a rebase (to fix the conflict) so we can get a CI run kicked off...

rwjblue:
Can you update the description a bit (include a list of the breaking changes, mention _generally_ what we are trying to do here, etc)?

packages/@glimmer/reference/lib/iterable.ts
```diff
@@ -275,3 +278,40 @@ class ArrayIterator implements OpaqueIterator {
     return { key, value, memo };
   }
 }
+
+export class NativeIteratorDelegate<T = unknown> implements IteratorDelegate {
```
pzuraq, line 282: This was necessary for testing the native iterator stuff. We should probably removed defaults in general from the environment

packages/@glimmer/reference/lib/template.ts
```diff
@@ -121,75 +83,46 @@ export abstract class RootReference<T = unknown> implements TemplatePathReferenc
...
   value() {
     return this.inner;
   }
+
+  isConst() {
+    return true;
+  }
+
+  // Make type checker happy...
+  compute() {
```
pzuraq, line 101: This will be fixed with Monomorphic References

pzuraq:
Benchmark results show a small improvement overall: [results.pdf](https://github.com/glimmerjs/glimmer-vm/files/5093348/results.pdf)

PR #1118, opened 2020-07-21
[WIP][Refactor] Streamline main rendering loop

pzuraq:
Streamlines the main rendering loop to avoid branches that trigger
deoptimizations.


NullVoxPopuli:
@pzuraq what all is left here?

pzuraq:
I don't think this ended up having the perf benefits we were hoping for, so probably doesn't need to be merged.

lifeart:
Or we need to improve test setup, as I understand, we render list only once, and likely chain it never been optimized by browser.

Issue #1116, opened 2020-07-17
Issue with TypeScript 3.9


brodybits:
I am able to do `yarn build:typecheck` with TypeScript 3.8 but not TypeScript 3.9 update such as 3.9.2 or 3.9.7. Here is an example (last one on the list):

```
packages/@glimmer/validator/lib/validators.ts(113,37): error TS2422: A class can only implement an object type or intersection of object types with statically known members.
error Command failed with exit code 1.
```

Some more context is coming soon.

brodybits:
I have been updating Prettier to use TypeScript to do JSDoc type check on various parts of the code. One part of the code does `require("@glimmer/syntax")`, and I had to add `@glimmer/reference` to `devDependencies` to get past a bunch of "cannot find module" errors in <https://github.com/prettier/prettier/pull/8778>.

But now <https://github.com/prettier/prettier/pull/8778> is blocked due to the following error from TypeScript 3.9.6:

```
node_modules/@glimmer/validator/dist/types/lib/validators.d.ts:78:45 - error TS2422: A class can only implement an object type or intersection of object types with statically known members.

78 declare class MonomorphicTagImpl implements MonomorphicTag {
                                               ~~~~~~~~~~~~~~
```

I don't think it would be practical for Prettier to downgrade to TypeScript 3.8. If there is any way we can resolve just the one issue in `@glimmer/validator`, it would unblock my work in <https://github.com/prettier/prettier/pull/8778>.

Issue #1117, opened 2020-07-17
@glimmer/syntax & TypeScript


brodybits:
We can see from <https://github.com/prettier/prettier/pull/8778> and the context I gave for #1116 that I encountered an issue between `@glimmer/syntax` and the TypeScript compiler. I had to add `@glimmer/reference` to `devDependencies` to get past some "module not found" errors.

I think this has much lower priority than #1116 but does not seem to be very user friendly.

PR #1120, opened 2020-07-27
Fix `MonomorphicTag` type error for downstream consumers

dfreeman:
### Context

As of TypeScript 3.9, [intersections of types with a conflicting discriminator value now resolve to `never`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/).

Previously, the `MonomorphicTag` type defined here amounted to `Tag & { [TYPE]: never }`, but as of 3.9 that `never` now propagates up and makes the entire `MonomorphicTag` type `never`, which is invalid as the target of an `implements` clause, causing the error described in #1116.

### This Change

This PR makes `MonomorphicTagImpl` parametric over the specific enum value it has as its `[TYPE]`. The runtime behavior is untouched, and I believe this is at least as typesafe as the prior approach. Any properties added to one of the exported `{Dirtyable,Updatable,Combinator,Constant}Tag` types will produce a type error elsewhere in the `validators` module if `MonomorphicTagImpl` doesn't fulfill that contract.

### Other Work

There are several other type errors that crop up on bumping the local version of TypeScript in this repo, but #1116 seems to be the only one that leaks into the public declarations.

Obviously TS stability is [still something the community is working out](https://github.com/typed-ember/ember-cli-typescript/pull/1158) and the core Ember/Glimmer repos haven't made any commitments in that area, but it would be great to at least start flagging breakage in the current stable/beta versions of TypeScript in CI.

From the look of things, that might be as simple as adding a couple extra steps to https://github.com/glimmerjs/glimmer-vm/blob/ac39630bf193fcbecc2ada2d60a75a4f77ab9d08/.github/workflows/ci.yml#L87-L95 to install `typescript@latest` and/or `typescript@next` and then running `yarn test:types` again, but I'm not sure 
how y'all would best like to reason about that.

/cc @pzuraq @chriskrycho since we've discussed this particular error in the past

### DIFF
--- a/packages/@glimmer/validator/lib/validators.ts
+++ b/packages/@glimmer/validator/lib/validators.ts
@@ -1,5 +1,5 @@
 import { DEBUG } from '@glimmer/env';
-import { UnionToIntersection, symbol } from './utils';
+import { symbol } from './utils';
 import { assertTagNotConsumed } from './debug';
 
 //////////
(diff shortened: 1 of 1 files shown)


rwjblue:
This _seems_ like a good change to me, but I've pinged @krisselden / @pzuraq for review as they are much more in touch with the original goals of the monomorphic tag setup **and** TypeScript specifics.

rwjblue:
RE: testing nightly/early TS versions; Yes, we should absolutely setup multiple TS version tests (including nightly for early warning).

rwjblue:
@dfreeman - I'm curious, how did you run into the issue? Was it due to something about how `@glimmer/component` is setup or some other lib?

dfreeman:
> I'm curious, how did you run into the issue? Was it due to something about how @glimmer/component is setup or some other lib?

I hit both this issue and the one described in https://github.com/glimmerjs/glimmer-vm/issues/1117 via a dependency on `@glimmer/syntax` in one of my projects (which is usually how I end up running into weird type stuff here 🙂)

As far as more sanctioned entrypoints go, both the stable and beta releases of `@glimmer/component` look like they avoid this, since they don't have a dependency on the `validator` package.

The stable release of `@glimmer/tracking` does depend on `@glimmer/validator`, but its type declarations never actually refer to it, so the type error never comes up in normal usage. The beta 2.0 version of `tracking` has `export { setPropertyDidChange } from '@glimmer/validator'` though, so that _does_ trigger the error as soon as you import `tracked` in a project.

rwjblue:
Thanks for explaining @dfreeman!

PR #1122, opened 2020-07-30
[Refactor] Use a map to dedupe constants instead of indexOf

pzuraq:
This PR updates the constants pool to use a Map for deduping rather than `indexOf`, which can get very expensive in larger applications. It also caches reified string arrays, so we don't re-reify for every component invocation, and refactors arguments so that we don't mutate those arrays when currying.

[artifact-19.pdf](https://github.com/glimmerjs/glimmer-vm/files/5003731/artifact-19.pdf)

### DIFF
--- a/packages/@glimmer/program/lib/constants.ts
+++ b/packages/@glimmer/program/lib/constants.ts
@@ -1,12 +1,4 @@
-import {
-  SymbolTable,
-  CompileTimeConstants,
-  EMPTY_ARRAY,
-  ConstantPool,
-  RuntimeConstants,
-} from '@glimmer/interfaces';
-
-const UNRESOLVED = {};
+import { CompileTimeConstants, ConstantPool, RuntimeConstants } from '@glimmer/interfaces';
 
 export const WELL_KNOWN_EMPTY_ARRAY_POSITION = 0;
 const WELL_KNOW_EMPTY_ARRAY = Object.freeze([]);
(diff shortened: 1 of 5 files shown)


packages/@glimmer/program/lib/constants.ts
```diff
@@ -14,26 +14,35 @@ const WELL_KNOW_EMPTY_ARRAY = Object.freeze([]);
...
+  protected values: unknown[] = [WELL_KNOW_EMPTY_ARRAY];
+  protected indexMap: Map<unknown, number> = new Map();
+
+  // protected strings: string[] = [];
+  // protected arrays: number[][] | EMPTY_ARRAY = [WELL_KNOW_EMPTY_ARRAY];
+  // protected tables: SymbolTable[] = [];
+  // protected handles: number[] = [];
+  // protected resolved: unknown[] = [];
+  // protected numbers: number[] = [];
+  // protected others: unknown[] = [];
```
rwjblue, line 26:
```suggestion
```

packages/@glimmer/program/lib/constants.ts
```diff
@@ -51,45 +60,29 @@ export class WriteOnlyConstants implements CompileTimeConstants {
...
   toPool(): ConstantPool {
     return {
-      strings: this.strings,
-      arrays: this.arrays,
-      handles: this.handles,
-      numbers: this.numbers,
+      strings: [],
+      arrays: [],
+      handles: [],
+      numbers: [],
```
rwjblue, line 85: Hmm, why is this returning empty arrays?

PR #1123, opened 2020-07-31
[REFACTOR] Simplifies immediate and handle encoding

pzuraq:
Simplifies immediate and constant handle encoding in the following ways:

1. Immediates are stored as negative numbers instead of positive, as
  handles are much more common in general.

2. `true`, `false`, `undefined`, and `null` are the first four negatives.
  The remaining negatives are divided between positive numbers and
  negative, with positives taking up the higher negatives since they're
  expected to be more common.

3. Constants in the constant pool now all exist in a single array of
  values. This makes lookups and storage code simpler overall and
  appears to have a positive impact on performance (per the last PR).

4. Handles are no longer divided between string and number handles,
  since all constants are pulled from the same pool. This allows us to
  avoid encoding/decoding handles in general.

Benchmark results: [artifact-33.pdf](https://github.com/glimmerjs/glimmer-vm/files/5007813/artifact-33.pdf)

### DIFF
--- a/packages/@glimmer/bundle-compiler/lib/debug-constants.ts
+++ b/packages/@glimmer/bundle-compiler/lib/debug-constants.ts
@@ -1,46 +1,24 @@
 import { WriteOnlyConstants } from '@glimmer/program';
-import { assert } from '@glimmer/util';
 import { RuntimeConstants } from '@glimmer/interfaces';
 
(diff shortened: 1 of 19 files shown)


pzuraq:
Superceded by #1131

PR #1124, opened 2020-07-31
[REFACTOR] Simplify and warm tag computation

pzuraq:
Simplifies tag computation by using the same `subtag` field for both
combinator and updatable tags, warms the compute function. In some cases
the engine optimizes COMPUTE before it sees a combinator or updatable
tag, and specifically a updatable tag with a `subtagBufferCache`. We
want to make sure it doesn't do that.

### DIFF
--- a/packages/@glimmer/validator/lib/validators.ts
+++ b/packages/@glimmer/validator/lib/validators.ts
@@ -100,9 +100,7 @@ class MonomorphicTagImpl<T extends MonomorphicTagTypes = MonomorphicTagTypes> {
   private lastValue = INITIAL;
 
   private isUpdating = false;
-  private subtags: Tag[] | null = null;
-
-  private subtag: Tag | null = null;
+  private subtag: Tag | Tag[] | null = null;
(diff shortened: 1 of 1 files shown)


packages/@glimmer/validator/lib/validators.ts
```diff
@@ -125,24 +123,24 @@ class MonomorphicTagImpl<T extends MonomorphicTagTypes = MonomorphicTagTypes> {
       this.lastChecked = $REVISION;
 
       try {
-        let { subtags, subtag, subtagBufferCache, lastValue, revision } = this;
+        let { subtag, subtagBufferCache, lastValue, revision } = this;
```
krisselden, line 126: don't read subtagBufferCache and lastValue here, they aren't used.  They are only used in one spot, just use this.subtagBufferCache and this.lastValue

pzuraq: 👍

pzuraq:
Merging as test failures are an issue with BrowserStack, and this change doesn't do anything that could affect IE11.

PR #1131, opened 2020-08-10
[REFACTOR] Simplifies immediate and handle encoding

pzuraq:
Simplifies immediate and constant handle encoding in the following ways:

1. Immediates are stored as negative numbers instead of positive, as
  handles are much more common in general.

2. `true`, `false`, `undefined`, and `null` are stored as handles
  instead of immediates. These handles are predefined/prepopulated, and
  well known ahead of time.

3. Constants in the constant pool now all exist in a single array of
  values. This makes lookups and storage code simpler overall and
  appears to have a positive impact on performance (per the last PR).

4. Handles are no longer divided between string and number handles,
  since all constants are pulled from the same pool. This allows us to
  avoid encoding/decoding handles in general.

5. We now use different methods for pushing and popping handles and immediates on the stack, namely `pushJs`, `pushSmallInt`, `pushNull`, etc. and `popJs`, `peekJs`, `popSmallInt`, etc. 

Benchmark results: 
[artifact-65.pdf](https://github.com/glimmerjs/glimmer-vm/files/5053922/artifact-65.pdf)

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/stack.ts
+++ b/packages/@glimmer/runtime/lib/vm/stack.ts
@@ -1,11 +1,14 @@
 import { LOCAL_DEBUG } from '@glimmer/local-debug-flags';
 import {
-  decodeImmediate,
   isSmallInt,
   encodeImmediate,
   encodeHandle,
   decodeHandle,
+  constants,
+  assert,
+  decodeImmediate,
   isHandle,
+  ImmediateConstants,
 } from '@glimmer/util';
 import { Stack as WasmStack } from '@glimmer/low-level';
 import { MachineRegister, $sp, $fp } from '@glimmer/vm';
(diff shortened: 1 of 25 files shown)


packages/@glimmer/runtime/lib/vm/stack.ts
```diff
@@ -47,39 +42,41 @@ export class InnerStack {
...
-    }
-  }
-
   writeJs(pos: number, value: unknown): void {
     let idx = this.js.length;
     this.js.push(value);
     this.inner.writeRaw(pos, encodeHandle(idx));
   }
 
+  writeNumber(pos: number, value: number): void {
```
krisselden, line 51: this requires that things that call writeNumber guard isSmallInt not just typeof x === 'number' and there are a few places that are only doing the typeof check and calling pushNumber

PR #1135, opened 2020-08-19
[REFACTOR] Simplify CapturedArguments

pzuraq:
Now that the VM uses tracking internally, some of the previous structure
of things doesn't make much sense anymore. For CapturedArguments in
particular, they previously had to b their own class structure because
they needed to combine and forward all of the tags of the argument
references. Now, the references will be consumed naturally whenever they
are accessed, so that's not necessary as much anymore.

This PR makes CapturedPositionalArguments a `Reference[]`, and
CapturedNamedArguments a `Record<string, Reference>`, keeping the data
structures as simple as possible and inline with their general usage.
It also exposes `reifyNamed` and `reifyPositional` as convenience
methods for getting all of the values of the references, and it exposes
a `capturedArgs` function for generating captured arguments manually,
which is done several times in Ember.

The interfaces for CapturedArguments use a declared symbol to also make
it so that users cannot generate them themselves. This allows us to
ensure that `CapturedArguments` is always created with the same shape.

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/arguments.ts
+++ b/packages/@glimmer/runtime/lib/vm/arguments.ts
@@ -18,9 +18,9 @@ import {
   Scope,
   BlockArguments,
 } from '@glimmer/interfaces';
-import { PathReference } from '@glimmer/reference';
+import { PathReference, CachedReference, ConstReference } from '@glimmer/reference';
 import { Tag, CONSTANT_TAG } from '@glimmer/validator';
-import { PrimitiveReference, UNDEFINED_REFERENCE } from '../references';
+import { UNDEFINED_REFERENCE } from '../references';
 import { CheckBlockSymbolTable, check, CheckHandle, CheckOption, CheckOr } from '@glimmer/debug';
 import {
   CheckPathReference,
@@ -122,7 +122,7 @@ export class VMArgumentsImpl implements VMArguments {
     let positional = this.positional.length === 0 ? EMPTY_POSITIONAL : this.positional.capture();
     let named = this.named.length === 0 ? EMPTY_NAMED : this.named.capture();
 
-    return new CapturedArgumentsImpl(positional, named, this.length);
+    return { named, positional } as CapturedArguments;
   }
 
   clear(): void {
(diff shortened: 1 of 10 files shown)


packages/@glimmer/runtime/lib/vm/arguments.ts
```diff
@@ -544,21 +454,72 @@ class CapturedBlockArgumentsImpl implements CapturedBlockArguments {
   }
 }
 
-export class CapturedArgumentsImpl implements CapturedArguments {
-  constructor(
-    public positional: CapturedPositionalArguments,
-    public named: CapturedNamedArguments,
-    public length: number
-  ) {}
+export function capturedArgs(named: Dict<PathReference>, positional: PathReference[]) {
```
rwjblue, line 457: The name here seems a tad odd to me, shouldn't it be `captureArgs`?

pzuraq: maybe `createCapturedArgs`? It's not exactly capturing them, it's making them manually

packages/@glimmer/runtime/lib/vm/arguments.ts
```diff
@@ -169,8 +169,8 @@ export class PositionalArgumentsImpl implements PositionalArguments {
     return check(stack.get(position, base), CheckPathReference) as T;
   }
 
-  capture(): CapturedPositionalArgumentsImpl {
-    return new CapturedPositionalArgumentsImpl(this.references);
+  capture(): CapturedPositionalArguments {
+    return this.references as CapturedPositionalArguments;
```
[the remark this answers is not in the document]
pzuraq: `CapturedPositionalArgs, `CapturedNamedArgs`, and `CapturedArgs` all have a fake symbol as part of the interface. The idea being, you must use one of these APIs to make them, you can't just make a dictionary and use it as arguments. We don't want to actually make or assign the symbol because that would be costly, so we use `as` to typecast it in the creation APIs.

packages/@glimmer/runtime/lib/vm/arguments.ts
rwjblue: I see, ya the thing that seemed odd was the past tense phrasing. `createCapturedArgs` seems good to me.

pzuraq:
Benchmark results: 
[results.pdf](https://github.com/glimmerjs/glimmer-vm/files/5099663/results.pdf)

PR #1138, opened 2020-08-20
[REFACTOR] Extract environment variables into global context

pzuraq:
There are a number of environment variables that are currently threaded
through glimmer that are:

1. Only environment variables for compatibility reasons (e.g. legacy
   Ember compat)
2. Never expected to change over the lifetime of an append
3. Never expected to differ between different apps or render trees in
   the same context.

This PR consolidates these into a single global context, exported from
`@glimmer/global-context`, which is set once by the embedding
environment before rendering. It also upstreams the logic for
`protocolForUrl`, which was not specific to Ember, and simplifies
the `attributeFor` logic to be as minimal as possible an interface for
Ember.


packages/@glimmer/global-context/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@glimmer/global-context",
+  "version": "0.58.0",
+  "repository": "https://github.com/glimmerjs/glimmer-vm/tree/master/packages/@glimmer/global-context",
```
rwjblue, line 4: We should update these (in a future PR) to something like:

```json
{
  "repository": {
    "type" : "git",
    "url" : "https://github.com/glimmerjs/glimmer-vm.git",
    "directory": "packages/@glimmer/global-context"
  }
}
```

https://docs.npmjs.com/configuring-npm/package-json.html#repository

packages/@glimmer/global-context/package.json
```diff
@@ -0,0 +1,8 @@
...
+  "dependencies": {
+    "@glimmer/env": "^0.1.7"
+  }
```
rwjblue, line 7: Adding this will ensure that the first publishing doesn't fail (the default for `npm publish` for scoped packages is private):

```suggestion
  },
  "publishConfig": {
    "access": "public"
  }
```

packages/@glimmer/reference/lib/iterable.ts
```diff
@@ -192,17 +189,17 @@ export class IterableReference extends CachedReference<boolean> {
   }
 
   private createIterator(): OpaqueIterator {
-    let { parentRef, key, env } = this;
+    let { parentRef, key } = this;
 
     let iterable = parentRef.value() as { [Symbol.iterator]: any } | null | false;
 
-    let keyFor = makeKeyFor(key, env.getPath);
+    let keyFor = makeKeyFor(key, getPath);
```
rwjblue, line 196: `makeKeyFor` is not exported, and it is not used again in this file. Any objection to moving to:

```suggestion
    let keyFor = makeKeyFor(key);
```

And simplifying `makeKeyFor` to:

```ts
function makeKeyFor(key: string) {
  // ...snip...
}
```

And `keyForPath` becomes:

```ts
function keyForPath(path: string): KeyFor {
  // ...snip...
}
```

packages/@glimmer/reference/lib/template.ts
```diff
@@ -58,7 +42,7 @@ export abstract class RootReference<T = unknown> extends CachedReference<T>
   private children = dict<PropertyReference>();
   debugLabel?: string;
 
-  constructor(protected env: TemplateReferenceEnvironment) {
+  constructor() {
     super();
   }
```
rwjblue, line 48: The `constructor` is not needed anymore

```suggestion
```

packages/@glimmer/reference/test/template-test.ts
```diff
@@ -161,27 +168,28 @@ module('@glimmer/reference: template', () => {
...
+          assert.equal(key, 'foo');
+
+          return 123;
+        },
+      });
 
       let fooRef = ref.get('foo');
       assert.equal(fooRef.value(), 123);
+
+      testClearGlobalContextOverride!(originalContext);
```
rwjblue, line 185: This should be in a try/finally:

```suggestion
      let originalContext = testOverrideGlobalContext!({
        getProp(obj: unknown, key: string) {
          assert.equal(obj, component);
          assert.equal(key, 'foo');

          return 123;
        },
      });

      try {
        let fooRef = ref.get('foo');
        assert.equal(fooRef.value(), 123);
      } finally {
        testClearGlobalContextOverride!(originalContext);
      }
```

packages/@glimmer/runtime/lib/dom/sanitized-values.ts
```diff
@@ -30,8 +29,30 @@ export function requiresSanitization(tagName: string, attribute: string): boolea
...
+  protocolForUrl = (url: string) => {
+    let protocol = null;
+
+    if (typeof url === 'string') {
+      protocol = nodeURL.parse(url).protocol;
+    }
+
+    return protocol === null ? ':' : protocol;
+  };
+}
```
rwjblue, line 53: I would prefer something like:

```suggestion
if (typeof URL === 'function') {
  // supports Node >= 10 and all browsers except IE11
  protocolForUrl = (urlString: string) => {
    let url = new URL(urlString, typeof document !== 'undefined' ? document.baseURI : undefined);
    
    return url.protocol;
  };
} else if (typeof document !== 'undefined') {
  // supports IE11 environment
  let parsingNode = document.createElement('a');

  protocolForUrl = (url: string) => {
    parsingNode.href = url;
    return parsingNode.protocol;
  };
}
```

This unifies implementation for all modern platforms, and makes it really easy to know what to remove when IE11 support is no longer needed.

Also, see [this note in the Node URL module documentation](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost):

> Use of the legacy url.parse() method is discouraged. Users should use the WHATWG URL API. Because the url.parse() method uses a lenient, non-standard algorithm for parsing URL strings, security issues can be introduced. Specifically, issues with host name spoofing and incorrect handling of usernames and passwords have been identified.

packages/@glimmer/runtime/lib/references.ts
```diff
@@ -1,5 +1,6 @@
 import { Option, Recast } from '@glimmer/interfaces';
 import { ConstReference, PathReference, Reference, CachedReference } from '@glimmer/reference';
+import { toBool as defaultToBool } from '@glimmer/global-context';
```
rwjblue, line 3: Why can't we call this `toBool` (e.g. not alias)? Seems like the only consumer that is trying to customize is the curried component definition (which uses a custom subclass), could we update `ConditionalReference` to not take a second constructor arg and have the subclass assign `this.toBool` itself?

packages/@glimmer/runtime/lib/references.ts
```diff
@@ -62,20 +63,13 @@ export const TRUE_REFERENCE: PrimitiveReference<boolean> = new ValueReference(tr
 export const FALSE_REFERENCE: PrimitiveReference<boolean> = new ValueReference(false);
 
 export class ConditionalReference extends CachedReference<boolean> implements Reference<boolean> {
-  constructor(
-    private inner: Reference<unknown>,
-    private toBool: (value: unknown) => boolean = defaultToBool
-  ) {
+  constructor(private inner: Reference<unknown>, private toBool = defaultToBool) {
     super();
   }
```
rwjblue, line 68: Thinking:

```suggestion
  constructor(private inner: Reference<unknown>) {
    super();
    
    this.toBool = toBool;
  }
```

Then the custom subclass can customize in its constructor after calling super...

packages/@glimmer/global-context/index.ts
```diff
@@ -0,0 +1,197 @@
...
+  warnIfStyleNotTrusted = context.warnIfStyleNotTrusted;
+}
+
+export let assertGlobalContextWasSet: (() => void) | undefined;
+export let testOverrideGlobalContext:
+  | ((context: Partial<GlobalContext> | null) => GlobalContext | null)
+  | undefined;
+export let testClearGlobalContextOverride:
+  | ((context: Partial<GlobalContext> | null) => GlobalContext | null)
+  | undefined;
```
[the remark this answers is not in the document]
pzuraq: I was debating this myself, I think I'll update it to do that after all. It's a little confusing, but this API is primarily for unit testing the (few) places where it matters, so I think it should be fine.

packages/@glimmer/runtime/lib/dom/sanitized-values.ts
pzuraq: So I considered doing this, the main reason I didn't is I was afraid about matching behavior. For instance, `url.parse('image.jpg');` works as a relative url and does not error, but `new URL('image.jpg', undefined);` does not.

So in your solution, in Node, that would throw an error. I think we would need to wrap `new URL` in a try/catch, because if a _valid_ URL is passed it'll parse correctly and everything will be fine, but if a _relative_ URL is passed it'll throw.

packages/@glimmer/runtime/lib/references.ts
pzuraq: There is one `ConditionalReference` usage that currently uses a custom `toBool`, this will be fixed in the Monomorphic References PR (since they'll both just become compute refs), so I decided to keep it as is for the time being.

PR #1142, opened 2020-08-25
[REFACTOR] Monomorphic Reference

pzuraq:
This PR refactors references to be monomorphic, simplifying them
overall. There are now 4 fundamental types of references:

- `Constant`: References that are initialized with a constant value.
  These references are used for roots (`this` in templates) and for
  primitive references.
- `Compute`: References that compute a value dynamically. These are
  essentially  helper references - they represent a tracked computation
  which can update over time based on tracked inputs. They may also be
  updatable, as in the case of property references.
- `Unbound`: References that are constant and will never update. The
  difference between `Unbound` and `Constant` is that `Unbound`
  propogates its constantness to its children, they are also `Unbound`.
- `Invokable`: This reference is used specifically for supporting `mut`
  in Ember. It essentially tells certain consumers - `{{action}}` and
  `{{fn}}`, that if they encounter it, they should invoke its update
  function. This type will likely be deprecated in the future when we
  figure out a better way to expose the box pattern (if we decide to).
  It has to be part of the main Ref implementation due to debug aliases.

In this model, all references can have children, which simplifies things
significantly as well - there's no need to distinguish between
`Reference` and `PathReference`.

 ## Breaking Changes

- Removes `PathReference` interface
- Removes all reference base classes. References must be created using
  one of the create functions exposed from the refs package now.
- Removes the `.value` method from references. The `valueForRef()`
  function must be used to access the value of a reference now.
- Removes the `.isConst` method from references. The `isConstRef()`
  function must be used to determine if a ref is constant now.


pzuraq:
Benchmark results:
[results.pdf](https://github.com/glimmerjs/glimmer-vm/files/5130935/results.pdf)

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -1,95 +1,273 @@
...
-    this.lastValue = currentValue;
+        if (isDict(parent)) {
+          return setProp(parent, path, val);
+        }
+      }
+    );
 
-    return currentValue;
+    if (DEBUG) {
+      child.debugLabel = `${parentRef.debugLabel}.${path}`;
```
rwjblue, line 241: This should probably confirm that `parentRef.debugLabel` is not `false` before chaining

packages/@glimmer/reference/test/iterable-test.ts
```diff
@@ -1,46 +1,38 @@
...
-
 class IterableWrapper {
-  private iterable: IterableReference;
+  private iterable: Reference<{ next(): OpaqueIterationItem | null }>;
 
   constructor(obj: unknown, key = '@identity') {
-    let valueRef = new VolatileReference(obj);
-    this.iterable = new IterableReference(valueRef, key);
+    let valueRef = createComputeRef(() => {
+      consumeTag(VOLATILE_TAG);
```
rwjblue, line 22: Why? This seems like an odd thing to test, since AFAICT none of our real environments actually use VOLTATILE_TAG like this.

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+  });
 
-    compute(): string {
-      computed++;
-      return this.dict.get(this.key);
-    }
-  }
+  hooks.after(() => {
+    testOverrideGlobalContext!(originalContext);
+  });
```
rwjblue, line 58: FWIW, you have this in a few spots now, might be nice to clean this up into something like:

```js
module('asdf', hooks => {
  setupGlobalContext(hooks, {
    // the overriden bits here
  });
});
```

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+      dict.set('foo', 'bar');
+
+      assert.equal(valueForRef(ref), 'bar');
+      assert.equal(valueForRef(ref), 'bar');
+      assert.equal(valueForRef(ref), 'bar');
+
+      assert.strictEqual(count, 4, 'computed');
+    });
+
+    test('CachedReference caches nested computation correctly', assert => {
```
rwjblue, line 148: I don't think this name is correct any more

```suggestion
    test('compute reference caches nested computation correctly', assert => {
```

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+    test('can create invokable refs', assert => {
+      let ref = createComputeRef(
+        () => {},
+        () => {}
+      );
+
+      let invokableRef = createInvokableRef(ref);
+
+      assert.ok(isInvokableRef(invokableRef));
+    });
```
rwjblue, line 255: Should have tests that ensure this functions properly: should be cached, should be able to have children, etc.

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+      assert.equal(setCount, 1, 'set called correct number of times');
+
+      parent.child = new Child();
+
+      assert.equal(valueForRef(valueRef), 123, 'value updated correctly when parent changes');
+      assert.equal(getCount, 4, 'get called correct number of times');
+    });
+  });
+
+  module('unbound ref', () => {
```
rwjblue, line 226: Should included tests for the unbound thing itself being not tracked too (we only have tests for children)

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+        () => {}
+      );
+
+      let invokableRef = createInvokableRef(ref);
+
+      assert.ok(isInvokableRef(invokableRef));
+    });
+  });
+
+  module('read only ref', () => {
```
rwjblue, line 258: Should test the "normal battery of things", this test assumes details about implementation and it shouldn't.

packages/@glimmer/reference/test/references-test.ts
```diff
@@ -37,111 +34,284 @@ class TrackedDict<T> {
...
+
+      let readOnlyRef = createReadOnlyRef(computeRef);
+
+      assert.ok(isUpdatableRef(computeRef), 'original ref is updatable');
+      assert.notOk(isUpdatableRef(readOnlyRef), 'read only ref is not updatable');
+    });
+  });
+
+  if (DEBUG) {
+    module('debugAliasRef', () => {
```
rwjblue, line 279: We need non-DEBUG tests for this too (should do nothing, but function).

packages/@glimmer/debug/lib/opcode-metadata.ts
```diff
@@ -905,16 +905,6 @@ METADATA[Op.ExitList] = {
...
 
-METADATA[Op.PutIterator] = {
-  name: 'PutIterator',
-  mnemonic: 'toiter',
-  before: null,
-  stackChange: 0,
-  ops: [],
-  operands: 0,
-  check: true,
-};
```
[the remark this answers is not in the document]
pzuraq: I forgot about that, updated it and reran the generation instead so it's correct

packages/@glimmer/interfaces/lib/vm-opcodes.d.ts
```diff
@@ -70,10 +70,8 @@ export const enum Op {
   ToBoolean = 71,
   EnterList = 72,
   ExitList = 73,
-  PutIterator = 74,
   Iterate = 75,
   Main = 76,
-  IsComponent = 77,
```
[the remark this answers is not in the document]
pzuraq: Updated

packages/@glimmer/reference/lib/iterable.ts
```diff
@@ -149,49 +154,9 @@ function uniqueKeyFor(keyFor: KeyFor) {
...
-    return item;
-  }
-
-  private createIterator(): OpaqueIterator {
-    let { parentRef, key } = this;
-
-    let iterable = parentRef.value() as { [Symbol.iterator]: any } | null | false;
+export function createIteratorRef(listRef: Reference, key: string) {
+  return createComputeRef(() => {
+    let iterable = valueForRef(listRef) as { [Symbol.iterator]: any } | null | false;
```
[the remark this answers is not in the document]
pzuraq: This isn't new code, but we should make an issue to track fixing this later on

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
```diff
@@ -28,8 +28,7 @@ export function main(): CompileActions {
  */
 export function StdAppend(trusting: boolean): CompileActions {
   return [
-    op(Op.ContentType),
-    SwitchCases(when => {
+    ContentTypeSwitchCases(when => {
```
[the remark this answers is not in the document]
pzuraq: yeah, this seems to be somewhat conventional in the opcode compiler at the moment, but I do think we should rethink that.

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -1,95 +1,273 @@
...
+  update: Option<(value: T) => void> = null,
+  debugLabel: false | string = 'unknown'
+): Reference<T> {
+  let ref = new ReferenceImpl<T>(ReferenceType.Compute);
+
+  ref.compute = compute;
+  ref.update = update;
+
+  if (DEBUG) {
+    ref.debugLabel = `(result of a \`${debugLabel}\` helper)`;
```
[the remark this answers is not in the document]
pzuraq: I think we can add `getDebugName` to helper managers, and let the manager figure out how to pass this along

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -1,95 +1,273 @@
...
+  }
+
+  return ref;
+}
+
+export function createReadOnlyRef(ref: Reference): Reference {
+  if (!isUpdatableRef(ref)) return ref;
+
+  return createComputeRef(() => valueForRef(ref), null, ref.debugLabel);
```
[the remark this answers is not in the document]
pzuraq: we could do that, this was just matching previous behavior

packages/@glimmer/runtime/lib/compiled/opcodes/vm.ts
```diff
@@ -57,12 +67,22 @@ APPEND_OPCODES.add(Op.Primitive, (vm, { op1: primitive }) => {
...
+    ref = UNDEFINED_REFERENCE;
+  } else if (value === null) {
+    ref = NULL_REFERENCE;
+  } else if (value === true) {
+    ref = TRUE_REFERENCE;
+  } else if (value === false) {
+    ref = FALSE_REFERENCE;
+  } else {
+    ref = createPrimitiveRef(value);
+  }
```
rwjblue, line 83: Seems like this is precisely what `switch` is for?

packages/@glimmer/runtime/lib/component/curried-component.ts
```diff
@@ -1,63 +1,86 @@
...
+import { Option, symbol, assert, _WeakSet } from '@glimmer/util';
 import { VMArgumentsImpl } from '../vm/arguments';
+import { capabilityFlagsFrom } from '../capabilities';
+import { ComponentInstance } from '../compiled/opcodes/component';
 
-const CURRIED_COMPONENT_DEFINITION_BRAND: unique symbol = symbol('CURRIED COMPONENT DEFINITION');
+const INNER: unique symbol = symbol('INNER');
+const ARGS: unique symbol = symbol('ARGS');
+
+const CURRIED_COMPONENT_DEFINITIONS = new _WeakSet();
```
rwjblue, line 10: This will break some of the shenanigans that I'm doing over in:

https://github.com/rwjblue/ember-template-invocation-location/blob/4d13448d972b069bdcf4c21293b6a86af618a881/addon/helpers/-template-invocation-info-is-component-or-helper.js#L6-L7

Can you double check / confirm that `isCurriedComponentDefinition` is exported from `@glimmer/runtime` itself?

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -1,95 +1,273 @@
...
+
+  if (children === null) {
+    children = parentRef.children = new Map();
+  } else {
+    child = children.get(path)!;
+
+    if (child !== undefined) {
+      if (DEBUG) {
+        // TODO: Make this return a new reference for debug tracing
+        return child;
```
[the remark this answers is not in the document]
pzuraq: Ah, yes, but I ended up doing this a different way with `createDebugAliasRef`

packages/@glimmer/reference/lib/reference.ts
pzuraq: it will only every be `false` in production, is the idea, so it should always be a string here. I guess we could assert that here

packages/@glimmer/reference/test/iterable-test.ts
pzuraq: yeah, I think this should probably be updated at some point, but for now it's just a quick way to avoid caching

packages/@glimmer/runtime/lib/compiled/opcodes/vm.ts
pzuraq: @krisselden has said we should avoid using `switch` with non-numbers, because it's not very efficient.

packages/@glimmer/reference/lib/iterable.ts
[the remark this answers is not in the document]
pzuraq: https://github.com/glimmerjs/glimmer-vm/issues/1143

packages/@glimmer/util/lib/weak-set.ts
```diff
@@ -0,0 +1,18 @@
+export default (typeof WeakSet === 'function'
```
rwjblue, line 1: We should make sure that Ember drops its own `WeakSet` polyfill over in the integration PR.

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -1,95 +1,273 @@
...
+
+  let { lastRevision } = ref;
+  let lastValue;
+
+  if (tag === null || !validateTag(tag, lastRevision)) {
+    let { compute } = ref;
+
+    beginTrackFrame(DEBUG && ref.debugLabel);
+    lastValue = ref.lastValue = compute!();
+    tag = ref.tag = endTrackFrame();
```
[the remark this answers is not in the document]
rwjblue: I think this is resolved by the move to `track(()=> {})`

packages/@glimmer/reference/lib/reference.ts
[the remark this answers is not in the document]
rwjblue: OK, I'll try to do a high level test for that PR and confirm that it passes once this lands.

packages/@glimmer/runtime/lib/compiled/opcodes/vm.ts
rwjblue: O_o ooooookkkkkiiiiieeeeeeeee ddoiooooooookkkkkiiiieeee

PR #1144, opened 2020-08-26
Add tracerbench github action

krisselden:
* Adds a tracerbench-compare-action for krausest render 1000 rows.
* Builds the benchmark app against dist and control-dist builds


lib/local-linker/index.js
```diff
@@ -13,13 +16,45 @@ module.exports = {
...
 
     symlinkOrCopy(`${dir}/@glimmer`, `${dir}/node_modules/@glimmer`);
+
+    if (!dir.endsWith('control-dist') && process.env.EMBER_ENV === 'production') {
+      await buildAll();
+    }
   },
 };
+
+function findOutputPath() {
```
rwjblue, line 42: We'll remove this once we've updated to ember-cli@3.21.0 in a week or so.

PR #1146, opened 2020-08-27
Update Prettier to 2.1

rwjblue:


rwjblue:
Easiest to review the first commit on its own, the second commit is specifically _just_ `eslint . --fix`.

PR #1150, opened 2020-09-02
[WIP] Refactor precompiler

wycats:
The TL;DR of this PR is that it refactors the precompiler to remove a lot of legacy cruft and set it up better for strict mode. This PR also upgrades to TypeScript 4.

I need to do a handful of cleanups still before it's ready for merge, and the PR should be ready in the next day or two.


.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "singleQuote": true,
   "trailingComma": "es5",
-  "printWidth": 100
+  "printWidth": 100,
+  "arrowParens": "avoid"
```
rwjblue, line 5: I don't think we should change this setting.

package.json
```diff
@@ -98,7 +98,10 @@
     "semver": "^6.3.0",
     "testem-failure-only-reporter": "^0.0.1",
     "toml": "^3.0.0",
-    "typescript": "3.7.5"
+    "typescript": "^4.0.2"
+  },
+  "resolutions": {
+    "typescript": "^4.0.2"
```
rwjblue, line 104: We should not need a resolution, what forced you to add this?

packages/@glimmer/compiler/lib/PASSES.md
```diff
@@ -0,0 +1,29 @@
+---
+noteId: "5d0de690ecd611ea8388874e1424078a"
```
rwjblue, line 2: I'm curious, what does this front matter do?

packages/@glimmer/compiler/lib/PASSES.md
```diff
@@ -0,0 +1,29 @@
+---
```
rwjblue, line 1: This document is great, thank you for putting it together! 

Maybe it should move to `packages/@glimmer/compiler/README.md`?

wycats:
> Can you please split the unrelated package updates apart from the main changes? There is a very large number of seemingly unrelated changes in this PR. I would expect at least 3 PR's for this set of changes:

Totally fair! On it!

> eslint and other linting related packge updates
> typescript 4 upgrade

These have to go together, as the eslint upgrades were caused by the TS upgrades (and old eslint doesn't work with the new TS).

> precompiler refactor

👍

.prettierrc
wycats: I agree not to change it right now, but I disagree with this decision 😄

package.json
wycats: broccoli-typescript-compiler, iirc.

I did it preemptively during the RC period, because prereleases don't resolve correctly. More recently, I looked in the lockfile and saw that something was asking for ~3.9.something and kept it in.

packages/@glimmer/compiler/lib/PASSES.md
wycats: Seems legit. My main concern was that there's more going on in the whole package than just the passes, and wanted to be clear about what each directory is doing, but I'd happily move it into a root README (or perhaps a link off the root README?)

rwjblue:
Ready for a bit of a rebase now that #1154 landed.

PR #1152, opened 2020-09-02
[BUGFIX] Prevent accidental dynamic component lookup

pzuraq:
Dynamic components invoked via `{{component}}` share a codepath with
ones invoked via `<this.foo>`, and because of that currently if
`<this.foo>` is a referring to a string, it will dynamically lookup the
component. This is not expected behavior, and this PR prevents it by
using a different, more restricted opcode for these lookups.

Closes https://github.com/glimmerjs/glimmer-vm/pull/1149
Addresses https://github.com/emberjs/ember.js/issues/17744


packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -174,15 +180,24 @@ APPEND_OPCODES.add(Op.ResolveDynamicComponent, (vm, { op1: _meta }) => {
...
 
   stack.pushJs(definition);
 });
 
+APPEND_OPCODES.add(Op.ResolveCurriedComponent, (vm) => {
+  let stack = vm.stack;
+  let definition = check(
+    valueForRef(check(stack.popJs(), CheckReference)),
+    CheckCurriedComponentDefinition
+  );
```
rwjblue, line 195: This needs to be done in DEBUG mode and actually throw an error that is **not** stripped during publishing. Ember and Glimmer.js users must be able to see this error.

```suggestion
  let stack = vm.stack;
  let value = valueForRef(check(stack.popJs(), CheckReference));
  if (DEBUG && !isCurriedComponentDefinition(value)) {
    throw new Error(`A curried component definition. You may have accidentally done <this.dynamicComponent>, where dynamicComponent was a string instead of a curried component definition. You must use the {{component}} helper to create a component definition when invoking dynamically.`);
  }
  
  let definition = value as CurriedComponentDefinition;
```

packages/@glimmer/runtime/lib/compiled/opcodes/-debug-strip.ts
```diff
@@ -106,6 +107,20 @@ export const CheckComponentDefinition: Checker<ComponentDefinition> = CheckInter
...
+  type!: CurriedComponentDefinition;
+
+  validate(value: unknown): value is CurriedComponentDefinition {
+    return isCurriedComponentDefinition(value);
+  }
+
+  expected(): string {
+    return `A curried component definition. You may have accidentally done <this.dynamicComponent>, where dynamicComponent was a string instead of a curried component definition. You must use the {{component}} helper to create a component definition when invoking dynamically.`;
+  }
+}
```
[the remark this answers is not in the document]
pzuraq: mm, good point. Will add an extra error message in the opcode itself instead.

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -174,12 +180,26 @@ APPEND_OPCODES.add(Op.ResolveDynamicComponent, (vm, { op1: _meta }) => {
...
+  stack.pushJs(definition);
+});
+
+APPEND_OPCODES.add(Op.ResolveCurriedComponent, (vm) => {
+  let stack = vm.stack;
+  let value = valueForRef(check(stack.popJs(), CheckReference));
+
+  if (DEBUG && !isCurriedComponentDefinition(value)) {
+    throw new Error(
+      `A curried component definition. You may have accidentally done <this.dynamicComponent>, where dynamicComponent was a string instead of a curried component definition. You must use the {{component}} helper to create a component definition when invoking dynamically.`
```
lifeart, line 196: is it possible to use real path / value for error?

one more use case is:

```hbs
{{#let "my-component" as |AnyComponent|}}
   <AnyComponent />
{{/let}}
```

pzuraq: Updated the error message to reference the debugLabel of the reference, which should give better context in general.

Re: the use case, that is what we are trying to prevent, and is not a valid use case in general. You should convert forward to:

```hbs
{{#let (hash AnyComponent=(component "my-component")) as |ctx|}}
   <ctx.AnyComponent />
{{/let}}
```

lifeart: I know it not valid, my comment point is to have debug information pointing to such usage :)
(error may appear not only in `this.foo` invocation)

PR #1154, opened 2020-09-02
Upgrade to TypeScript 4.0

wycats:
This also updates the eslint-typescript packages

TypeScript 4.0 invalidated the way that we were casting SimpleDOM to
browser DOM and vice versa. This commit changes our casting strategy so
that casts go through additional checks that accomodate TypeScript's new
stricter requirements.

In general, this was a good change: it adds some checks to the blind
casts we were making, and caught at least one genuine mistake in our
casts (casting a textarea to input).

However, these checks add some minor overhead and should be unnecessary
in production. This commit structures the casts so they can be stripped
out relatively easily in the future, and we should do so.


rwjblue:
Looks like some linting failures (likely due to typescript-eslint changing around the camelcase rule).

krisselden:
IE 11 is failing with "Error: cannot cast a SimpleElement into an BrowserElementUtils outside of a browser environment"

wycats:
> IE 11 is failing with "Error: cannot cast a SimpleElement into an BrowserElementUtils outside of a browser environment"

Fix incoming.

rwjblue:
Once we get the linting failures fixed up, I'll work on ensuring the `cast` infrastructure is not used in production builds. At the moment, this introduces a ~ 1% performance regression (and I think we can fix that fairly easily with stripping out the type assertion related code).

rwjblue:
Just pushed two changes: 

* Move cast into @glimmer/util
* Ensure `cast` is stripped in production builds

After those two tweaks, I confirmed that `cast` was indeed stripped when doing `ember build -prod`.

rwjblue:
Just rebased to pull in the changes to #1156 so the PerformanceCheck CI job can run properly.

rwjblue:
FYI - Labeled as `breaking` (to be safe)...

wycats:
@rwjblue thanks for doing the heavy lifting!

PR #1158, opened 2020-09-07
attempt to add failing test to nested p tags rehydration

lifeart:
related issue: https://github.com/emberjs/ember.js/issues/19114

image

may be related: https://stackoverflow.com/questions/8460993/p-end-tag-p-is-not-needed-in-html

assert bubbling from 

image

According to current parser implementation, unclosed `<p>` throw error: https://astexplorer.net/#/gist/0e4452d39e7a5d86ac852414ec56fa99/054dedaf474d1710d2fabcc7e0e93a3e63692c4f


lifeart:
closing it as old

PR #1159, opened 2020-09-14
validate if in-element has already been transformed

patricklx:
I was using it for some preprocessing, and then it broke at the end when glimmer was parsing the final template (only if it had in-element)

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -450,7 +450,7 @@ export default class TemplateCompiler implements Processor<InputOps> {
       let { key, value } = pairs[i];
 
       if (isInElement) {
-        if (key === 'guid' && !value.value.startsWith('%cursor:')) {
+        if (key === 'guid' && value.type === 'StringLiteral' && !value.value.startsWith('%cursor:')) {
           throw new SyntaxError(
             `Cannot pass \`guid\` to \`{{#in-element}}\` on line ${value.loc.start.line}.`,
             value.loc
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -450,7 +450,7 @@ export default class TemplateCompiler implements Processor<InputOps> {
       let { key, value } = pairs[i];
 
       if (isInElement) {
-        if (key === 'guid') {
+        if (key === 'guid' && !value.value.startsWith('%cursor:')) {
           throw new SyntaxError(
             `Cannot pass \`guid\` to \`{{#in-element}}\` on line ${value.loc.start.line}.`,
             value.loc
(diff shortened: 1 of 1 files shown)


lifeart:
@patricklx 
image

lifeart:
@patricklx thank you for changes! Could you fix fomatting using `--fix` flag for eslint?

lifeart (approved):

packages/@glimmer/compiler/lib/template-compiler.ts
```diff
@@ -450,7 +450,11 @@ export default class TemplateCompiler implements Processor<InputOps> {
       let { key, value } = pairs[i];
 
       if (isInElement) {
-        if (key === 'guid') {
+        if (
+          key === 'guid' &&
+          value.type === 'StringLiteral' &&
+          !value.value.startsWith('%cursor:')
```
rwjblue, line 456: Hmm, why would we expect `guid` to be passed?

patricklx: this happens when I do glimmer preprocess -> print (at this stage its passing guid) -> preprocess printed result again fails.

rwjblue: Hmm, this only happens when compiling not when parsing / printing. Maybe you are using an older version of glimmer-vm that did this during syntax parsing?

Can you describe the tooling that you are using (and the versions) that hit the issue?

rwjblue: I've created https://github.com/glimmerjs/glimmer-vm/pull/1164 which shows that we can do the following:

```js
import { preprocess, print } from '@glimmer/syntax';

let template = '{{#in-element this.someElement}}Content here{{/in-element}}';
let ast = preprocess(template);
let reprintedOutput = print(ast);

assert.equal(reprintedOutput, template);
```

patricklx: ah, that could be. was trying out custom hbs preprocessing (not ast plugin).
So, i was using the glimmer package directly (v0.42.2).
But then the template was processed by ember it failed to process it.
"ember-source": "^3.21.1"

actually with embroider:
"@embroider/core": "^0.23.0"

rwjblue: > > this happens when I do glimmer preprocess -> print (at this stage its passing guid) -> preprocess printed result again fails.
> 
> Maybe you are using an older version of glimmer-vm that did this during syntax parsing?

Specifically, I think you are probably using a version of glimmer-vm prior to this PR: https://github.com/glimmerjs/glimmer-vm/pull/1086. That landed in `@glimmer/syntax@0.52.1`.

patricklx: just tested again to be sure, but this does still fail:
```
const glimmer = require('@glimmer/syntax');

const result = glimmer.preprocess(`<div>{{#in-element destinationElement~}}
    {{yield yieldHash}}
  {{~/in-element}}</div>`);

const printed = glimmer.print(result);
console.log(printed);

glimmer.preprocess(printed);
```

patricklx: ah, updated to latest version works!

PR #1160, opened 2020-09-18
[REFACTOR] Converts to using @handlebars/parser

pzuraq:
Converts to using the new `@handlebars/parser` package directly instead
of pulling it in from handlebars.js


wycats (approved):
Looks good!

rwjblue:
@pzuraq what is the impact here to Ember? Seems like we no longer bundle Handlebars (at least the "inliner" is removed), can you setup an integration branch on the Ember side to account for these changes?

pzuraq:
Yes, it should mean that we just need to bundle this package instead of `handlebars.js`. We should do a release before landing @wycats branch to integrate this, then merge that PR

PR #1162, opened 2020-09-18
Simplify `SimpleDOM` casting (allow for proper production mode stripping)

pzuraq:
The new `cast` function added for the recent TS upgrade has a couple of
issues:

1. It creates wrapping objects which modify the values that are
   wrapped/casted. Removing `cast()`, as the production build currently
   does, breaks because of this.
2. These wrapping objects conflate casting _to_ SimpleDOM and casting
   _away_ from it, into browser APIs.

The new system makes `cast` into two pure validation utility functions,
which either downlevels to SimpleDOM (without any checks, as SimpleDOM
is a subset of the real DOM), or uplevels to DOM with appropriate checks
(e.g. are we actually in the browser? Is the node a DOM node? Is it the
expected type of DOM node?).

### DIFF
--- a/packages/@glimmer/benchmark-env/src/benchmark/on-modifier.ts
+++ b/packages/@glimmer/benchmark-env/src/benchmark/on-modifier.ts
@@ -1,6 +1,6 @@
 import { ModifierManager, VMArguments } from '@glimmer/interfaces';
 import { Reference, valueForRef } from '@glimmer/reference';
-import { cast } from '@glimmer/util';
+import { castToBrowser } from '@glimmer/util';
 import { createUpdatableTag } from '@glimmer/validator';
 import { SimpleElement } from '@simple-dom/interface';
 
@@ -30,13 +30,13 @@ class OnModifierManager implements ModifierManager<OnModifierState, null> {
   install(state: OnModifierState) {
     const name = valueForRef(state.nameRef);
     const listener = valueForRef(state.listenerRef);
-    cast(state.element, 'ELEMENT').addEventListener(name, listener);
+    castToBrowser(state.element, 'ELEMENT').addEventListener(name, listener);
     state.listener = listener;
     state.name = name;
   }
 
   update(state: OnModifierState) {
-    const element = cast(state.element, 'ELEMENT');
+    const element = castToBrowser(state.element, 'ELEMENT');
     const name = valueForRef(state.nameRef);
     const listener = valueForRef(state.listenerRef);
     if (name !== state.name || listener !== state.listener) {
--- a/packages/@glimmer/util/lib/simple-cast.ts
+++ b/packages/@glimmer/util/lib/simple-cast.ts
@@ -1,139 +1,5 @@
 import { unreachable } from './platform-utils';
-import {
-  Namespace,
-  NodeType,
-  SimpleDocument,
-  SimpleElement,
-  SimpleNode,
-} from '@simple-dom/interface';
-
-/**
- * This class is used to return a result from casting when only a simple
- * node is available.
- */
-class SimpleNodeUtils<S extends SimpleNode = SimpleNode> {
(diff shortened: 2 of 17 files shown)


rwjblue:
> The new cast function added for the recent TS upgrade has a couple of issues

Where/when were these issues identified?

> It creates wrapping objects which modify the values that are wrapped/casted. Removing cast(), as the production build currently does, breaks because of this.

Does this indicate a lack of testing? Specifically, it would seem that we _probably_ can't be testing the production builds (since this issue would have likely caused those tests to fail), right?

packages/@glimmer/util/lib/simple-cast.ts
```diff
@@ -244,22 +101,26 @@ function isElement(node: Node | SimpleNode | SimpleElement): node is Element | S
   return node.nodeType === NodeType.ELEMENT_NODE;
 }
 
-function isBrowserNode(node: Node | SimpleNode): node is Node {
-  return typeof document !== undefined && node.ownerDocument === document;
-}
+export function checkNode<S extends SugaryNodeCheck>(
```
rwjblue, line 104: Just to be clear, this is a `DEBUG` only utility (or actually, it may be a `LOCAL_DEBUG`?) right?

rwjblue (approved):

rwjblue:
Hmm. I don't fully grok why this would impact the performance check (~2% / 21ms slower). 🤔 

> A statistical analysis test was used (Wilcoxon Rank-Sum Test) to determine that the results are significant meaning they are worth looking at. A statistics estimator (Hodges–Lehmann estimator) was used to determine "Experiment" is slower by 21 ms. TracerBench is 95% confident it is slower between 3 ms to 38 ms based on 50 samples using a confidence interval.

packages/@glimmer/benchmark-env/src/benchmark/on-modifier.ts
```diff
@@ -30,13 +30,13 @@ class OnModifierManager implements ModifierManager<OnModifierState, null> {
   install(state: OnModifierState) {
     const name = valueForRef(state.nameRef);
     const listener = valueForRef(state.listenerRef);
-    cast(state.element, 'ELEMENT').addEventListener(name, listener);
+    castToBrowser(state.element, 'ELEMENT').addEventListener(name, listener);
```
rwjblue, line 33: Is this removed for `@glimmer/benchmark-env` performance test also?

packages/@glimmer/util/lib/simple-cast.ts
pzuraq: Yes, it is LOCAL_DEBUG

packages/@glimmer/benchmark-env/src/benchmark/on-modifier.ts
pzuraq: Ah, it may not be, we can check

pzuraq:
Production builds are not tested, no. It’s a gap we should probably prioritize filling

rwjblue:
> Production builds are not tested, no. It’s a gap we should probably prioritize filling

👍 totally agree, created https://github.com/glimmerjs/glimmer-vm/issues/1165 to track that

pzuraq:
It looks like there was some code that was not running at all in production builds due to the casting failure, specfically:

https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/runtime/lib/compat/svg-inner-html-fix.ts#L31

https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/runtime/lib/compat/text-node-merging-fix.ts#L28

Tested by disabling these on this branch, and it seemed to move the needle back to mostly no stat-sig change, so I think that may have been the cause for the perf difference. This was clearly a bug, so I think it makes sense to merge and if we need to fixup perf in the future, find somewhere else to fix things.

PR #1163, opened 2020-09-21
[BUGFIX] args lost when passed to {{component}}

rreckonerr:
As reported by @samselikoff  in emberjs/ember.js#19061,
arguments passed to {{component}} helper, except for the first one, are ignored.

Component
```hbs
{{component "reproduce" one="first" two="second" three="third" four="fourth"}}
```
Resulted in 
```js
hash = [["@one", "two", "three"], ["first", "second", "third"]]
```
instead of
```js
hash = [["@one", "@two", "@three"], ["first", "second", "third"]]
```

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/emberish-components.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/emberish-components.ts
@@ -50,25 +50,37 @@ export class EmberishComponentTests extends RenderTest {
 
   @test({ kind: 'glimmer' })
   'Static block component helper'() {
-    this.registerComponent('Glimmer', 'A', 'A {{#component "B" arg1=@one}}{{/component}}');
-    this.registerComponent('Glimmer', 'B', 'B {{@arg1}}');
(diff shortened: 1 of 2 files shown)


pzuraq (approved):
This looks great! Thank you for digging in and solving this issue, we really appreciate it! This is a great PR also, tests look fantastic, everything is solid. Going to merge right away.

@rwjblue do you think we need to backport this for Ember LTS? I'm not sure if it impacted much existing code, but may make sense to.

rreckonerr:
Thank you! 
I think it makes sense to backport it! According to git blame bug was brought in two years ago https://github.com/glimmerjs/glimmer-vm/commit/72406388bdabe534a797a0d1e97fef9a359beb1d

rwjblue:
Ya, backporting seems totally fine/safe to me. In order to backport, we'd need PRs targeting the `release-0-56` branch (for Ember 3.20 / 3.21) and the `release-0-59` branch (for Ember 3.22 beta).

PR #1169, opened 2020-09-30
[FEAT] Adds better logging for render errors

pzuraq:
Screen Shot 2020-09-29 at 6 23 58 PM

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/components.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/components.ts
@@ -886,4 +886,52 @@ export class BasicComponents extends RenderTest {
 
     this.assertHTML('', 'destroys correctly');
   }
+
+  @test({ kind: 'fragment' })
+  'throwing an error during rendering gives a readable error stack'(assert: Assert) {
+    let originalConsoleError = console.error;
+
+    console.error = (message: string) => {
+      this.assert.ok(
+        message.match(/Error occurred while rendering:(\n\nBar\n {2}Foo)?/),
+        'message logged'
+      );
+    };
+
+    try {
+      assert.expect(7);
+
+      this.registerComponent(
+        'Glimmer',
+        'Foo',
+        'Hello',
+        class extends EmberishGlimmerComponent {
+          constructor(args: EmberishGlimmerArgs) {
+            super(args);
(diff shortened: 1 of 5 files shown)


rwjblue (commented):
Chatted with @pzuraq offline, we should migrate this away from `try`/`catch` back to using `try`/`finally`. As a `try`/`catch` we somewhat fundamentally break developer ergonomics (when using "break on uncaught exceptions" it should stop in userland helper code, not our own internals).

The original need that caused this to be moved to `try`/`catch` was the deopts that were being triggered in that code that should only be ran when there is an actual exception. We can avoid that same issue by doing something akin to this instead:

```js
let didError = true;
try {
  doSomething();
  didError = false;
} finally {
  if (didError) {
    console.log('error yo!')
  }
}
```

Doing this avoids the dev ergonomics issues **and** the deopt that was the original reason for moving to `try`/`catch`.

---

In this context, I think we should change this PR from assigning `e.message` to directly `console.log`'ing instead.

rwjblue:
For those that are curious, here is a smallish codepen that demos the dev ergonomics issue I was referring to: https://codepen.io/rwjblue/full/PozYKqw.

When you run it, make sure to have DevTools open and "Pause on uncaught exceptions" set in the sources tab. The execution **should** pause in the `someBadFunction` implementation where `throw new Error()` is called, but it is actually pausing inside the `catch` block where we `throw e`.

rwjblue:
https://github.com/glimmerjs/glimmer-vm/pull/1172 addresses the removal of `try`/`catch` here, this will need a rebase after that lands

rwjblue:
kk, ready for a rebase / rework now @pzuraq

packages/@glimmer/integration-tests/lib/suites/components.ts
```diff
@@ -886,4 +886,50 @@ export class BasicComponents extends RenderTest {
...
+
+    this.assertHTML(
+      '<div class="first"></div><div class="second"></div><!---->',
+      'values rendered before the error rendered correctly'
+    );
+    this.destroy();
+
+    this.assertHTML('', 'destroys correctly');
+
+    console.error = originalConsoleError;
```
rwjblue, line 933: This needs to be a `finally`

PR #1168, opened 2020-09-24
Strict mode precompiler

wycats:
This PR needs a detailed commit message and is currently being reviewed by @pzuraq.

The very short TL;DR is that it refreshes `@glimmer/syntax` and `@glimmer/compiler` to clean up many long-standing issues, provide more reliable source locations to all nodes, and prepare for strict mode.

I did not intentionally introduce any breaking changes in this pull request, and attempted to document the architecture far more fully than the previous version.

PR #1170, opened 2020-10-06
Refactor `@glimmer/syntax` and `@glimmer/compiler` to enable future strict mode functionality.

pzuraq:
Rebases and continues the work of #1168 

This PR refreshes `@glimmer/syntax` and `@glimmer/compiler` in preparation for strict mode. It makes a number of changes:

1. Creates a new syntax tree, ASTv2, in `@glimmer/syntax`. The original syntax tree is still exposed externally via the existing `@glimmer/syntax` public APIs, and via AST transforms. The plan is to eventually update these, but to land these refactors first so that we aren't forcing a large amount of churn on the ecosystem. The new syntax tree parses into strict mode or loose mode directly, with strict mode emitting strict variable lookup and loose mode emitting resolver based lookup opcodes.

2. Updates the compiler to perform a number of explicit passes on the tree before turning it into the wire format. These passes are documented in their respective READMEs.

3. Updates the wireformat and opcode compiler in a few ways:
  - Removes the `InElement` macro in favor of an explicit opcode
  - Extracts more values into `upvars` in preparation for strict mode
  - Refactors the existing `Get` opcode types to align with strict mode
  - Refactors the wireformat to use arrays from blocks instead of objects, resulting in a much more efficient format overall.

Test results for the new wire format:

```
{
  master: {
    version: 'master',
    original: 10668459,
    compiled: 14711999,
    brotli: 3053244,
    gzip: 3053285
  },
  'strict-precompiler': {
    version: 'strict-precompiler',
    original: 10668459,
    compiled: 13805388,
    brotli: 2939812,
    gzip: 2939843
  }
}
```

## Breaking Changes

Breakage is primarily in `@glimmer/syntax`, in the AST. Changes include:

- `loc` is now required in builders
- `loc` is now a SourceSpan object, which has a different public API than the previous `SourceLocation` object
- You must use builders in general to build values in template transforms now


rwjblue (commented):
If `compiler/lib/builder/builder-interface.ts` is only used in tests, can we move it out of `lib/`?

rwjblue (approved):

PR #1171, opened 2020-10-10
Refactor resolver for strict

pzuraq:
Contains a number of refactors including:

- Cleaning up the opcode-compiler in general
- Removing template macros
- Unifying the resolution flow in the opcode compiler
- Refactoring the opcode compiler to be a single-pass operation
- Cleaning up the program heap

See the individual commits for more detailed descriptions of each individual refactor.

## Benchmarks

Initial benchmark of these changes showed a ~2% improvement in a large scale production application:

[artifact-73.pdf](https://github.com/glimmerjs/glimmer-vm/files/5562250/artifact-73.pdf)

## Breaking Changes

- `StdlibOperand` -> `StdLibOperand`
- Template macros have been removed
- `SyntaxCompilationContext` has been removed
- `syntaxCompilationContext` -> `programCompilationContext`
- Keywords are explicitly reserved now - attempting to use a keyword in a way its not meant to be used will throw an error

Ember integration PR: https://github.com/emberjs/ember.js/pull/19278


packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
```diff
@@ -185,4 +185,53 @@ export const APPEND_KEYWORDS = keywords('Append')
       });
       return Ok(new mir.AppendTextNode({ loc: node.loc, text }));
     },
+  })
+  .kw('component', {
```
rwjblue, line 189: Can you call this change out a bit more explicitly in the description? Specifically, this _probably_ means that:

```hbs
{{component}}
```

Where `{{component}}` is ultimately going to be using `{{this.component}}`. (I mean, its bonkers, not saying we have to support it, but I want to call it out as I'm **certain** _someone_ has done ^).

rwjblue (commented):
* Can you add code (and tests) to confirm that **none** of our keywords can be overridden? I _think_ that might be as simple as preventing yielding a block param of those names? (e.g. `as |else|`, `as |if|`, etc).

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/resolution.ts
```diff
@@ -0,0 +1,186 @@
...
+      throw new Error(
+        'Attempted to resolve a component, but no owner was associated with the template it was being resolved from'
+      );
+    }
+  }
+
+  return (meta as unknown) as RequiredContainingMetadata;
+}
+
+export function resolveComponent(
```
rwjblue, line 74: Moved over from `push-resolution.ts`

packages/@glimmer/opcode-compiler/lib/utils.ts
```diff
@@ -60,16 +50,8 @@ export function namedBlocks(blocks: WireFormat.Core.Blocks, meta: ContainingMeta
...
 
   for (let i = 0; i < keys.length; i++) {
-    out[keys[i]] = compilableBlock(values[i]!, meta);
+    out[keys[i]] = values[i];
   }
 
   return new NamedBlocksImpl(out);
 }
-
-export function lookupLocal(meta: ContainingMetadata, name: string): ExpressionCompileActions {
```
rwjblue: 😢

packages/@glimmer/program/lib/program.ts
```diff
@@ -3,7 +3,7 @@ import {
   SerializedHeap,
   STDLib,
   RuntimeHeap,
-  StdlibOperand,
+  StdLibOperand,
```
rwjblue, line 5: Should call this out as a "breakage"

rwjblue:
I think (other than inline comments) the only other thing is to update description (add detailed explanation of PR, add list of breakages, add link to ember PR, add basic benchmark info).

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/block.ts
```diff
@@ -7,82 +7,473 @@ import { VISIT_EXPRS } from '../visitors/expressions';
...
+    },
+  })
+  .kw('-with-dynamic-vars', {
+    assert(
+      node: ASTv2.InvokeBlock
+    ): Result<{
+      named: ASTv2.NamedArguments;
+    }> {
+      return Ok({ named: node.args.named });
+    },
```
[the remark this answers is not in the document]
pzuraq: Yes, but I think that might be a bit orthogonal here, I made an issue for following up on this: https://github.com/glimmerjs/glimmer-vm/issues/1202

Issue #1176, opened 2020-10-16
Impossible to override some attributes with `...attributes`


rreckonerr:
At least for the `type` attribute.

With glimmer component
```hbs
// my-component.hbs
<button type="button" ...attributes>{{yield}}</button>
```

And using it like this
```hbs
<MyComponent type="submit">Click Me!</MyComponent>
```

Will result in
```html
<button type="button">Click Me!<button>
```

Instead of 
```html
<button type="submit">Click Me!<button>
```

PR #1178, opened 2020-10-16
Fix splattributes handling of type attribute.

thousand:
This PR resolves an issue where a `type` attribute, passed in a component invocation, would not correctly override the `type` attribute in a template when applied with `...attributes`. 

In the existing implementation, we special case the `type` attribute in two places. One is at compile time, when we reorder the `setAttribute` opcodes to apply the `type` attribute last. However, this does not address dynamic cases that are not known at compile time, such as `...attributes` or legacy features such as Ember's `attributeBindings`. A second special case exists in the `ComponentElementOperations` class, which collects aggregate attribute values and applies them once all values are known.

However, currently we do not disable the first special casing when the second will apply. This introduced a bug with `...attributes`, which relies on order to determine attribute precedence. Because we always "moved" the `type` attribute to the last position, values specified in a template could never be overridden by the invocation.

To fix this, we rely on the fact that these cases are mutually exclusive and only apply one or the other. At compile time, if we see `...attributes`, we do not reorder the `type` attribute because we know it will be appropriately re-ordered at runtime.

Closes emberjs/ember.js#18232

### DIFF
--- a/packages/@glimmer/compiler/lib/template-compiler.ts
+++ b/packages/@glimmer/compiler/lib/template-compiler.ts
@@ -129,8 +129,15 @@ export default class TemplateCompiler implements Processor<InputOps> {
       // TODO: Assert no attributes
       let typeAttr: Option<AST.AttrNode> = null;
       let attrs = action.attributes;
+
       for (let i = 0; i < attrs.length; i++) {
-        if (attrs[i].name === 'type') {
+        // Unlike most attributes, the `type` attribute can change how
+        // subsequent attributes are interpreted by the browser. To address
+        // this, in simple cases, we special case the `type` attribute to be set
+        // last. For elements with splattributes, where attribute order affects
+        // precedence, this re-ordering happens at runtime instead.
+        // See https://github.com/glimmerjs/glimmer-vm/pull/726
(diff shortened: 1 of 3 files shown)


tomdale (approved):

rreckonerr:
closes https://github.com/glimmerjs/glimmer-vm/issues/1176
related PR https://github.com/glimmerjs/glimmer-vm/pull/1177

pzuraq (approved):

pzuraq:
Looks good, thanks for fixing this!

chancancode:
Great commit message, thank you for providing the context!

chancancode:
@rreckonerr want to review this PR and check if the test coverage here seems adequate compared to your reproduction? If you spotted anything else that should be incorporated, feel free to make suggestions here or rebase+retarget your PR against this one. 

We will also probably want to add similar tests in ember-source as well as part of landing this in Ember.

tomdale:
@rreckonerr Does this look good to merge?

rreckonerr (approved):
Looks great!
:shipit:

Serabe:
Haven't worked with Ember for a while now, so my review is not needed. Thank you!

PR #1179, opened 2020-10-20
Adding support for partial rehydration

chiragpat:
**Why**
Currently if using rehydration with glimmer-vm, the rehydrate builder expects rehydration to resume from the root component that was rendered on the server. This eventually leads to shipping a whole bunch of static templates to the browser that are never used. This PR enables support for rehydrating from any arbitrary dom node that was generated on the server. This will allow glimmer apps to further improve their rehydration performance by only sending components to the client that have interaction logic.

**How**
- Update rehydrate-builder to insert a opening and closing block next to blocks where we would be rehydrating
- Update rehydrate-builder to calculate block depth based on initial blocks offset

### DIFF
--- a/packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
+++ b/packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
@@ -1,0 +1,77 @@
+import {
+  PartialRehydrationDelegate,
+  test,
+  RenderTest,
+  suite,
+  content,
+  OPEN,
+  CLOSE,
+  replaceHTML,
+  qunitFixture,
+} from '..';
+import { castToSimple } from '@glimmer/util';
+
+export class PartialRehydrationTest extends RenderTest {
+  static suiteName = 'partial rehydration';
+  delegate!: PartialRehydrationDelegate;
+
+  @test
+  'can rehydrate from non starting blocks'() {
+    this.delegate.registerBasicComponent('RehydratingComponent', '{{@a}}{{@b}}{{@c}}');
+
+    this.delegate.registerBasicComponent(
+      'Root',
+      '<div id="placeholder"><RehydratingComponent @a={{@a}} @b={{@b}} @c={{@c}}/></div>'
+    );
+
+    const args = {
+      a: 'a',
+      b: 'b',
+      c: 'c',
+    };
+
+    const html = this.delegate.renderComponentServerSide('Root', args);
+    this.assert.equal(
+      html,
+      content([
+        OPEN,
+        OPEN,
+        '<div id="placeholder">',
+        OPEN,
+        OPEN,
+        'a',
+        CLOSE,
+        OPEN,
+        'b',
+        CLOSE,
+        OPEN,
+        'c',
+        CLOSE,
+        CLOSE,
+        '</div>',
+        CLOSE,
+        CLOSE,
+      ]),
+      'Expect server output to match'
+    );
+
+    replaceHTML(qunitFixture(), html);
+
+    this.element = castToSimple(document.getElementById('placeholder')!);
+    this.renderResult = this.delegate.renderComponentClientSide(
+      'RehydratingComponent',
+      args,
+      this.element
+    );
+    this.element = qunitFixture();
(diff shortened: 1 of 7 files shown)


packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
```diff
@@ -45,17 +46,38 @@ export class RehydrateBuilder extends NewElementBuilder implements ElementBuilde
...
-    assert(
-      node,
-      `Must have opening comment <!--${SERIALIZATION_FIRST_NODE_STRING}--> for rehydration.`
-    );
+    assert(node, 'Must have opening comment for rehydration.');
     this.candidate = node;
+    const startingBlockOffset = getBlockDepth(node as SimpleComment);
+    if (startingBlockOffset !== 0) {
+      // We are rehydrating from a partial tree and not the root component
+      // We need to add an extra block before the first block to rehydrate correctly
```
rwjblue, line 60: Just curious, why do we have to add an extra block for this case?

rwjblue (commented):
Looks good to me overall, had one question inline. 

Also, will ping others for review as well.

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
chadhietala: I may be completely off here, but if I remember correctly when we are invoking from `renderComponent` we actually [synthesis an invocation](https://github.com/glimmerjs/glimmer-vm/blob/5ec6a513d7c3ccba9c6b3e572a25162ea5a0a561/packages/%40glimmer/runtime/lib/render.ts#L55). This is going to push blocks onto the stack.

I'm not sure if the test suite is using this under the hood though.

chiragpat: That is precisely it, using renderComponent pushes an extra block onto the stack and we use renderComponent in both glimmerx and glimmer.js. The test suite uses renderMain for all the tests. There was one suite that used renderComponent but that suite doesn't seem to be getting executed anymore.

rwjblue: Gotcha, thank you for explaining!

packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
```diff
@@ -0,0 +1,77 @@
...
+  suite,
+  content,
+  OPEN,
+  CLOSE,
+  replaceHTML,
+  qunitFixture,
+} from '..';
+import { castToSimple } from '@glimmer/util';
+
+export class PartialRehydrationTest extends RenderTest {
```
chadhietala, line 14: Can we test multiple call sites in the document?

chadhietala (commented):
Overall I think this is likely the best we will be able to do given the current state of the rehydrating builder. That said this whole class likely needs to be revisited as I'm not sure what real benefit it has being a subclass of `NewElementBuilder`.

I would like to see some more testing of this specifically more cases like:

- Multiple rehydratable sections within a document
- Mismatch cases like in [`ChaosMonkeyRehydration`](https://github.com/glimmerjs/glimmer-vm/blob/5774aa1dfed852f8b8d06a0b81ee89592029ee6a/packages/%40glimmer/integration-tests/test/chaos-rehydration-test.ts#L24) suite

packages/@glimmer/runtime/lib/vm/rehydrate-builder.ts
pzuraq: We should work toward removing `renderMain` in general and refactoring this I think, but that's outside fo the scope for this PR

pzuraq (approved):
Looks good! I agree we may want to rethink the element builder setup here but that seems to be a larger refactor than is necessary at the moment

packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
chiragpat: Yes. Added a test for this.

chadhietala (approved):

PR #1180, opened 2020-10-23
[WIP] Add support for nextSibling for partial rehydration

chiragpat:
Putting up a draft PR for supporting nextSibling use cases for partial rehydration. I am not sure we want to do this before a refactoring of element builders but putting up a draft for future reference. 

**Why**
With partial rehydration we need to be able to let the rehydrate builder which particular component in the container is being rehydrated. This can be achieved by passing a nextSibling in the initial cursor to the renderComponent call

**How**
- Modify rehydrate builder to walk backwards from the nextSibling to find the first closing block
- Then walk backwards from there to find the first opening block at the same depth


NullVoxPopuli:
Is this still worth pursuing?

PR #1184, opened 2020-10-28
[REFACTOR] Upstream `setComponentTemplate`

pzuraq:
Upstreams

- `setComponentTemplate`
- `getComponentTemplate`
- `templateOnlyComponent`
- `isTemplateOnlyComponent`

In doing so, the tests needed to be refactored to use these APIs rather
than the resolution system that existed already. The resolution system
was also very different from the one in Ember, and was ultimately more
aligned with Glimmer.js v1.

This PR simplifies the test setup so that templates are no longer
registered on the registry at all. Instead, they are associated via
`setComponentTemplate`, then reified the first time the component is
looked up, similar to the way they are in Ember when using
`setComponentTemplate`. This helps to simplify the test infrastructure
overall.


rwjblue (approved):
LGTM

rwjblue:
Labeled as breaking (removes arguments from some public interfaces)

PR #1185, opened 2020-10-28
[REFACTOR] Upstreams Owner, simplifies template meta/ownership

pzuraq:
Upstreams the concept of the owner, and simplifies logic surrounding
templates. Currently templates receive a generic referrer, which the
embedding environment is supposed to define. The reality is that this
referrer boils down to:

1. `moduleName`
2. `owner`

Which are then threaded through the resolution system and used by the
embedding environment to resolve values. This system was going to be
extended for Module Unification, but with the pivot to strict mode its
no longer needed at all. Instead, scope will cover resolution in
general.

So, given that, this PR makes the interface concrete - users must
provide a `moduleName` for all template factories (this is also helpful
for  debugging purposes, so it's something we want to maintain into the
future) and they may optionally provide an Owner when generating a
template from a template factory. The Owner is _required_ for resolution
mode, but _not_ for strict mode, so we make it optional and then assert
whenever we resolve something in resolution mode.

Notes:

- Key interfaces and APIs have been made concrete, but have not been
  changed for backwards compat, including:
  - Template factories expose `__id` and `__meta`, as they would in Ember.
  - Templates expose `id` and `referrer`, matching the previous interface
  - `precompile` and `preprocess` can still receive a `meta` option. The
  interface has been changed to match the meta that is provided by
  `ember-cli-htmlbars` (`{ moduleName: string; }`). It is still passed to
  AST transforms, and in general should function similarly.


packages/@glimmer/integration-tests/test/precompile-test.ts
```diff
@@ -32,75 +24,56 @@ QUnit.module('templateFactory', {
...
 QUnit.test('meta is accessible from factory', (assert) => {
-  let factory = templateFactory(serializedTemplate);
-  assert.deepEqual(factory.meta, {
-    kind: 'unknown',
-    meta: {},
-    module: 'template/module/name',
-    name: 'default',
-  });
+  let factory = templateFactory(serializedTemplate) as TemplateFactoryWithIdAndMeta;
+  assert.deepEqual(factory.__meta, { moduleName: 'template/module/name' });
```
krisselden, line 53: why are you changing this to __meta?

packages/@glimmer/integration-tests/test/precompile-test.ts
```diff
@@ -32,75 +24,56 @@ QUnit.module('templateFactory', {
...
-        owner,
-      })
-    )
-  );
+  let template = unwrapTemplate(factory(owner)) as TemplateWithIdAndReferrer;
 
   assert.strictEqual(template.referrer.owner, owner, 'is owner');
   assert.deepEqual(
-    template.referrer as AnnotatedModuleLocator & { owner: unknown },
+    template.referrer,
```
krisselden, line 74: is this used? you made referrer owner in the lookup API

krisselden (commented):
Overall I think this is a good change but doesn't quite feel cohesive.

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -8,6 +8,9 @@ export interface TemplateIdFn {
 }
 
 export interface PrecompileOptions extends CompileOptions, PreprocessOptions {
+  meta?: {
```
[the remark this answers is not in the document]
pzuraq: Answered offline, yes it can be and if so, a default module name is used instead. In the future we'll likely want to switch to using a top level `moduleName` option, and deprecate this one (left a comment to this effect).

packages/@glimmer/integration-tests/test/precompile-test.ts
pzuraq: Template factories are wrapped by Ember an extra time, and expose it as `__meta` instead of `meta`.

packages/@glimmer/integration-tests/test/precompile-test.ts
pzuraq: It's unclear if this API is still used, but I've seen references to it in a few places in code that use `__meta`. I decided to keep it around as a backwards compatible API for the time being, and remove it in the near future. I added a TODO comment to track removing it.

PR #1192, opened 2020-11-10
[REFACTOR] Integrate DebugRenderTree

pzuraq:
Integrates the DebugRenderTree into Glimmer VM directly, so it doesn't
have to be managed by individual component managers.

Major changes:

1. `getDebugRenderTreeName` API added if managers need to override the
  name provided by `getDebugName`
2. `getCustomDebugRenderTree` API added, allows managers like `outlet`
  and `mount` to specify custom render trees via a list of render nodes.
3. TemplateOnly components can now integrate in the render tree without
  having to optionally have `createArgs` or `createInstance`
  capabilities.

Breaking changes:

* `env.debugRenderTree` has been made optional, it could potentially be undefined and will not assert if it is not
* `DebugRenderTree#setTemplate` has been removed

### DIFF
--- a/packages/@glimmer/interfaces/lib/components/component-manager.d.ts
+++ b/packages/@glimmer/interfaces/lib/components/component-manager.d.ts
@@ -5,13 +5,14 @@ import ComponentCapabilities from '../component-capabilities';
 import { ComponentDefinitionState, PreparedArguments, ComponentInstanceState } from '../components';
 import { Option, Destroyable } from '../core';
 import { Bounds } from '../dom/bounds';
-import { VMArguments } from '../runtime/arguments';
+import { CapturedArguments, VMArguments } from '../runtime/arguments';
 import { ElementOperations } from '../runtime/element';
 import { Environment } from '../runtime/environment';
 import { RuntimeResolver } from '../serialize';
-import { CompilableProgram, Template } from '../template';
+import { Template } from '../template';
 import { ProgramSymbolTable } from '../tier1/symbol-table';
 import { DynamicScope } from '../runtime/scope';
+import { RenderNode } from '../runtime/debug-render-tree';
 
 export interface ComponentManager<
   ComponentInstanceState = unknown,
@@ -21,6 +22,24 @@ export interface ComponentManager<
   getSelf(state: ComponentInstanceState): Reference;
   getDestroyable(state: ComponentInstanceState): Option<Destroyable>;
   getDebugName(state: ComponentDefinitionState): string;
+}
+
+interface CustomRenderNode extends RenderNode {
+  bucket: object;
+}
+
+export interface WithCustomDebugRenderTree<
+  ComponentInstanceState = unknown,
+  ComponentDefinitionState = unknown
+> extends ComponentManager<ComponentInstanceState, ComponentDefinitionState> {
+  // APIs for hooking into the debug render tree, used by components that
+  // represent multiple logical components. Specifically, {{mount}} and {{outlet}}
+  getDebugCustomRenderTree(
+    definition: ComponentDefinitionState,
+    state: ComponentInstanceState,
+    args: CapturedArguments,
+    template?: Template
+  ): CustomRenderNode[];
 }
 
 export interface WithPrepareArgs<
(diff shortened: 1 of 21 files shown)


packages/@glimmer/interfaces/lib/components/component-manager.d.ts
```diff
@@ -21,6 +22,25 @@ export interface ComponentManager<
   getSelf(state: ComponentInstanceState): Reference;
   getDestroyable(state: ComponentInstanceState): Option<Destroyable>;
   getDebugName(state: ComponentDefinitionState): string;
+  getDebugRenderTreeName?(state: ComponentDefinitionState): string;
```
rwjblue, line 25: We discussed this a bit separately, but I think we should remove this extra hook and handle the "while rendering" thing completely separately (e.g. where we have the `try`/`finally` during append/update VM's).

packages/@glimmer/interfaces/lib/components/component-manager.d.ts
```diff
@@ -21,6 +22,25 @@ export interface ComponentManager<
...
+interface CustomRenderNode extends RenderNode {
+  bucket: object;
+}
+
+export interface WithCustomDebugRenderTree<
+  ComponentInstanceState = unknown,
+  ComponentDefinitionState = unknown
+> extends ComponentManager<ComponentInstanceState, ComponentDefinitionState> {
+  // APIs for hooking into the debug render tree, used by components that
+  // represent multiple logical components. Specifically, {{mount}} and {{outlet}}
```
rwjblue, line 37: What about `{{component`? (no need to update the message for that, but I'm curious if that falls into the same bucket)

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/components.ts
```diff
@@ -221,6 +222,10 @@ export function InvokeStaticComponent({
     // to blocks (they aren't accessible from userland anyways), so we push an
     // empty array instead of the actual block names.
     out.push(op(Op.PushArgs, strArray(names), strArray(EMPTY_ARRAY), flags));
+
+    // And push an extra pop to pop the args off before we begin setting
```
rwjblue, line 226: > push an extra pop to pop

Is this phrasing right? It reads a bit confusingly

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -699,14 +766,32 @@ APPEND_OPCODES.add(Op.InvokeComponentLayout, (vm, { op1: _state }) => {
...
+  let { manager, state, capabilities } = instance;
   let bounds = vm.elements().popBlock();
 
-  let mgr = check(manager, CheckInterface({ didRenderLayout: CheckFunction }));
+  if (vm.env.debugRenderTree !== undefined) {
+    if (hasCustomDebugRenderTreeLifecycle(manager)) {
+      let nodes = manager.getDebugCustomRenderTree(instance.definition.state, state, EMPTY_ARGS);
 
-  mgr.didRenderLayout(state, bounds);
+      nodes.reverse().forEach((node) => {
```
rwjblue, line 777: Why in reverse? Also, do we really want to mutate the array directly?

rwjblue (commented):
Overall seems good, a few high level questions though:

* Can you add tests for this using the highest level APIs possible? (rendering a tree of things, asserting on the deep state of the render tree)
* Can you make a draft PR to Ember showing the integration side (so we can get a feel for how it looks over there)?

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
pzuraq: It's in reverse because we want to pop off arrays from the opposite order that they were pushed on, and reversing is fine as we return a new array every time

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/components.ts
pzuraq: Can rephrase, "push an extra pop operation to remove the args"

packages/@glimmer/interfaces/lib/components/component-manager.d.ts
pzuraq: It does not currently, but it could in the future. It couldn't have done this previously since `{{component}}` is implemented in the VM, so it didn't have access

pzuraq:
Here is the integration PR in Ember: https://github.com/emberjs/ember.js/pull/19261

packages/@glimmer/validator/lib/debug.ts
```diff
@@ -121,13 +121,18 @@ if (DEBUG) {
...
     try {
-      fn();
+      let value = fn();
+      didError = false;
+      return value;
     } finally {
-      endTrackingTransaction!();
+      if (didError !== true) {
+        endTrackingTransaction!();
+      }
```
rwjblue, line 135: This seems a tad surprising, since `runInTrackingTransaction` always calls `beginTrackingTransaction` why wouldn't it always call `endTrackingTransaction`?

Is it because we already call `resetTracking` (basically in our callsite)?

pzuraq: Exactly that. With the current structure of things, there is a case where this can happen unfortunately. I think we can work on restructuring it in the future to make it a bit more bulletproof.

rwjblue (approved):

PR #1197, opened 2020-11-12
[REFACTOR] Upstreams basic manager infrastructure and types

pzuraq:
Upstreams the basic manager infrastructure and types to Ember. This will
be followed by upstreaming the respective managers for each type, along
with some of the base managers.

Breaking Changes:

- Renamed `ComponentManager` -> `InternalCompnoentManager`
- Renamed `ModifierManager` -> `InternalModifierManager`
- Upstreamed manager types from Ember
- Upstreamed `setManager` and `getManager` functions from Ember
- Upstreamed `buildCapabilities` from Ember

Ember integration: https://github.com/emberjs/ember.js/pull/19267


packages/@glimmer/benchmark-env/src/benchmark/create-registry.ts
```diff
@@ -51,7 +51,7 @@ export interface Registry {
     name: string,
     template: SerializedTemplateWithLazyBlock,
     component: T,
-    manager: ComponentManager<unknown, T>
+    manager: InternalComponentManager<unknown, T>
```
rwjblue, line 54: Curious, why can't this use the "public" component manager?

packages/@glimmer/interfaces/lib/managers/component.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  '3.4': {
+    asyncLifecycleCallbacks?: boolean;
+    destructor?: boolean;
+  };
+
+  '3.13': {
+    asyncLifecycleCallbacks?: boolean;
+    destructor?: boolean;
+    updateHook?: boolean;
+  };
```
rwjblue, line 14: I definitely have no objection here (and this isn't a blocking concern), but we **need** to figure out how we are going to pick these numbers in the future.

packages/@glimmer/interfaces/lib/managers/internal/component.d.ts
```diff
@@ -1,24 +1,118 @@
...
+ *
+ * Certain features in the VM come with some overhead, so the compiler and
+ * runtime use this information to skip unnecessary work for component types
+ * that don't need it.
+ *
+ * For example, a component that is template-only (i.e., it does not have an
+ * associated JavaScript class to instantiate) can skip invoking component
+ * manager hooks related to lifecycle events by setting the `elementHook`
+ * capability to `false`.
+ */
```
rwjblue, line 33: Nice ❤️

packages/@glimmer/runtime/lib/component/manager.ts
```diff
@@ -15,8 +15,8 @@ import { MINIMAL_CAPABILITIES } from './interfaces';
 import { Reference, UNDEFINED_REFERENCE } from '@glimmer/reference';
 
 // TODO: This manager appears to be unused
```
rwjblue, line 17: 🤷‍♂️ remove it?

packages/@glimmer/runtime/lib/managers/index.ts
```diff
@@ -0,0 +1,236 @@
...
+  if (factory !== undefined) {
+    let manager = getManagerInstanceForOwner(owner, factory);
+
+    if (
+      DEBUG &&
+      !isInternalModifierManager(manager) &&
+      !FROM_CAPABILITIES!.has(manager.capabilities)
+    ) {
+      throw new Error(
+        `Custom modifier managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.13' | '3.22')\` (imported via \`import { capabilities } from '@ember/modifier';\`). Received: \`${JSON.stringify(
```
rwjblue, line 150: Assuming that Glimmer.js will also leverage this infrastructure (I assume it has to), we need to allow this suggested string to be "host dependent". Can we have this configured by the host instead?

Obviously, we should also work to make the value the same (so that both Ember and Glimmer.js apps import from the same place), but we'll need an RFC for that...

rwjblue: IMHO, making the error message _less_ useful is not a valid choice here, I think if we have to land this as `@ember/modifier` in the string for now that is "fine" (by some definition 😝 ) but we should make an issue (and an inline comment pointing to that issue) to come back and clean this up (either by allowing the host to customize the value, or by doing the RFC in Ember land to allow the import to be from an `@glimmer` namespace).

packages/@glimmer/runtime/lib/managers/index.ts
```diff
@@ -0,0 +1,236 @@
...
+  definition: object
+): HelperManager<unknown> | Helper | undefined {
+  const factory = getManager(HELPER_MANAGERS, definition);
+
+  if (factory !== undefined) {
+    let manager = getManagerInstanceForOwner(owner, factory);
+
+    if (DEBUG && !isInternalHelper(manager) && !FROM_CAPABILITIES!.has(manager.capabilities)) {
+      throw new Error(
+        `Custom helper managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.23')\` (imported via \`import { capabilities } from '@ember/helper';\`). Received: \`${JSON.stringify(
```
rwjblue, line 180: Same as above RE: the error message

packages/@glimmer/runtime/lib/managers/index.ts
```diff
@@ -0,0 +1,236 @@
...
+  if (factory !== undefined) {
+    let manager = getManagerInstanceForOwner(owner, factory);
+
+    if (
+      DEBUG &&
+      !isInternalComponentManager(manager) &&
+      !FROM_CAPABILITIES!.has(manager.capabilities)
+    ) {
+      throw new Error(
+        `Custom component managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.4' | '3.13')\` (imported via \`import { capabilities } from '@ember/component';\`). Received: \`${JSON.stringify(
```
rwjblue, line 217: Same RE: error message

rwjblue (changes requested):
Some open questions:

* We should migrate things that _can_ use the public managers to using them (e.g. benchmark, integration-tests for things like "glimmerishcomponent", etc). Can you make an issue with a TODO so that we don't loose track?
* Can you link to the Ember integration PR in the description?
* Have you reviewed the work to integration in Glimmer.js? Is it roughly the same / very similar to the work for Ember integration?
* Can you add tests to confirm that `buildCapabilities` is required for each of the types (we have had a number of regressions on this over in the Ember side, and it would be good to ensure this is under test)? This would mean both that `.capabilities` must be on your manager instance and that it is the result of calling `buildCapabilities`..

Future looking questions:

* We have to figure out where the API versions come from now that the code itself is not colocated in Ember. It is quite difficult to know what version of Ember a given WIP in glimmer-vm will target, so choosing the version for new capabilities versions might be pretty hard.
* Now that the managers (and the user-facing manager APIs) are moved into glimmer-vm, do we still want the delegate system? Naively (to me) it seems like that is just wasted effort (aliasing hooks from one to the other, instantiating extra objects, etc).

packages/@glimmer/interfaces/lib/managers/component.d.ts
pzuraq: Agreed, wholeheartedly. I think honestly just versioning each API independently is probably a decent solution, as long as we document them in the API docs. It means users will have to do a bit of extra looking-up to figure out what capabilities are available (and supported), but that's a given with this APIs anyways.

pzuraq:
> Can you add tests to confirm that buildCapabilities is required for each of the types (we have had a number of regressions on this over in the Ember side, and it would be good to ensure this is under test)? This would mean both that .capabilities must be on your manager instance and that it is the result of calling buildCapabilities..

We do have tests for these: https://github.com/glimmerjs/glimmer-vm/pull/1197/files#diff-1e72e3bda8309fc230d43679cb9d9112b08cc1b676f6a7899f1907b7984f094eR224

packages/@glimmer/runtime/lib/component/manager.ts
pzuraq: 👍

pzuraq: Actually it is now used by the benchmark-env it seems, I'll update the comment, we should remove it in the next set of refactors

packages/@glimmer/benchmark-env/src/benchmark/create-registry.ts
rwjblue: Sounds good, can you make an issue to track that)/

rwjblue:
> > Can you add tests to confirm that buildCapabilities is required for each of the types (we have had a number of regressions on this over in the Ember side, and it would be good to ensure this is under test)? This would mean both that .capabilities must be on your manager instance and that it is the result of calling buildCapabilities..
> 
> We do have tests for these: #1197 (files)

Hmm, that only confirms that you have a capabilities property, not that it is the result of calling the internal branded thing.

pzuraq:
Good point! Updated and added another set of tests specifically for a non-built value, not a missing value.

packages/@glimmer/benchmark-env/src/benchmark/create-registry.ts
pzuraq: Yup! Made it at: https://github.com/glimmerjs/glimmer-vm/issues/1201

pzuraq:
Re: Updating on the Glimmer side, the updates are relatively minimal. This is the PR for it: https://github.com/glimmerjs/glimmer.js/pull/317

rwjblue (approved):

PR #1198, opened 2020-11-12
bug: Failing test for partial rehydrated yielded blocks

chadhietala:
This is a failing test that shows that `{{yield}}`s do not rehydrate properly when trying to rehydrate a portion of a document.

Output is:

```html
<!--%+b:0%--><!--%+b:1%--><div id="placeholder">Bar </div><!--%-b:1%--><!--%-b:0%-->
```

When it should be

```html
<!--%+b:0%--><!--%+b:1%--><div id="placeholder">Bar Foo</div><!--%-b:1%--><!--%-b:0%-->
```

/cc @chiragpat

### DIFF
--- a/packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
+++ b/packages/@glimmer/integration-tests/test/partial-rehydration-test.ts
@@ -73,6 +73,53 @@ export class PartialRehydrationTest extends RenderTest {
   }
 
   @test
+  'can rehydrate yielded blocks'() {
+    this.delegate.registerTemplateOnlyComponent('RehydratingComponent', 'Bar {{yield}}');
+
+    this.delegate.registerTemplateOnlyComponent(
+      'Root',
+      '<div id="placeholder"><RehydratingComponent>Foo</RehydratingComponent></div>'
+    );
+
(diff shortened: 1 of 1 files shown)


chadhietala:
After thinking about this more this is actually the correct behavior. The top-level component cannot be in the open state. 

@pzuraq, for strict mode my understanding is that `{{component}}` will only create a definition or a result, there is no lookup. Below are some examples to make sure I understand correctly.

**Passing Bar with curried arguments**
```hbs
import { component } from '...somewhere...';

<!-- Creates a curried component that can be invoked -->
<Foo @Component={{component Bar thing=1}} />
```

In layout

```hbs
<@Component @moreStuff="Baz" />
```

**Passing the evaluated component**
```hbs
import { component } from '...somewhere...';

<!-- Creates a curried component that can be invoked -->
<Foo @Component={{(component Bar thing=1)}} />
```

In layout

```hbs
<@Component />
```

If this is true, how does one pass a component with args through `renderComponent`? Can you use `component` in the same way e.g.

```js
await renderComponent(Example, {
  args: {
     Component: component(Bar, { thing: 1 })
  }
})
```

PR #1199, opened 2020-11-13
bugfix: Fix debug logging

chadhietala:
Seems like the concept of `getOther` is gone


locks:
Oops, I didn't notice the work is done in master already before trying to resolve the conflict.

PR #1203, opened 2020-11-18
Update syntax errors to not be a custom subclass

pzuraq:
This updates syntax errors to throw normal errors, with an additional property added. This allows them to work with browser tooling the right way.

### DIFF
--- a/packages/@glimmer/syntax/lib/syntax-error.ts
+++ b/packages/@glimmer/syntax/lib/syntax-error.ts
@@ -2,36 +2,23 @@ import { SourceSpan } from './source/span';
 
 export interface GlimmerSyntaxError extends Error {
   location: SourceSpan | null;
-  constructor: SyntaxErrorConstructor;
+  code: string | null;
 }
 
-export interface SyntaxErrorConstructor {
-  new (message: string, location: SourceSpan | null): GlimmerSyntaxError;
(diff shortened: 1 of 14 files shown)


packages/@glimmer/syntax/lib/syntax-error.ts
```diff
@@ -2,36 +2,20 @@ import { SourceSpan } from './source/span';
...
 
-    let { module, loc } = location;
-    let { line, column } = loc.start;
+  let error = new Error(
+    `Syntax Error: ${message}: ${quotedCode}(error occurred in '${module}' @ line ${line} : column ${column})`
+  ) as GlimmerSyntaxError;
 
-    let code = location.asString();
-    let quotedCode = code ? `\n\n|\n|  ${code.split('\n').join('\n|  ')}\n|\n\n` : '';
+  error.location = location;
```
krisselden, line 18: You should set error.name = "SyntaxError" and don't include `Syntax Error:` in the message

krisselden (changes requested):

packages/@glimmer/syntax/lib/syntax-error.ts
rwjblue, line 18: Can you also add `error.code`?

rwjblue (approved):

Issue #1207, opened 2020-11-19
Missing <output> in attr_overrides


KrishnaRPatel:
Line:
https://github.com/glimmerjs/glimmer-vm/blob/25fb030869c6e047e3c7b963399b98c8ee6bd7c4/packages/%40glimmer/runtime/lib/dom/props.ts#L48

Bug:
https://github.com/emberjs/ember.js/issues/19272

Setting an `<output>`'s `form` with a non-ancestor form's `id` should be allowed, but is set as `readonly` in typescript (as it is in the whatwg idl). `<output>`'s form attribute is similar to the others (https://github.com/whatwg/html/issues/6165) so we should add the `output` tag as an additional override here.

rwjblue:
Fixed by #1209

PR #1209, opened 2020-11-19
[BUGFIX] Add output tag to attributes override

KrishnaRPatel:
Added `<output>` to `ATTR_OVERRIDES` as [MDN details how the `form` attribute can be set on `<output> tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output).

Failing test before fix:
image

Should fix this issue:
https://github.com/glimmerjs/glimmer-vm/issues/1207


lifeart (approved):

rwjblue:
Thank you @KrishnaRPatel!

PR #1217, opened 2020-11-25
Revert "[BREAKING] Update benchmark-env API"

pzuraq:
Reverts glimmerjs/glimmer-vm#1216

Reverting to give time for more discussion, cc @krisselden


pzuraq:
Failed benchmark is expected, see #1216

PR #1220, opened 2020-12-02
[BUGFIX] Absorb breakage in benchmark-env instead of benchmark

pzuraq:
Some previous PRs made changes to the benchmark, but this made the benchmark
itself incompatible with previous versions of Glimmer. This is something that
is valuable to be able to test, so for now are going to keep absorbing all
breakage in the benchmark-env.


pzuraq:
Krausest will fail for this change, since the previous commit depended on `setComponentTemplate` being called within the benchmark and not the benchmark-env. I've tested it locally, and the experiment does serve succesfully.

PR #1224, opened 2020-12-05
[BUGFIX] Make the custom tag system non-enumerable

pzuraq:
Currently, the custom tag system relies on an enumerable "symbol"-like
key, which is then placed on the objects that need to use it. This is an
expensive system, since it means we need to do a brand check to see if
the function exists, which means crawling the prototype chain every time
we look up a tag for a property on an object.

This PR instead places the functions in a WeakMap. This means that in
general, we can check to see if the WeakMap has the value, and if so
then we can use the function directly. WeakMap lookup is generally
faster, as it means we don't have to walk the entire prototype chain of
a megamorphic object, so this should generally be more performant.

It does mean that the functions in question need to be bound to the
instance, which could impact performance, but the functions should still
be inline-able when called (since they are called independent of the
object), and given that usage of tag lookup with non-proxies likely
dominates over usage with proxies, it should matter less than the
savings from using a WeakMap.

Finally, this means that CUSTOM_TAG_FOR is not exposed at all, so there
cannot be any observable side-effects such as the ones causing the bug
in: https://github.com/hjdivad/ember-m3/issues/957

### Unrelated Changes

In addition, I did a quick refactor of the proxy handlers to extract
them as classes. This is important for performance (POJOs with methods
on the instance do not inline, even as proxy handlers), I just never got
around to it before.

### DIFF
--- a/packages/@glimmer/manager/lib/util/args-proxy.ts
+++ b/packages/@glimmer/manager/lib/util/args-proxy.ts
@@ -6,10 +6,18 @@ import {
   CapturedPositionalArguments,
 } from '@glimmer/interfaces';
 import { Reference, valueForRef } from '@glimmer/reference';
-import { HAS_NATIVE_PROXY, enumerableSymbol } from '@glimmer/util';
+import { HAS_NATIVE_PROXY } from '@glimmer/util';
 import { Tag, track } from '@glimmer/validator';
 
-export const CUSTOM_TAG_FOR = enumerableSymbol('CUSTOM_TAG_FOR');
+const CUSTOM_TAG_FOR = new WeakMap<object, (obj: object, key: string) => Tag>();
+
+export function getCustomTagFor(obj: object): ((obj: object, key: string) => Tag) | undefined {
+  return CUSTOM_TAG_FOR.get(obj);
+}
+
+export function setCustomTagFor(obj: object, customTagFn: (obj: object, key: string) => Tag) {
+  CUSTOM_TAG_FOR.set(obj, customTagFn);
+}
 
 function convertToInt(prop: number | string | symbol): number | null {
   if (typeof prop === 'symbol') return null;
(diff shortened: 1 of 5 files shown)


rwjblue:
Were you able to confirm that this fixes https://github.com/hjdivad/ember-m3/issues/957?

packages/@glimmer/util/lib/platform-utils.ts
```diff
@@ -39,7 +39,7 @@ export type Lit = string | number | boolean | undefined | null | void | {};
 
 export const tuple = <T extends Lit[]>(...args: T) => args;
 
-export function enumerableSymbol(key: string): any {
+function enumerableSymbol(key: string): any {
```
rwjblue, line 42: We should still probably export so that we can drop the duplicate copy over in Ember

packages/@glimmer/manager/lib/util/args-proxy.ts
```diff
@@ -6,10 +6,18 @@ import {
...
 } from '@glimmer/interfaces';
 import { Reference, valueForRef } from '@glimmer/reference';
-import { HAS_NATIVE_PROXY, enumerableSymbol } from '@glimmer/util';
+import { HAS_NATIVE_PROXY } from '@glimmer/util';
 import { Tag, track } from '@glimmer/validator';
 
-export const CUSTOM_TAG_FOR = enumerableSymbol('CUSTOM_TAG_FOR');
+const CUSTOM_TAG_FOR = new WeakMap<object, (key: string) => Tag>();
+
+export function getCustomTagFor(obj: object): ((key: string) => Tag) | undefined {
```
rwjblue, line 14: Who uses this?

rwjblue: We can avoid requiring the function to be bound eagerly by passing the object as the first arg. Not sure it matters too much though.

pzuraq:
Confirmed that it does fix the bug!

packages/@glimmer/util/lib/platform-utils.ts
pzuraq: If that's the case I will also move `intern` upstream, cause I don't want us to accidentally de-opt without using that for enumerable symbols

packages/@glimmer/manager/lib/util/args-proxy.ts
pzuraq: 👍  good suggestion, will do!

packages/@glimmer/util/lib/platform-utils.ts
rwjblue: Ya, that makes sense to me.

rwjblue (approved):

PR #1225, opened 2020-12-09
[FEAT] Implements Strict Mode

pzuraq:
Implements resolution in strict mode, and adds a number of tests for
strict mode functionality, including:

- Testing that components, helpers, and modifiers work when imported in
  strict
- Testing that keywords can be shadowed by locals
- Testing that the VM does not fallback on resolution or `this` lookup
  if a value is not in scope
- Testing that using strings in `{{component}}` does not work
- Testing that constant values work in templates
- Using `{{partial}}` does not work in strict teplates

And more.

## Breaking Changes

- CompileTimeResolver must now implement `resolveBuiltInHelper` and `resolveBuiltInModifier`


rwjblue:
@pzuraq can you add a test to confirm that `{{partial` does not function (in strict mode)?

packages/@glimmer/benchmark-env/src/benchmark/render-benchmark.ts
```diff
@@ -18,7 +17,7 @@ export default async function renderBenchmark(
   context: CompileTimeCompilationContext,
   runtimeResolver: RuntimeResolver,
   component: ResolvedComponentDefinition,
-  root: Dict,
```
rwjblue, line 21: Is this just not needed?

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -64,17 +64,13 @@ const defaultOptions: PrecompileOptions = {
  * @param {string} string a Glimmer template string
  * @return {string} a template javascript string
  */
-export function precompileJSON(
-  source: string,
-  options?: PrecompileOptions
-): SerializedTemplateBlock;
```
rwjblue, line 70: wat, why did we overload this?

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -116,10 +112,20 @@ export function precompile(
...
   };
 
   // JSON is javascript
-  return JSON.stringify(templateJSONObject);
+  let stringified = JSON.stringify(templateJSONObject);
+
+  if (usedLocals.length > 0) {
+    let scopeFn = `()=>[${usedLocals.join(',')}]`;
+
+    stringified = stringified.replace('"scope":null', `"scope":${scopeFn}`);
```
rwjblue, line 125: This seems somewhat brittle to me. Is there a different / better way? One would be to add `scope: 'SOME-GUID-VALUE'` and then we know precisely what to replace...

packages/@glimmer/compiler/lib/compiler.ts
```diff
@@ -116,10 +112,20 @@ export function precompile(
...
     moduleName: moduleName ?? '(unknown template module)',
+    scope: null,
+    isStrictMode: options.strictMode ?? false,
   };
 
   // JSON is javascript
-  return JSON.stringify(templateJSONObject);
+  let stringified = JSON.stringify(templateJSONObject);
+
+  if (usedLocals.length > 0) {
```
rwjblue, line 122: Should this be something like:

```suggestion
  if (options.strictMode && usedLocals.length > 0) {
```

?

pzuraq:
@rwjblue ah yes, that is also here, forgot to mention it in the summary: https://github.com/glimmerjs/glimmer-vm/pull/1225/files#diff-6591e40dde2e8325baf58e9e99ba1e8758a1c34a6a9aab9744ffc53d815cf679R30

packages/@glimmer/benchmark-env/src/benchmark/render-benchmark.ts
pzuraq: Correct, I refactored `renderComponent` to wrap the args internally, because that code was getting duplicated everywhere (I think 5 places now, including Glimmer.js). So now this wrapping happens internally, all you do is pass an object containing arguments to `renderComponent`.

packages/@glimmer/compiler/lib/compiler.ts
pzuraq: 🤷  I think it was a mistake

packages/@glimmer/compiler/lib/compiler.ts
pzuraq: yeah, I can do that, sounds good 👍

packages/@glimmer/compiler/lib/compiler.ts
pzuraq: yeah, that makes sense. In theory `usedLocals` should only exist if we are in strict mode, but it's better to be explicit I think.

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -0,0 +1,723 @@
...
+  //   const Foo = defineComponent({}, '{{@asHelper 123}} {{@asValue}} {{@asValueWithArg}}');
+  //   const Bar = defineComponent(
+  //     { plusOne, Foo },
+  //     '<Foo @asHelper={{plusOne}} @asValue={{(plusOne)}} @asValueWithArg={{plusOne 456}}/>'
+  //   );
+
+  //   this.renderComponent(Bar);
+  //   this.assertHTML('124 1 457');
+  //   this.assertStableRerender();
+  // }
```
rwjblue, line 27: Why is this commented out?

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -0,0 +1,723 @@
...
+  //     '<div {{this.foo}}></div>',
+  //     class extends GlimmerishComponent {
+  //       foo = foo;
+  //     }
+  //   );
+
+  //   this.renderComponent(Bar);
+  //   this.assertHTML('<div>Hello, world!</div>');
+  //   this.assertStableRerender();
+  // }
```
rwjblue, line 597: Why are these commented?

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -0,0 +1,723 @@
...
+    this.assertStableRerender();
+  }
+
+  // @test
+  // 'Can use a nested in scope value as dynamic helper'() {
+  // }
+
+  // @test
+  // 'Can use a nested in scope value as dynamic modifier'() {
+  // }
```
rwjblue, line 616: ✂️ 🤔

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -0,0 +1,723 @@
...
+  //     '<div {{this.foo}}></div>',
+  //     class extends GlimmerishComponent {
+  //       foo = foo;
+  //     }
+  //   );
+
+  //   this.assert.throws(() => {
+  //     this.renderComponent(Bar);
+  //   }, /aoeu/);
+  // }
```
rwjblue, line 718: 🤔

packages/@glimmer/interfaces/lib/serialize.d.ts
```diff
@@ -90,6 +90,11 @@ export interface CompileTimeResolver<O extends Owner = Owner> {
   lookupModifier(name: string, owner: O): Option<ModifierDefinitionState>;
   lookupComponent(name: string, owner: O): Option<ResolvedComponentDefinition>;
   lookupPartial(name: string, owner: O): Option<PartialDefinition>;
+
+  // TODO: These are used to lookup keywords that are implemented as helpers/modifiers.
+  // We should try to figure out a cleaner way to do this.
+  lookupBuiltInHelper(name: string, owner: O): Option<HelperDefinitionState>;
+  lookupBuiltInModifier(name: string, owner: O): Option<ModifierDefinitionState>;
```
rwjblue, line 97: Doesn't this addition mean that this is a (very slight) breaking change?

packages/@glimmer/integration-tests/test/strict-mode-test.ts
pzuraq: These tests are tests that I've written and should be included in the near future (e.g. next PR or two) but that I didn't want to include in this PR. I can keep them locally instead, wasn't sure which would be better to do. I _think_ there's a `@skip` potentially also, but tbh I think comments are basically equivalent to that.

packages/@glimmer/interfaces/lib/serialize.d.ts
pzuraq: Ah, good point, yes. I'll add that to the list of breaking changes.

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/resolution.ts
```diff
@@ -210,22 +314,91 @@ export function resolveOptionalComponentOrHelper(
...
+  constants: ResolutionTimeConstants,
+  type: string
+): number {
+  let { upvars, owner } = assertResolverInvariants(meta);
+
+  let name = upvars[expr[1]];
+  let helper = resolver.lookupBuiltInHelper(name, owner);
+
+  if (DEBUG && helper === null) {
+    // Keyword helper did not
```
rwjblue, line 395: did not......!?!?!?!?!

packages/@glimmer/integration-tests/test/strict-mode-test.ts
rwjblue: Ya, I'd prefer to break these out or @skip, but I don't feel super strongly.

pzuraq: converted to use `skip: true`, and made that work

rwjblue (approved):

PR #1226, opened 2020-12-09
[FEAT] Enable sub-expressions in mustache position

pzuraq:
Allows users to use sub-expressions directly within mustaches, either in
append or in argument position. This is the bare-minimum needed for
strict-mode, and does not yet support using sub-expressions in call
position recursively. For that, we need to implement dynamic helpers
(coming up in the next PR).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -867,10 +867,10 @@
   resolved "https://registry.yarnpkg.com/@glimmer/env/-/env-0.1.7.tgz#fd2d2b55a9029c6b37a6c935e8c8871ae70dfa07"
   integrity sha1-/S0rVakCnGs3psk16MiHGucN+gc=
 
-"@handlebars/parser@^1.1.0":
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/@handlebars/parser/-/parser-1.1.0.tgz#d6dbc7574774b238114582410e8fee0dc3532bdf"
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -872,6 +872,11 @@
   resolved "https://registry.yarnpkg.com/@handlebars/parser/-/parser-1.1.0.tgz#d6dbc7574774b238114582410e8fee0dc3532bdf"
   integrity sha512-rR7tJoSwJ2eooOpYGxGGW95sLq6GXUaS1UtWvN7pei6n2/okYvCGld9vsUTvkl2migxbkszsycwtMf/GEc1k1A==
 
+"@handlebars/parser@^2.0.0":
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/@handlebars/parser/-/parser-2.0.0.tgz#5e8b7298f31ff8f7b260e6b7363c7e9ceed7d9c5"
+  integrity sha512-EP9uEDZv/L5Qh9IWuMUGJRfwhXJ4h1dqKTT4/3+tY0eu7sPis7xh23j61SYUnNF4vqCQvvUXpDo9Bh/+q1zASA==
```
rwjblue, line 878: Do we still have the old version ? I would have expected to add this version and remove the old one...

rwjblue (approved):

yarn.lock
pzuraq: ah, it was being used in a sub package, fixed!

PR #1232, opened 2020-12-11
[REFACTOR] Make Owner VM state

pzuraq:
This PR moves the owner directly onto the VM, where it is tracked as
part of the current scope. This allows us to associate the owner with a
template block, as we do currently, but without having to associate it
through the various opcodes. This is both conceptually simpler, and
also more performant - in the future we can rely more on the stdlib to
deduplicate common code segments.

## Breaking Changes

- `create` signature of component managers has changed, owner is now the
  first parameter.
- New component capability, `HasSubOwner`, which allows components to
push a new owner onto the owner stack.

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
@@ -28,12 +28,12 @@ import {
   WithDynamicTagName,
   WithElementHook,
   WithUpdateHook,
-  Owner,
   CapturedArguments,
   CompilableProgram,
   ComponentInstance,
   ModifierInstance,
   ComponentInstanceWithCreate,
+  Owner,
 } from '@glimmer/interfaces';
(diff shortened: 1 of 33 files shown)


packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -716,9 +723,29 @@ APPEND_OPCODES.add(Op.PopulateLayout, (vm, { op1: _state }) => {
...
+  } else {
+    // Check the temp register to see if an owner was resolved from currying
+    owner = vm.fetchValue<Owner | null>($t1);
+
+    if (owner === null) {
+      // If an owner wasn't found, default to using the current owner
+      owner = vm.getOwner();
+    } else {
+      vm.loadValue($t1, null); // Clear the temp register
+    }
```
rwjblue, line 745: Can you add a small example above each of these branches to indicate what thing the user would have done to end up in each case?

PR #1235, opened 2020-12-15
[FEATURE] Adds dynamic curly components and helpers

pzuraq:
Adds dynamic curly component and helper invocation, as well as the
ability to curry components with the `(helper)` keyword.

### DIFF
--- a/packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
+++ b/packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
@@ -251,4 +251,61 @@ export const APPEND_KEYWORDS = keywords('Append')
           })
       );
     },
+  })
+  .kw('helper', {
+    assert(
+      node: ASTv2.AppendContent
+    ): Result<{
+      args: ASTv2.Args;
+    }> {
+      let { args } = node;
+
+      let definition = args.nth(0);
+
+      if (definition === null) {
+        return Err(
+          generateSyntaxError(
+            `{{helper}} requires a helper definition as its first positional parameter, did not receive any parameters.`,
+            args.loc
+          )
+        );
+      }
+
+      if (definition.type === 'Literal') {
+        return Err(
+          generateSyntaxError(
+            '{{helper}} cannot resolve string values, you must pass a helper definition directly',
+            node.loc
+          )
+        );
+      }
+
+      args = new ASTv2.Args({
+        positional: new ASTv2.PositionalArguments({
+          exprs: args.positional.exprs.slice(1),
+          loc: args.positional.loc,
+        }),
+        named: args.named,
+        loc: args.loc,
+      });
+
+      return Ok({ definition, args });
+    },
+
+    translate(
+      { node, state }: { node: ASTv2.AppendContent; state: NormalizationState },
+      { definition, args }: { definition: ASTv2.ExpressionNode; args: ASTv2.Args }
+    ): Result<mir.AppendTextNode> {
+      let definitionResult = VISIT_EXPRS.visit(definition, state);
+      let argsResult = VISIT_EXPRS.Args(args, state);
+
+      return Result.all(definitionResult, argsResult).mapOk(([definition, args]) => {
+        let text = new mir.CallExpression({ callee: definition, args, loc: node.loc });
+
+        return new mir.AppendTextNode({
+          loc: node.loc,
+          text,
+        });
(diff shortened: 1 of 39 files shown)


packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
```diff
@@ -251,4 +251,62 @@ export const APPEND_KEYWORDS = keywords('Append')
...
+      if (definition === null) {
+        return Err(
+          generateSyntaxError(
+            `{{helper}} requires a helper definition as its first positional parameter, did not receive any parameters.`,
+            args.loc
+          )
+        );
+      }
+
+      if (state.isStrict && definition.type === 'Literal') {
```
rwjblue, line 275: As discussed, should be

```suggestion
      if (definition.type === 'Literal') {
```

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -646,6 +737,19 @@ class DynamicStrictModeTest extends RenderTest {
...
+    const Bar = defineComponent(
+      {},
+      '<div {{this.foo 123}}></div>',
+      class extends GlimmerishComponent {}
+    );
+
+    this.renderComponent(Bar);
+    this.assertHTML('');
+    this.assertStableRerender();
+  }
```
rwjblue, line 751: This is just future work, right?

packages/@glimmer/vm/lib/opcodes.toml
```diff
@@ -187,6 +196,18 @@ operand-stack = [
   ["unknown"]
 ]
 
+[syscall.rconstrefload]
+
+format = ["ConstantReference", "constant:unknown"]
+operation = """
+  Push an reference constant onto the stack that is not
```
rwjblue, line 203:
```suggestion
  Push a reference constant onto the stack that is not
```

packages/@glimmer/integration-tests/test/strict-mode-test.ts
pzuraq: Yes, that's why the test was skipped, was just easier to write it out now while I had the context

packages/@glimmer/syntax/lib/v2-a/loose-resolution.ts
```diff
@@ -76,7 +76,7 @@ export function AppendSyntaxContext(node: ASTv1.MustacheStatement): ASTv2.FreeVa
       ? ASTv2.LooseModeResolution.trustingAppend({ invoke: isInvoke })
       : ASTv2.LooseModeResolution.append({ invoke: isInvoke });
   } else {
-    return isInvoke ? ASTv2.STRICT_RESOLUTION : ASTv2.LooseModeResolution.fallback();
+    return ASTv2.LooseModeResolution.fallback();
```
rwjblue, line 79: Hmm, does this introduce property fallback behavior?

rwjblue (commented):
Looks good, thanks for fixing issues with helper invocation from local paths!

I think we need some tests for those cases though (previously it was impossible).

packages/@glimmer/syntax/lib/v2-a/loose-resolution.ts
rwjblue: Chatted with @pzuraq, we already have tests confirming implicit this fallback does **not** happen for strict.

@pzuraq - if you have a chance, would you mind linking to that test?

rwjblue:
Chatted with @pzuraq and I think we need to add a couple more test cases for the new loose mode capability introduced here:

```hbs
{{foo.bar "derp"}}
```

* loose mode, `foo` is yielded block param with a `bar` prop that is a helper
* loose mode, `{{foo.bar "derp"}}` resolves to `this.foo.bar`  (and can be invoked)

For strict mode, we already have these tests:

* strict mode, foo is a local in scope that has a bar prop that is a helper:
* strict mode, foo is not a local in scope (should error)

packages/@glimmer/syntax/lib/v2-a/loose-resolution.ts
pzuraq: In Ember, this previously would have been compiled to `{{component foo.bar baz=123}}` or something like that, so I think the ship has sailed in that regard. That's why I included the change here.

rwjblue (approved):

PR #1236, opened 2020-12-15
[REFACTOR] Cache destroyable meta instead of destroyables directly

pzuraq:
Caches the destroyable meta rather than destroyables directly for block
opcodes, allowing us to avoid looking up the parent meta each time we
associate a new destroyable child value with the block. This should be
more efficient overall.


rwjblue:
@pzuraq - Is this something that we will end up wanting to go forward with?

pzuraq:
No, I don't think so, not at the moment anyways. We'd need to bench to see which strategy is actually more effective in the end, since this would require us to add a slot to destroyable meta.

PR #1237, opened 2020-12-16
[REFACTOR] Unifies currying for all values

pzuraq:
Unifies currying to be a single opcode with a particular type passed to
it, to de-duplicate the currying logic and ship less code overall.


packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/curry.ts
```diff
@@ -0,0 +1,56 @@
...
+    args: ASTv2.Args;
+  }> => {
+    let { args } = node;
+
+    let definition = args.nth(0);
+
+    if (definition === null) {
+      return Err(
+        generateSyntaxError(
+          `${type} requires a component definition or identifier as its first positional parameter, did not receive any parameters.`,
```
rwjblue, line 21: I think the message here is a bit wrong for helpers. The phrasing "component definition" doesn't make sense in a helper context.

packages/@glimmer/compiler/lib/wire-format-debug.ts
```diff
@@ -270,6 +264,19 @@ export default class WireFormatDebugger {
...
+  private formatCurryType(value: number) {
+    switch (value) {
+      case CurriedType.Component:
+        return 'component';
+      case CurriedType.Helper:
+        return 'helper';
+      case CurriedType.Modifier:
+        return 'modifier';
+      default:
+        throw new Error('unknown curry type');
```
rwjblue, line 276: Can you make this an `exhausted()` type thing (with a `never`, so that we can fail type checking if a new `CurriedType` is added that isn't accounted for)?

packages/@glimmer/runtime/lib/references/curry-value.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    if (isCurriedType(value, type)) {
+      curriedDefinition = args ? curry(type, value, owner, args) : args;
+    } else if (type === CurriedType.Component && typeof value === 'string' && value) {
+      // Only components should enter this path, as helpers and modifiers do not
+      // support string based resolution
+
+      if (DEBUG && isStrict) {
+        if (isStrict) {
```
rwjblue, line 39: This double `isStrict` check seems odd. Isn't the outer `if (DEBUG && isStrict) {` enough?

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/curry.ts
pzuraq: ahh, yes, definitely need to update that

packages/@glimmer/runtime/lib/references/curry-value.ts
pzuraq: yeah, that was a mistake, the `isStrict` should only be on the inside, and the whole block should execute regardless of strictness (only one assertion is related to strict)

packages/@glimmer/compiler/lib/wire-format-debug.ts
pzuraq: 👍

rwjblue (approved):

PR #1238, opened 2020-12-16
[FEATURE] Adds dynamic modifiers

pzuraq:
Adds dynamic modifiers and the `(modifier)` keyword. Modifiers can be
invoked dynamically in general using expression syntax.

One key difference between dynamic helpers and dynamic modifiers is that
dynamic modifiers do _not_ support `this` fallback. This is because
`this` fallback is not necessary for backwards compatibility, unlike
for helpers and components. Instead, a helpful error message is given,
which lets the user know they should consider adding `this` to the path.

### DIFF
--- a/packages/@glimmer/debug/lib/stack-check.ts
+++ b/packages/@glimmer/debug/lib/stack-check.ts
@@ -234,7 +234,7 @@ class ObjectChecker implements Checker<unknown> {
   type!: object;
 
   validate(obj: unknown): obj is object {
-    return typeof obj === 'object' && obj !== null;
+    return typeof obj === 'function' || (typeof obj === 'object' && obj !== null);
   }
 
   expected(): string {
--- a/packages/@glimmer/integration-tests/test/strict-mode-test.ts
+++ b/packages/@glimmer/integration-tests/test/strict-mode-test.ts
@@ -721,7 +721,7 @@ class DynamicStrictModeTest extends RenderTest {
     this.assertStableRerender();
   }
 
-  @test({ skip: true })
+  @test
   'Can use a dynamic modifier'() {
     const foo = defineSimpleModifier((element: Element) => (element.innerHTML = 'Hello, world!'));
     const Bar = defineComponent(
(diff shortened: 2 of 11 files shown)


packages/@glimmer/debug/lib/stack-check.ts
```diff
@@ -234,7 +234,7 @@ class ObjectChecker implements Checker<unknown> {
   type!: object;
 
   validate(obj: unknown): obj is object {
-    return typeof obj === 'object' && obj !== null;
+    return typeof obj === 'function' || (typeof obj === 'object' && obj !== null);
```
pzuraq, line 237: Originally was using this and realized it was incorrect, so I fixed it. Ended up not actually needing it, but wanted to make sure it was fixed.

packages/@glimmer/integration-tests/test/strict-mode-test.ts
```diff
@@ -904,31 +978,15 @@ class DynamicStrictModeTest extends RenderTest {
...
+      '{{#if (this.foo)}}{{/if}}',
       class extends GlimmerishComponent {
         foo = foo;
       }
     );
 
     this.assert.throws(() => {
       this.renderComponent(Bar);
-    }, /Error: Attempted to resolve a modifier in a strict mode template, but it was not in scope: foo/);
+    }, /Expected a dynamic helper definition, but received an object or function that did not have a helper manager associated with it. The dynamic invocation was `{{this.foo}}` or `\(this.foo\)`, and the incorrect definition is the value at the path `this.foo`, which was:/);
```
pzuraq, line 989: Realized this test was incorrect and fixed it tangentially, not directly related to this PR.

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -89,12 +97,16 @@ export function getInternalModifierManager(definition: object): InternalModifier
...
 
-  if (DEBUG && manager === undefined) {
-    throw new Error(
-      `Attempted to load a modifier, but there wasn't a modifier manager associated with the definition. The definition was: ${debugToString!(
-        definition
-      )}`
-    );
+  if (manager === undefined) {
+    if (isOptional === true) {
+      return null;
```
rwjblue, line 102: What is `isOptional` for conceptually?

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -93,16 +105,71 @@ APPEND_OPCODES.add(Op.Modifier, (vm, { op1: handle }) => {
...
-  let { manager } = definition;
 
-  let stack = vm.stack;
+  invokeModifier(vm, owner, args, definition);
+});
+
+APPEND_OPCODES.add(Op.DynamicModifier, (vm) => {
+  if (vm.env.isInteractive === false) {
+    return;
+  }
```
rwjblue, line 118: It seems weird to bake this assumption into the VM layer, no?

rwjblue (commented):
Need a few tests:

* confirm multi layer merging strategy for both positional and named arguments with `(modifier` (both strict and loose modes)
* using this paths to invoke a modifier (strict and loose)
* using a named argument to invoke a modifier without arguments (e.g. `<div {{@foo}}>`)
* using a named argument path (e.g. `<div {{@foo.bar 'baz'}}>`)

packages/@glimmer/manager/lib/internal/index.ts
pzuraq: For components and helpers, it's actually valid for a value to not be a component or a helper, but we still have to check if it is one (e.g. in append position, `{{maybeComponentOrHelper}}`), so that's why they have this capability. For modifiers, I added it so we can throw a better error message for dynamic modifiers specifically, because we have more context (what path the modifier was invoking from).

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
rwjblue: Looks like that ship has sailed unfortunately. 

https://github.com/glimmerjs/glimmer-vm/blob/3f29812a6b2904d2949a56c178babe1d9d8fa387/packages/%40glimmer/runtime/lib/compiled/opcodes/dom.ts#L92

I think this is conceptually wrong, and we should rethink but it is not a blocking concern for now.

pzuraq: I think for the moment it doesn't really, it's pretty baked in already that modifiers should only run in the browser. If we want to change that, I think it's going to be a larger refactor necessarily, and probably will look like us checking the capability/timing of the manager + `vm.env.isInteractive`, but doing it in approximately the same spot.

If we don't do it here, then we end up creating, scheduling, and running everything for the modifier, even though it isn't actually going to do anything, which is pretty negative for perf overall.

pzuraq:
Re: Tests

1. Have tests for strict-positional, will add them for named and dynamic: https://github.com/glimmerjs/glimmer-vm/pull/1238/files#diff-6591e40dde2e8325baf58e9e99ba1e8758a1c34a6a9aab9744ffc53d815cf679R778
2. Have this one for strict: https://github.com/glimmerjs/glimmer-vm/pull/1238/files#diff-6591e40dde2e8325baf58e9e99ba1e8758a1c34a6a9aab9744ffc53d815cf679R725
3. Have this one for strict, will add one for loose: https://github.com/glimmerjs/glimmer-vm/pull/1238/files#diff-6591e40dde2e8325baf58e9e99ba1e8758a1c34a6a9aab9744ffc53d815cf679R741 
4. Will add 👍

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
rwjblue: Its a bit odd to say its "very ingrained" since it is a new development in the VM space.

For example, it isn't baked into the VM of any release of Ember (all released versions of Ember that support modifiers, guard the interactivity in the Ember-side modifier manager).

pzuraq: That's not exactly true, that code was a double guard. We've guarded modifiers actually running in the VM in the environment for some time: https://github.com/glimmerjs/glimmer-vm/blob/v0.56.0/packages/@glimmer/runtime/lib/environment.ts#L330 

But yeah, happy to revisit.

rwjblue: For example, this change (currently slated for Ember 3.25 I think?) is plausibly introducing a breaking change for Ember users that use `data-ember-action` for styling **and** use FastBoot/SSR (this is a non-zero subset of users).

https://github.com/emberjs/ember.js/blob/907494ec67e516b7d213970545cd2527fe73001a/packages/%40ember/-internals/glimmer/lib/modifiers/action.ts#L280-L281

pzuraq: Prior to that, it was guarded in the environment in Ember: https://github.com/emberjs/ember.js/blob/v3.16.10/packages/@ember/-internals/glimmer/lib/environment.ts#L83

So as far as I can tell, modifiers have never run in the browser, at least back to 3.16

rwjblue (approved):

PR #1240, opened 2020-12-17
[FEATURE] Adds inline if and unless keywords

pzuraq:
Directly adds the inline versions of the `if` and `unless` keywords to
the VM. It does this by adding a new `IfInline` opcode, which is
combined with a new `Not` opcode that inverst the condition in the case
of `unless`. This also allows us to remove the `Unless` opcode and
rewrite it in terms of the `If` opcode in general. Since `unless` is
generally much less common of an opcode, this shouldn't be an issue
performance-wise, and we benefit from having a simpler set of
wireformat/opcodes.

### DIFF
--- a/packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
+++ b/packages/@glimmer/compiler/lib/passes/1-normalization/keywords/append.ts
@@ -8,6 +8,7 @@ import { VISIT_EXPRS } from '../visitors/expressions';
 import { keywords } from './impl';
 import { assertValidCurryUsage } from './utils/curry';
 import { assertValidHasBlockUsage } from './utils/has-block';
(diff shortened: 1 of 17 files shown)


lifeart:
wondering, could we have `not` instead of `unless` ?

* it's widely used outside ember
* easier to understand (for non-english speakers)
* less symbols to write

pzuraq:
@lifeart yeah, that's actually my preferred usage. `(not)` is currently in RFC, and the thinking is that this could be how it works under the hood. We need to decide if `not` is a keyword or a helper first though, so we'll see.

As for removing `unless`, it already exists in the language so it would need an RFC to remove it. This change means we don't have to pay for it conceptually at least, in the future we could discuss removing.

rwjblue (approved):

PR #1243, opened 2020-12-17
[FEATURE] Upstreams {{log}} keyword 

pzuraq:
Upstreams the `{{log}}` keyword, which can be used to log values for
debugging purposes.


rwjblue (commented):
Can you add some tests around updating a property and having the log called multiple times?

rwjblue (commented):
LGTM once we land some updating tests.

PR #1245, opened 2020-12-18
[FEATURE] Upstream built-in helpers and modifiers from Ember

pzuraq:
Upstreams the built-in helpers and modifiers from Ember:

- `on`
- `fn`
- `array`
- `hash`
- `get`
- `concat`

Tests were adapted directly from the existing test suites in Ember, with omissions wherever particular test functionality did not make sense (e.g. integration testing `get` with `<Input/>`).

## Breaking Changes

- `setPath` is now required on the global context

### DIFF
--- a/packages/@glimmer/global-context/index.ts
+++ b/packages/@glimmer/global-context/index.ts
@@ -101,6 +101,15 @@ export let setProp: (obj: object, prop: string, value: unknown) => void;
 export let getPath: (obj: object, path: string) => unknown;
 
 /**
+ * Hook for specifying how Glimmer should update paths in cases where it needs
+ * to. For instance, when updating a template reference (e.g. 2-way-binding)
+ *
+ * @param obj The object provided to get a value from
+ * @param path The path to get the value from
+ */
+export let setPath: (obj: object, path: string, value: unknown) => unknown;
+
+/**
  * Hook to warn if a style binding string or value was not marked as trusted
  * (e.g. HTMLSafe)
  */
@@ -117,6 +126,7 @@ export interface GlobalContext {
   getProp: (obj: object, path: string) => unknown;
   setProp: (obj: object, prop: string, value: unknown) => void;
   getPath: (obj: object, path: string) => unknown;
+  setPath: (obj: object, prop: string, value: unknown) => void;
   warnIfStyleNotTrusted: (value: unknown) => void;
 }
 
@@ -139,6 +149,7 @@ export default function setGlobalContext(context: GlobalContext) {
   getProp = context.getProp;
   setProp = context.setProp;
   getPath = context.getPath;
+  setPath = context.setPath;
   warnIfStyleNotTrusted = context.warnIfStyleNotTrusted;
 }
 
@@ -167,6 +178,7 @@ if (DEBUG) {
           getProp,
           setProp,
           getPath,
+          setPath,
           warnIfStyleNotTrusted,
         }
       : null;
@@ -185,6 +197,7 @@ if (DEBUG) {
     getProp = context?.getProp || getProp;
     setProp = context?.setProp || setProp;
     getPath = context?.getPath || getPath;
+    setPath = context?.setPath || setPath;
     warnIfStyleNotTrusted = context?.warnIfStyleNotTrusted || warnIfStyleNotTrusted;
 
     return originalGlobalContext;
--- a/packages/@glimmer/integration-tests/test/modifiers/on-test.ts
+++ b/packages/@glimmer/integration-tests/test/modifiers/on-test.ts
@@ -1,0 +1,452 @@
+import { castToBrowser, expect, HAS_NATIVE_PROXY } from '@glimmer/util';
+import { getInternalModifierManager } from '@glimmer/manager';
+import { on } from '@glimmer/runtime';
+
+import { jitSuite, RenderTest, test } from '../..';
+
+// check if window exists and actually is the global
+const hasDom =
+  typeof self === 'object' &&
+  self !== null &&
+  (self as Window['self']).Object === Object &&
+  typeof Window !== 'undefined' &&
+  self.constructor === Window &&
+  typeof document === 'object' &&
+  document !== null &&
+  self.document === document &&
+  typeof location === 'object' &&
+  location !== null &&
+  self.location === location &&
+  typeof history === 'object' &&
+  history !== null &&
+  self.history === history &&
+  typeof navigator === 'object' &&
+  navigator !== null &&
+  self.navigator === navigator &&
+  typeof navigator.userAgent === 'string';
+
+// do this to get around type issues for these values
+let global = window as any;
+const isChrome = hasDom
+  ? typeof global.chrome === 'object' && !(typeof global.opera === 'object')
+  : false;
+const isFirefox = hasDom ? typeof global.InstallTrigger !== 'undefined' : false;
+const isIE11 = !global.ActiveXObject && 'ActiveXObject' in window;
+
+interface Counters {
+  adds: number;
+  removes: number;
+}
+
+interface OnManager {
+  counters: Counters;
+  SUPPORTS_EVENT_OPTIONS: boolean;
+}
+
+function getOnManager() {
+  return (getInternalModifierManager(on) as unknown) as OnManager;
+}
+
+if (hasDom) {
+  class OnTest extends RenderTest {
+    static suiteName = '{{on}} Modifier';
+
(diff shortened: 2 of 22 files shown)


rwjblue:
Current failure is:

```
not ok 1393 Chrome 87.0 - [2 ms] - [integration] jit :: {{on}} Modifier: SUPPORTS_EVENT_OPTIONS is correct (private API usage)
    ---
        actual: >
            null
        stack: >
                at _loop (http://localhost:7357/153953091774/tests/assets/glimmer-vm.js:18481:21)
                at http://localhost:7357/153953091774/tests/assets/glimmer-vm.js:18491:11
                at suite (http://localhost:7357/153953091774/tests/assets/glimmer-vm.js:18493:9)
                at jitSuite (http://localhost:7357/153953091774/tests/assets/glimmer-vm.js:18407:12)
                at Module.callback (http://localhost:7357/153953091774/tests/assets/tests.js:13379:20)
                at Module.exports (http://localhost:7357/153953091774/tests/assets/loader.js:106:32)
                at requireModule (http://localhost:7357/153953091774/tests/assets/loader.js:27:18)
                at http://localhost:7357/153953091774/tests/index.html?hidepassed:139:11
        message: >
            Expected 0 assertions, but 1 were run
        negative: >
            false
        browser log: |
```

packages/@glimmer/runtime/lib/helpers/get.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+      },
+      'get'
+    );
+  }
+});
+
+function isObject(obj: unknown): obj is object {
+  return typeof obj === 'function' || (typeof obj === 'object' && obj !== null);
+}
```
rwjblue, line 131: Don't we have a util function for this somewhere?

packages/@glimmer/runtime/lib/helpers/get.ts
```diff
@@ -0,0 +1,131 @@
...
+    // child references will update accordingly.
+    let path = valueForRef(pathRef);
+
+    if (path === undefined || path === null || path === '') {
+      return NULL_REFERENCE;
+    } else if (typeof path === 'string' && path.indexOf('.') > -1) {
+      return childRefFromParts(sourceRef, path.split('.'));
+    } else {
+      return childRefFor(sourceRef, String(path));
+    }
```
rwjblue, line 107: It seems vaguely odd to hard code the details of `getPath` here. In the else case just below we directly use `getPath` which can be customized by the host, but in this branch we manually encode _one possible_ implementation of `getPath`. IMHO, either the host is allowed to customize `getPath` behavior or it isn't. If it _can_ this truthy branch shouldn't make assumptions about how `getPath` works.

The concrete concern here is that I'd vaguely prefer to _avoid_ making Glimmer.js have the same `.` means path behavior (and therefore preventing users from being able to use `get` to access properties with `.` in them). Things like https://github.com/emberjs/ember.js/issues/17529 https://github.com/emberjs/ember.js/issues/17927 are discussing, but even if we don't actually want to support that this branch represents a "leaky abstraction" (and likely suggests that we don't need `getPath` on global context at all).

packages/@glimmer/util/lib/untouchable-this.ts
```diff
@@ -0,0 +1,38 @@
...
+import { HAS_NATIVE_PROXY } from './platform-utils';
+
+export default function buildUntouchableThis(source: string): null | object {
+  let context: null | object = null;
+  if (DEBUG && HAS_NATIVE_PROXY) {
+    let assertOnProperty = (property: string | number | symbol) => {
+      throw new Error(
+        `You accessed \`this.${String(
+          property
+        )}\` from a function passed to the ${source}, but the function itself was not bound to a valid \`this\` context. Consider updating to usage of \`@action\`.`
```
rwjblue, line 11: What is the right path for users in Glimmer.js here? I doubt `@action` is right, is it?

packages/@glimmer/integration-tests/test/modifiers/on-test.ts
```diff
@@ -0,0 +1,458 @@
...
+    @test
+    'SUPPORTS_EVENT_OPTIONS is correct (private API usage)'(assert: Assert) {
+      let { SUPPORTS_EVENT_OPTIONS } = getOnManager();
+
+      if (isChrome || isFirefox) {
+        assert.strictEqual(SUPPORTS_EVENT_OPTIONS, true, 'is true in chrome and firefox');
+      } else if (isIE11) {
+        assert.strictEqual(SUPPORTS_EVENT_OPTIONS, false, 'is false in IE11');
+      } else {
+        assert.expect(0);
```
rwjblue, line 88: This assertion is failing CI because there is an `afterEach` assertion. I think this branch should just be removed.

```suggestion
```

rwjblue (changes requested):
* Why are the various test files empty? 
  * packages/@glimmer/integration-tests/test/helpers/array-test.ts
  * packages/@glimmer/integration-tests/test/helpers/fn-test.ts
  * packages/@glimmer/integration-tests/test/helpers/get-test.ts
  * packages/@glimmer/integration-tests/test/helpers/hash-test.ts
  * packages/@glimmer/integration-tests/test/helpers/concat-test.ts
  * packages/@glimmer/integration-tests/test/helpers/array-test.ts

packages/@glimmer/integration-tests/test/keywords/array-test.ts
```diff
@@ -0,0 +1,314 @@
+import { jitSuite, RenderTest, strip, test, GlimmerishComponent, tracked } from '../..';
```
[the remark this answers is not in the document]
pzuraq: Yeah, this was a mistake, I meant to put them in the `/helpers` directory but was tired and duplicated them on accident haha, thus the empty test files as well.

packages/@glimmer/global-context/index.ts
```diff
@@ -100,6 +100,15 @@ export let setProp: (obj: object, prop: string, value: unknown) => void;
...
 export let getPath: (obj: object, path: string) => unknown;
 
+/**
+ * Hook for specifying how Glimmer should update paths in cases where it needs
+ * to. For instance, when updating a template reference (e.g. 2-way-binding)
+ *
+ * @param obj The object provided to get a value from
+ * @param path The path to get the value from
+ */
+export let setPath: (obj: object, path: string, value: unknown) => unknown;
```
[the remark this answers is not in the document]
pzuraq: This is specifically for `{{get}}`. In general, when we see a path in Glimmer, we break it down into a chain of references. However, when the path is dynamic, we cannot do that, we have to do it all within a single reference. We already had `getPath` because there was an instance of this before for getting a path (the `key` on `each`) but we never encountered a `setPath` before. `{{get}}` can also _update_ the value that is accessed (similar to how `mut` works), so that's why we need the ability to set a path - it's dynamic and not broken down into pieces.

packages/@glimmer/runtime/lib/helpers/get.ts
pzuraq: That seems reasonable, though I do think this could be an issue for cross-compatibility (annoyingly, as I agree). We can probably remove this branch and just make the default the dynamic path, I don't think it would be likely to have an impact on perf

packages/@glimmer/util/lib/untouchable-this.ts
pzuraq: ah, yes. We could say "a bound function", and give an arrow function as an example?

packages/@glimmer/runtime/lib/helpers/get.ts
pzuraq: we have one that is `isObject` in `@glimmer/util`, but it does not check if the value is a `function` 🙃 I was planning on doing a sweep to deduplicate it soon (and also confirm with @krisselden that deduplicating it wouldn't cause any perf issues)

pzuraq: Opened up an issue for this: https://github.com/glimmerjs/glimmer-vm/issues/1246

packages/@glimmer/global-context/index.ts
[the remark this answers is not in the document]
rwjblue: So to clarify, its for `(mut (get` being used together (since I didn't think that `{{something whatever=(get otherThing 'lol.path')}} allowed mutation of `otherThing.lol.path`)? Is that right?

rwjblue (approved):

PR #1249, opened 2020-12-21
Add test for usage of `{{@component.foo}}`

rwjblue:
Initial attempts to reproduce emberjs/ember.js#19313


pzuraq (approved):

pzuraq (approved):

lifeart:
wondering, should we add same check for `this.component` ?

rwjblue:
> wondering, should we add same check for this.component ?

@lifeart - What do you mean?

lifeart:
@rwjblue I mean `this.component` is not recognized as component helper

PR #1253, opened 2021-01-12
[BUGFIX] Convert component names after symbol allocation

pzuraq:
Currently component names are converted via the `customizeComponentName`
option prior to symbol allocation. This means that a component can end
up conflicting with an existing local variable, even though they
logically represent different things. This PR ensures that customization
happens after symbol allocation, as a final step within the symbol table
itself. This also means that the correct value is passed through the
rest of the AST and transformations, so it cannot conflict with keywords
as well as other language constructs.

Fixes: https://github.com/emberjs/ember.js/issues/19334

### DIFF
--- a/packages/@glimmer/syntax/lib/v2-a/normalize.ts
+++ b/packages/@glimmer/syntax/lib/v2-a/normalize.ts
@@ -38,7 +38,11 @@ export function normalize(
     options
   );
 
-  let top = SymbolTable.top(normalizeOptions.strictMode ? normalizeOptions.locals : []);
+  let top = SymbolTable.top(
+    normalizeOptions.strictMode ? normalizeOptions.locals : [],
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    options.customizeComponentName ?? ((name) => name)
+  );
   let block = new BlockContext(source, normalizeOptions, top);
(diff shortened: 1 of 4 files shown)


rwjblue:
Can you cross link the related ember issue for easier future archeology 😜

packages/@glimmer/syntax/lib/v2-a/normalize.ts
```diff
@@ -644,14 +650,6 @@ class ElementNormalizer {
     let pathLoc = variableLoc.withEnd(pathEnd);
 
     if (isComponent) {
-      // If the component name is uppercase, the variable is not in scope,
-      // and the template is not in strict mode, run the optional
-      // `customizeComponentName` function provided as an option to the
-      // precompiler.
-      if (!this.ctx.strict && uppercase && !inScope) {
```
rwjblue, line 651: This uppercase check is removed, but I don’t see it brought back. 

Can you add a test that confirms a template without a block param never resolves for `<rental></rental>`.

pzuraq: Test added!

rwjblue (approved):

rwjblue:
Still need to cross link the Ember issue in the description.

PR #1255, opened 2021-01-26
Don't call `customizeComponentName` on curly components

dfreeman:
The change in https://github.com/glimmerjs/glimmer-vm/pull/1253 caused `customizeComponentName` to be called on block-form curly components, while before it was only invoked for angle-bracket components. For anyone making *ahem* questionable use of `::` in curly component names in Ember, this meant those would always be converted to `/` instead.

This change restores the previous behavior and adds a test to cover it. Thanks to @pzuraq for confirming what was going on and the likely fix!

### DIFF
--- a/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
+++ b/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
@@ -72,6 +72,29 @@ module('[glimmer-compiler] precompile', ({ test }) => {
     assert.equal(componentName, 'rental', 'customized component name was used');
   });
 
+  test('customizeComponentName is not invoked on curly components', function (assert) {
+    let wire = JSON.parse(
+      precompile('{{#my-component}}hello{{/my-component}}', {
+        customizeComponentName(input: string) {
+          return input.toUpperCase();
+        },
+      })
+    );
+
+    let block: WireFormat.SerializedTemplateBlock = JSON.parse(wire.block);
+
+    let [[, componentNameExpr]] = block[0] as [WireFormat.Statements.Block];
+
+    glimmerAssert(
+      Array.isArray(componentNameExpr) &&
+        componentNameExpr[0] === SexpOpcodes.GetFreeAsComponentHead,
+      `component name is a free variable lookup`
+    );
+
+    let componentName = block[3][componentNameExpr[1]];
+    assert.equal(componentName, 'my-component', 'original component name was used');
+  });
+
   test('lowercased names are not resolved or customized in resolution mode', (assert) => {
     let wire = JSON.parse(
       precompile('<rental />', {
--- a/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
+++ b/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
@@ -95,6 +95,29 @@ module('[glimmer-compiler] precompile', ({ test }) => {
     assert.equal(componentName, 'my-component', 'original component name was used');
   });
 
+  test('customizeComponentName is not invoked on angle-bracket-like name invoked with curlies', function (assert) {
+    let wire = JSON.parse(
+      precompile('{{#MyComponent}}hello{{/MyComponent}}', {
+        customizeComponentName(input: string) {
+          return input.toUpperCase();
+        },
+      })
+    );
+
+    let block: WireFormat.SerializedTemplateBlock = JSON.parse(wire.block);
+
+    let [[, componentNameExpr]] = block[0] as [WireFormat.Statements.Block];
+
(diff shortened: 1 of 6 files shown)


packages/@glimmer/syntax/lib/symbol-table.ts
```diff
@@ -86,7 +87,9 @@ export class ProgramSymbolTable extends SymbolTable {
   }
 
   allocateFree(name: string, resolution: ASTv2.FreeVarResolution): number {
-    if (resolution.resolution() === SexpOpcodes.GetFreeAsComponentHead) {
+    // If the name in question is an uppercase (i.e. angle-bracket) component invocation, run
+    // the optional `customizeComponentName` function provided to the precompiler.
+    if (resolution.resolution() === SexpOpcodes.GetFreeAsComponentHead && isUpperCase(name)) {
```
rwjblue, line 92: We chatted a bit about this in discord, but I think even after this change we still have an issue. I would expect (for example) that `{{#Foo}}{{/Foo}}` should _also_ not call `customizeComponentName`, prior to #1253 we _only_ attempted transformation for `element.tag` specifically.

packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
```diff
@@ -72,6 +72,29 @@ module('[glimmer-compiler] precompile', ({ test }) => {
...
+
+    glimmerAssert(
+      Array.isArray(componentNameExpr) &&
+        componentNameExpr[0] === SexpOpcodes.GetFreeAsComponentHead,
+      `component name is a free variable lookup`
+    );
+
+    let componentName = block[3][componentNameExpr[1]];
+    assert.equal(componentName, 'my-component', 'original component name was used');
+  });
```
rwjblue, line 96: Can you make another (I think failing) test that looks like:

```hbs
{{#Foo}}hello{{/Foo}}
```

And also confirms that `customizeComponentName` is not called? I _believe_ that the current implementation will fail a test like this, but prior to #1253 would have passed.

rwjblue (commented):
I think this fix is a nice stepping stone (and the test cases are good), but I think we still have a bit more to fix up here...

packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
jamescdavis: Looks like you're right: https://github.com/dfreeman/glimmer-vm/pull/1/checks?check_run_id=1772791358#step:5:296

dfreeman:
Thanks for the review, @rwjblue! Chatting with @pzuraq, it sounds like this is going to be a bit more complex to completely restore the previous behavior.

We generally treat a green `ember-release` build in CI as required for our internal addons, since the canary/beta train usually gives us plenty of time to react to incoming changes. Given that this entered the scene relatively late in the 3.25 beta cycle, what do you think the likelihood is of either landing the partial fix or reverting the change before 3.25 becomes stable?

rwjblue:
The issue is that reverting #1253 also leaves us in a bad state, the following is broken (and was the original reported issue in https://github.com/emberjs/ember.js/issues/19334):

```hbs
<div class="rentals">
  <ul class="results">
    {{#each @model as |rental|}}
      <li><Rental @rental={{rental}} /></li>
    {{/each}}
  </ul>
</div>
```

Prior to #1253 this snippet would assume that `<Rental />` was invoking the block param `rental` (because `customizeComponentName('Rental')` results in `rental`).

rwjblue:
@dfreeman / @jamescdavis - Would one y'all mind opening an issue on the Ember side for this also? I would love to make sure that there is something that others running into issues in the 3.25.0-beta series could find upon a search over there (since not everyone understands the repo relationships).

jamescdavis:
I fixed up #1256 by making the no conflict test a todo, just for illustrative purposes (or, you know, if all else fails).

packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
dfreeman: Done (thanks @jamescdavis!)

rwjblue (approved):

chancancode:
When this is incorporated into beta, can someone open a PR to revert https://github.com/ember-learn/super-rentals-tutorial/pull/172/commits/cc75067669cf84ce881dd3f97496494b9a6b4c31 and confirm we can get a clean build?

rwjblue:
> When this is incorporated into beta

The PR is still pending https://github.com/emberjs/ember.js/pull/19363

jamescdavis:
@chancancode https://github.com/ember-learn/super-rentals-tutorial/pull/177

PR #1257, opened 2021-01-27
Add `assert` and `deprecate` to global context API

pzuraq:
This PR adds `assert` and `deprecate` to the global context API, and adds a new
package: `@glimmer/vm-babel-plugins`. This plugin exports a function that can
be used to generate Babel plugins which strip debug tooling in the VM. The idea
is that Ember/Glimmer.js will be able to re-export and use this in their own
Babel configs/utilities.

This also removes the `deprecate` and `assert` functions from the TransactionEnvironnment,
since these new utilities supersede those.

Integrated into Ember to ensure it all works.

### DIFF
--- a/packages/@glimmer/validator/lib/debug.ts
+++ b/packages/@glimmer/validator/lib/debug.ts
@@ -1,5 +1,6 @@
 import { Tag } from './validators';
 import { DEBUG } from '@glimmer/env';
+import { deprecate, assert } from '@glimmer/global-context';
 
 export let beginTrackingTransaction:
   | undefined
@@ -13,11 +14,7 @@ export let deprecateMutationsInTrackingTransaction: undefined | ((fn: () => void
 export let resetTrackingTransaction: undefined | (() => string);
 export let setTrackingTransactionEnv:
   | undefined
-  | ((env: {
-      assert?(message: string): void;
-      deprecate?(message: string): void;
-      debugMessage?(obj?: unknown, keyName?: string): string;
-    }) => void);
+  | ((env: { debugMessage?(obj?: unknown, keyName?: string): string }) => void);
 
 export let assertTagNotConsumed:
   | undefined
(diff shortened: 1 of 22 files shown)


packages/@glimmer/vm-babel-plugins/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Compiles out VM assertion and deprecation utilities and debug tooling based on environment",
+  "repository": "https://github.com/glimmerjs/glimmer.js",
+  "author": "Tom Dale <tom@tomdale.net>",
+  "license": "MIT",
+  "private": false,
+  "scripts": {
+    "test": "mocha -r esm"
+  },
+  "dependencies": {
+    "babel-plugin-debug-macros": "^0.3.3"
```
rwjblue, line 13: Will want to bump to `^0.3.4` now

https://github.com/ember-cli/babel-plugin-debug-macros/releases/tag/v0.3.4

ember-cli-build.js
```diff
@@ -21,6 +21,7 @@ module.exports = function () {
   // relative to the project root and we want to use the tsconfig as-is.
   let tsTree = funnel('packages/@glimmer', {
     destDir: 'packages/@glimmer',
+    exclude: ['vm-babel-plugins'],
```
rwjblue, line 24: Do we need to exclude? What breaks if we don't exclude?

packages/@glimmer/validator/lib/debug.ts
```diff
@@ -41,14 +38,14 @@ if (DEBUG) {
...
+    // },
 
-    deprecate(message: string): void {
-      // eslint-disable-next-line no-console
-      console.warn(message);
-    },
+    // deprecate(message: string): void {
+    //   // eslint-disable-next-line no-console
+    //   console.warn(message);
+    // },
```
rwjblue, line 48: 🤔 ✂️

rwjblue (approved):

ember-cli-build.js
pzuraq: Ah, we don't actually need this, this was before I moved the package into the `build-utils` sub-directory. TL;DR is the whole `packages/@glimmer` directory is really tied into being compiled for client side execution. Build utils are just a different category, so I decided to move it out.

PR #1261, opened 2021-02-06
Make defaultId function more robust

pzuraq:
The current `defaultId` function for templates attempts to load the `crypto` module
when in Node. However, it uses `require` directly, which may be
clobbered by AMD require (in the case of Ember). This was previously
handled by using Babel transforms, but that solution is somewhat
brittle. This updates the logic to check for the `module` global
instead, and defers to it if it exists, before checking for `require`.

### DIFF
--- a/packages/@glimmer/compiler/lib/compiler.ts
+++ b/packages/@glimmer/compiler/lib/compiler.ts
@@ -23,10 +23,12 @@ interface Crypto {
 }
 
 export const defaultId: TemplateIdFn = (() => {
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

rwjblue:
Can you cross link any issues that this fixes?

pzuraq:
Superseded by #1262

pzuraq:
Just to note, this PR did not solve any specific issue, it was related to the refactoring for the build pipeline in Ember.

PR #1263, opened 2021-02-10
[BUGFIX] Ensure owner is available during resolution for all components

pzuraq:
Previously, strict components were not passed an owner when they were
defined. This turned out to be incorrect, because even if strict
components did not use an owner to resolve components, the components
that they used in scope could potentially be non-strict, and resolve
components. Those components effectively become different components
based on the currently active owner, so we have to pass the owner no
matter what type of component we are rendering.

### DIFF
--- a/packages/@glimmer/integration-tests/test/owner-test.ts
+++ b/packages/@glimmer/integration-tests/test/owner-test.ts
@@ -1,4 +1,10 @@
-import { ResolvedComponentDefinition } from '@glimmer/interfaces';
+import {
+  InternalComponentCapabilities,
+  Owner,
+  ResolvedComponentDefinition,
+  WithCreateInstance,
+  WithSubOwner,
+} from '@glimmer/interfaces';
 import {
   test,
   suite,
@@ -7,7 +13,11 @@ import {
   JitRenderDelegate,
   createTemplate,
   TestJitRuntimeResolver,
+  GlimmerishComponent,
+  defineComponent,
 } from '..';
+import { NULL_REFERENCE, Reference } from '@glimmer/reference';
+import { setInternalComponentManager } from '@glimmer/manager';
 
 class OwnerJitRuntimeResolver extends TestJitRuntimeResolver {
   lookupComponent(name: string, owner: () => void): ResolvedComponentDefinition | null {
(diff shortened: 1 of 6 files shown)


packages/@glimmer/integration-tests/test/owner-test.ts
```diff
@@ -81,6 +169,34 @@ class OwnerTest extends RenderTest {
 
     assert.verifySteps(['foo-bar owner called', 'foo-baz owner called'], 'owners used correctly');
   }
+
+  @test
+  'owner can be changed by a component with the hasSubOwner capability'(assert: Assert) {
```
pzuraq, line 174: This test is unrelated, I just realized we didn't have it (or at least not an obvious one for it) so wanted to add it

krisselden (changes requested):
Can you add a test that passes a closure component into the model property of the mount and invoke that component inside of the mount with a yield block that invokes a component coming from the mount?

```
{{mount 'engine' model=MyComponent}}
```

inside the mounted 

```
{{#let @model as |OuterComponent|}}
   <OuterComponent>
      <InnerComponent/>
  </OuterComponent>
{{/let}}
```

I want to ensure you didn't conflate the owner (realm) concept with dynamic scope.

packages/@glimmer/program/lib/constants.ts
```diff
@@ -196,15 +196,16 @@ export class ConstantsImpl
...
   component(
     definitionState: ComponentDefinitionState,
+    object: object,
     isOptional: true
   ): ComponentDefinition | null;
   component(
     definitionState: ComponentDefinitionState,
-    isOptional?: true,
-    owner?: object
+    owner: object,
```
krisselden, line 207: if owner is optional shouldn't this allow undefined?

krisselden: I don't really understand this overload

krisselden (approved):

packages/@glimmer/integration-tests/test/owner-test.ts
```diff
@@ -81,6 +169,73 @@ class OwnerTest extends RenderTest {
...
+
+    const Mount2 = defineMountComponent(owner2, { CheckOwner2 }, `<CheckOwner2/><@CheckOwner2/>`);
+    const Mount1 = defineMountComponent(
+      owner1,
+      { CheckOwner1, CheckOwner2, Mount2 },
+      `<CheckOwner1/><Mount2 @CheckOwner2={{CheckOwner2}}/>`
+    );
+
+    this.renderComponent(Mount1);
+  }
```
rwjblue, line 228: For context, we (@pzuraq + @krisselden + myself) discussed this in detail. The rough conclusion is that we don't really like the thing this test is asserting (it is very strange to end up with an owner from a different invocation than where you were originally associated), but it is quite difficult to ensure that we **always** have the right owner (it isn't even possible to absorb all variants of things that could be done in JavaScript land).

At the moment, we intend to land this update over in Ember and add a debug only AST transform that wraps any values passed to `{{mount 'foo' model=` (because as far as we can tell, that is the only way to pass things "across owner / realm") and asserts if you pass something that is a component/helper/modifier that isn't wrapped in `(component`/`(helper`/`(modifier` helpers.

packages/@glimmer/program/lib/constants.ts
```diff
@@ -196,15 +196,11 @@ export class ConstantsImpl
...
   component(
     definitionState: ComponentDefinitionState,
-    isOptional: true
-  ): ComponentDefinition | null;
-  component(
-    definitionState: ComponentDefinitionState,
-    isOptional?: true,
-    owner?: object
+    owner: object,
+    isOptional?: true
```
krisselden, line 203: isOptional?: true should be boolean, I don't think the narrowness helps here.

PR #1267, opened 2021-02-12
[BUGFIX] Fixes dynamic helpers and modifiers with updating definitions

pzuraq:
Currently, when the definition of a dynamic helper/modifier is updated,
the instance of that modifier or helper is not also updated. This was an
oversight from the initial implementation. This PR adds this
functionality, and also fixes a bug where sub-expressions could not be
the callee in an expression (e.g. `{{(if @foo Bar) "baz"}}`).

Notes:

- Modifiers and helpers have been refactored to receive
  CapturedArguments instead of VMArguments. In general, most helpers and
  modifiers immediately capture the arguments anyways, so this shouldn't
  affect performance. This is necessary because dynamic
  helpers/modifiers _must_ captured the arguments, otherwise they could
  change, and if the helper or modifier is updated to a new definition,
  it would get the wrong arguments.
- Helpers no longer receive the VM itself. Instead, they receive the
  `owner` they belong to, and optionally they receive `dynamicScope`.
  This refactor unifies the way helpers are invoked in the VM and by
  `invokeHelper`, which means `invokeHelper` can now be used with
  built-in helpers such as `array` and `hash`.
- Only static helpers (e.g. non-dynamic, non-`invokeHelper`) receive
  `dynamicScope`. The only helper that needs this is `{{outlet}}`, which
  does not make sense to invoke dynamically currently. We will assert
  with a good error message in the implementation in Ember.
- Helpers that wish to associate destroyables should now associate them
  directly on the _reference_ they return. This allows us to have a more
  streamlined and unified API overall. Generally, only custom helpers
  and a few select built-in helpers need this ability.
- `hasDestroyableChildren` has been added in order to prevent
  associating non-custom helpers if they don't have any children. This
  should mean less overall memory usage/CPU cost in those cases, which
  are likely the most common.

As an aside, this refactor does make me feel like even moreso that we
should rethink the way `Helper` is represented in the VM. The current
structure has evolved organically to handle custom helpers, but it feels
not ideal. The VM should likely be more aware of custom helpers in some
way, but that's a larger refactor.


packages/@glimmer/benchmark-env/src/benchmark/if-helper.ts
```diff
@@ -1,18 +0,0 @@
-import { VMArguments } from '@glimmer/interfaces';
-import { Reference, createComputeRef, valueForRef } from '@glimmer/reference';
-
-export default function ifHelper({ positional }: VMArguments): Reference {
```
pzuraq, line 4: This helper is no longer used, because `{{if}}` is now a keyword. Leaving it was an oversight, and since the signature changed I noticed because I had to update it.

chancancode (commented):
> ￼ This is necessary because dynamic
helpers/modifiers must captured the arguments, otherwise they could
change, and if the helper or modifier is updated to a new definition,
it would get the wrong arguments.

Hm why is that? Specifically: the way I understand this is that this is brining the same functionality that always worked for components ("contextual components") to helpers and modifiers, so why is it that it works for components but not these?

pzuraq:
Dynamic components trigger an assertion, which puts the VM back into append mode. In append mode, we have the ability to use the proper VMArguments stack again, because it's like we're appending the component for the first time.

Dynamic helpers and modifiers, by contrast, do not trigger append mode, because they are not actually adding/removing DOM (via the VM at least). They use the updating VM instead, and so the updating VM needs to keep references to the captured arguments that the helper/modifier was invoked with originally.

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -194,15 +144,114 @@ function invokeModifier(
...
+  let ref = check(stack.popJs(), CheckReference);
+  let args = check(stack.popJs(), CheckArguments).capture();
+  let { constructing } = vm.elements();
+  let initialOwner = vm.getOwner();
+
+  let instanceRef = createComputeRef(() => {
+    let value = valueForRef(ref);
+    let owner: Owner;
+
+    if (typeof value !== 'function' && (typeof value !== 'object' || value === null)) {
```
rwjblue, line 166: 🤕

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -194,15 +144,114 @@ function invokeModifier(
...
+
+      hostDefinition = resolvedDefinition;
+      owner = curriedOwner;
+
+      if (positional !== undefined) {
+        args.positional = positional.concat(args.positional) as CapturedPositionalArguments;
+      }
+
+      if (named !== undefined) {
+        args.named = assign(named, args.named);
```
rwjblue, line 188: Chatted with @pzuraq RE: this, and `named` is a fresh object created by `resolveCurriedValue` which is why this mutation is safe.

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -219,6 +268,63 @@ export class UpdateModifierOpcode extends UpdatingOpcode {
...
+    if (newInstance !== instance) {
+      if (instance) {
+        let destroyable = instance.manager.getDestroyable(instance.state);
+
+        if (destroyable !== null) {
+          destroy(destroyable);
+        }
+      }
+
+      if (newInstance) {
```
rwjblue, line 298:
```suggestion
      if (newInstance !== undefined) {
```

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -219,6 +268,63 @@ export class UpdateModifierOpcode extends UpdatingOpcode {
...
+    this.lastUpdated = valueForTag(tag ?? CURRENT_TAG);
+  }
+
+  evaluate(vm: UpdatingVM) {
+    let { tag, lastUpdated, instance, instanceRef } = this;
+
+    let newInstance = valueForRef(instanceRef);
+
+    if (newInstance !== instance) {
+      if (instance) {
```
rwjblue, line 290:
```suggestion
      if (instance !== undefined) {
```

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
```diff
@@ -72,28 +75,57 @@ APPEND_OPCODES.add(Op.Curry, (vm, { op1: type, op2: _isStrict }) => {
...
   let ref = vm.fetchValue<Reference>(_definitionRegister);
-  let definition = valueForRef(ref);
 
-  if (isCurriedType(definition, CurriedType.Helper)) {
-    let [resolvedDef, owner] = resolveCurriedValue(definition, args);
+  let helperRef: Reference;
+  let initialOwner: Owner = vm.getOwner();
 
-    let helper = resolveHelper(vm[CONSTANTS], resolvedDef, ref);
+  let helperDefinitionRef = createComputeRef(() => {
```
rwjblue, line 84: Maybe call this "helper instance ref"?

packages/@glimmer/destroyable/index.ts
```diff
@@ -191,6 +191,12 @@ export function destroyChildren(destroyable: Destroyable) {
   iterate(children, destroy);
 }
 
+export function hasDestroyableChildren(destroyable: Destroyable) {
```
rwjblue, line 194: Maybe underscore? I'm vaguely worried that folks in glimmer.js land might have access to it directly...

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
```diff
@@ -72,28 +75,57 @@ APPEND_OPCODES.add(Op.Curry, (vm, { op1: type, op2: _isStrict }) => {
...
+      if (positional !== undefined) {
+        args.positional = positional.concat(args.positional) as CapturedPositionalArguments;
+      }
+
+      helperRef = helper(args, owner);
+
+      associateDestroyableChild(helperDefinitionRef, helperRef);
+    } else if (
+      typeof definition === 'function' ||
+      (typeof definition === 'object' && definition !== null)
```
rwjblue, line 109: Can we use the shared util for "is valid weakmap key"?

packages/@glimmer/runtime/lib/curried-value.ts
```diff
@@ -46,34 +45,40 @@ export class CurriedValue<T extends CurriedType = CurriedType> {
...
-            : curriedArgs.positional.concat(prependArgs);
+      let { named: curriedNamed, positional: curriedPositional } = curriedArgs;
+
+      if (curriedPositional.length > 0) {
+        positional =
+          positional === undefined ? curriedPositional : curriedPositional.concat(positional);
       }
 
-      args.named.merge(curriedArgs.named);
+      named = named === undefined ? assign({}, curriedNamed) : assign(curriedNamed, named);
```
rwjblue, line 81: Can you double check this? It seems odd to have `assign(named, curriedNamed)` as the last branch...

Maybe think about how we can test? (if we can)

packages/@glimmer/runtime/lib/helpers/invoke.ts
```diff
@@ -53,51 +55,27 @@ export function invokeHelper(
...
-      'Found a helper manager, but it was an internal built-in helper manager. `invokeHelper` does not support internal helpers yet.'
-    );
-  }
-
-  const manager = (internalManager as InternalHelperManager<object>).getDelegateFor(owner);
   let args = new SimpleArgsProxy(context, computeArgs);
-  let bucket = manager.createHelper(definition, args);
-
-  let cache: Cache<unknown>;
+  let helperRef = helper(args as CapturedArguments, owner);
```
rwjblue, line 74: Double check this, `args` is not a `CapturedArguments` (its a proxy to values, not a bunch of references).

Also need tests.

packages/@glimmer/runtime/lib/helpers/invoke.ts
```diff
@@ -53,51 +55,27 @@ export function invokeHelper(
...
-          `You attempted to get the value of a helper after the helper was destroyed, which is not allowed`
-        );
-      }
-
-      return manager.getValue(bucket);
-    });
-
-    associateDestroyableChild(context, cache);
-  } else {
-    throw new Error('TODO: unreachable, to be implemented with hasScheduledEffect');
```
rwjblue, line 93: Where did this error go?

rwjblue (changes requested):

packages/@glimmer/benchmark-env/src/benchmark/on-modifier.ts
```diff
@@ -13,11 +13,11 @@ interface OnModifierState {
 }
 
 class OnModifierManager implements InternalModifierManager<OnModifierState, object> {
```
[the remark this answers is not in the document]
pzuraq: This is true, but means we need to refactor the benchmark-env a bit since the built-in on modifier needs to be imported. We should do this in a followup

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
rwjblue: One thing that struck me, is that we could pass `args.named` in to `resolveCurriedValue` and avoid multiple `Object.assign`s.

packages/@glimmer/runtime/lib/curried-value.ts
```diff
@@ -46,34 +45,40 @@ export class CurriedValue<T extends CurriedType = CurriedType> {
...
-            : curriedArgs.positional.concat(prependArgs);
+      let { named: curriedNamed, positional: curriedPositional } = curriedArgs;
+
+      if (curriedPositional.length > 0) {
+        positional =
+          positional === undefined ? curriedPositional : curriedPositional.concat(positional);
       }
 
-      args.named.merge(curriedArgs.named);
+      named = named === undefined ? assign({}, curriedNamed) : assign({}, curriedNamed, named);
```
rwjblue, line 81: Hmm, why do we need this ternary? Object.assign ignores undefined/null values, and so does our helper

https://github.com/glimmerjs/glimmer-vm/blob/611b2e4b2fe3fe2677ff31b10c8350eb52ee8609/packages/%40glimmer/util/lib/object-utils.ts#L28

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -195,15 +145,114 @@ function invokeModifier(
...
+  let ref = check(stack.popJs(), CheckReference);
+  let args = check(stack.popJs(), CheckArguments).capture();
+  let { constructing } = vm.elements();
+  let initialOwner = vm.getOwner();
+
+  let instanceRef = createComputeRef(() => {
+    let value = valueForRef(ref);
+    let owner: Owner;
+
+    if (!isObject(value)) {
```
rwjblue, line 167: 🙏

rwjblue (approved):

PR #1268, opened 2021-02-12
[BUGFIX] Fix reactivity of dynamic attributes

pzuraq:
The updates to use autotracking through the VM resulted in a bug in
dynamic attributes. Previously, the updating opcode for dynamic
attributes would run its update function based on tags/autotracking -
whenever any tag updated (e.g. a tracked value was changed) the
attribute would call its `update` function. The `update` function would
then check the new value against the current value of the attribute, and
update if they were different.

The bug was we introduced a value comparison in the updating opcode
itself, based on the last value of the property to set the attribute to.
This meant that if the property was updated to the same value twice in
a row, it would not call `update`, even if the attribute's value had
changed (e.g. because the user typed something).

Fixes https://github.com/emberjs/ember.js/issues/19341

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
@@ -1,4 +1,4 @@
-import { Reference, valueForRef, isConstRef } from '@glimmer/reference';
+import { Reference, valueForRef, isConstRef, createComputeRef } from '@glimmer/reference';
 import { Revision, Tag, valueForTag, validateTag, consumeTag } from '@glimmer/validator';
 import {
   check,
@@ -17,6 +17,7 @@ import {
   Owner,
   CurriedType,
   ModifierDefinitionState,
+  Environment,
 } from '@glimmer/interfaces';
 import { $t0 } from '@glimmer/vm';
 import { APPEND_OPCODES, UpdatingOpcode } from '../../opcodes';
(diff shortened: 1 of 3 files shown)


packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -237,27 +238,34 @@ APPEND_OPCODES.add(Op.DynamicAttr, (vm, { op1: _name, op2: _trusting, op3: _name
...
 
-    if (currentValue !== lastValue) {
-      attribute.update(currentValue, vm.env);
-      this.lastValue = currentValue;
-    }
+    this.updateRef = createComputeRef(() => {
+      let value = valueForRef(reference);
+
+      if (initialized === true) {
+        attribute.update(value, env);
```
rwjblue, line 259: FWIW, I still don't think this guarantees that the `DynamicAttribute` will actually update things. Some of the subclasses of `DynamicAttribute` have caches that do not check if the DOM value was changed out from under them.

For example:

https://github.com/glimmerjs/glimmer-vm/blob/4b697982032bdf6bc1fdbcb2e427f8d50fa07e1b/packages/%40glimmer/runtime/lib/vm/attributes/dynamic.ts#L117-L127

The reason the test added above passes is because `input.value` has a custom `DynamicAttribute` subclass (that doesn't do the last value caching that `DefaultDynamicProperty` does):

https://github.com/glimmerjs/glimmer-vm/blob/4b697982032bdf6bc1fdbcb2e427f8d50fa07e1b/packages/%40glimmer/runtime/lib/vm/attributes/dynamic.ts#L175-L182

rwjblue: Note: this almost certainly does address the reported issue, it just doesn't 100% fix the full category of issues. I'm fairly sure you could end up with this same issue with other element + attribute combinations that the user could mutate. We have custom overrides for a few common ones (`input.value`, `textarea.value`, `option.selected`), but that list of overrides is not exhaustive (e.g. video element has some local mutable attributes | properties that could have been changed by the user).

PR #1269, opened 2021-02-12
[BUGFIX] Fix {{get}} when used on non-objects

pzuraq:
Previously, `{{get}}` would not work on non-object values, even though
they may have had properties (e.g. `String.length`). This PR fixes this
and makes sure that `{{get}}` has the same behavior as directly
referencing the property in a template without using `{{get}}`.

Fixes https://github.com/emberjs/ember.js/issues/19400

### DIFF
--- a/packages/@glimmer/runtime/lib/helpers/get.ts
+++ b/packages/@glimmer/runtime/lib/helpers/get.ts
@@ -1,6 +1,7 @@
 import { getPath, setPath } from '@glimmer/global-context';
 import { VMArguments } from '@glimmer/interfaces';
 import { createComputeRef, valueForRef } from '@glimmer/reference';
+import { isDict } from '@glimmer/util';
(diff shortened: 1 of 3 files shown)


packages/@glimmer/runtime/lib/helpers/get.ts
```diff
@@ -88,21 +89,17 @@ export default internalHelper((args: VMArguments) => {
     () => {
       let source = valueForRef(sourceRef);
 
-      if (isObject(source)) {
+      if (isDict(source)) {
```
rwjblue, line 92: For the lazy, this is doing `source !== null && source !== undefined`

https://github.com/glimmerjs/glimmer-vm/blob/1a54cba8a6b4175cb2d5a617ff857b616cf6f0b7/packages/%40glimmer/util/lib/collections.ts#L13-L15

rwjblue (approved):

PR #1271, opened 2021-02-16
Upstream modifier tests

pzuraq:
Upstreams modifier tests from Ember, fixing some bugs with
`Object.assign` being used when it shouldn't have been. Also, converts
the types for modifiers to use `Element` instead of `SimpleElement`,
since the current assumption and behavior is that modifiers will only
run in interactive environments with a real DOM.

### DIFF
--- a/packages/@glimmer/integration-tests/test/managers/modifier-manager-test.ts
+++ b/packages/@glimmer/integration-tests/test/managers/modifier-manager-test.ts
@@ -1,0 +1,445 @@
+import { Arguments, ModifierManager, Owner } from '@glimmer/interfaces';
+import { RenderTest, test, jitSuite, tracked, defineComponent, trackedObj } from '../..';
+
+import { setModifierManager, modifierCapabilities } from '@glimmer/manager';
+import { getOwner, setOwner } from '@glimmer/owner';
+
+class CustomModifier {
+  static create(args: Arguments) {
+    return new this(getOwner(args)!, args);
+  }
+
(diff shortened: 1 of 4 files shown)


packages/@glimmer/interfaces/lib/managers/modifier.d.ts
```diff
@@ -23,7 +23,7 @@ export interface ModifierCapabilities extends Capabilities {
 export interface ModifierManager<ModifierStateBucket> {
   capabilities: ModifierCapabilities;
   createModifier(factory: unknown, args: Arguments): ModifierStateBucket;
-  installModifier(instance: ModifierStateBucket, element: SimpleElement, args: Arguments): void;
+  installModifier(instance: ModifierStateBucket, element: Element, args: Arguments): void;
```
rwjblue, line 26: FWIW, I personally consider this a non-breaking change. `SimpleElement` is a strict subset of the `Element` API.

packages/@glimmer/integration-tests/test/managers/modifier-manager-test.ts
```diff
@@ -0,0 +1,445 @@
...
+class CustomModifier {
+  static create(args: Arguments) {
+    return new this(getOwner(args)!, args);
+  }
+
+  constructor(owner: Owner, public args: Arguments) {
+    setOwner(this, owner);
+  }
+
+  element!: Element;
```
rwjblue, line 16: Should this be `declare element: Element`?

rwjblue (approved):

PR #1272, opened 2021-02-16
[BUGFIX] Fix reactivity of dynamic attributes

pzuraq:
The updates to use autotracking through the VM resulted in a bug in
dynamic attributes. Previously, the updating opcode for dynamic
attributes would run its update function based on tags/autotracking -
whenever any tag updated (e.g. a tracked value was changed) the
attribute would call its `update` function. The `update` function would
then check the new value against the current value of the attribute, and
update if they were different.

The bug was we introduced a value comparison in the updating opcode
itself, based on the last value of the property to set the attribute to.
This meant that if the property was updated to the same value twice in
a row, it would not call `update`, even if the attribute's value had
changed (e.g. because the user typed something).


rwjblue:
FYI - This is a backport of https://github.com/glimmerjs/glimmer-vm/pull/1268 into the branch used by ember-source@3.24.

PR #1273, opened 2021-02-17
[BUGFIX] Use built-in Object.assign if possible

pzuraq:
Use the built-in `Object.assign` if it exists, so we aren't using our
polyfill all the time.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -35,6 +35,10 @@
       }
     }
 
+    // Bootstrap internal `assign` so it doesn't get clobbered by the debug
+    // assertion below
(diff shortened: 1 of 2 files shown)


test/index.html
```diff
@@ -35,6 +35,10 @@
       }
     }
 
+    // Bootstrap internal `assign` so it doesn't get clobbered by the debug
+    // assertion below
+    require('@glimmer/util');
```
rwjblue, line 40: 😩

rwjblue (approved):

PR #1276, opened 2021-02-23
[BUGFIX] Do not eagerly consume modifier args during destruction

pzuraq:
Currently, in the 3.13 modifier capabilities version, we eagerly consume
the arguments that are passed to the modifier when it is destroying.
This can cause issues with backtracking rerender when the state isn't
actually used, and neither is the consumption. This change makes it so
that we don't eagerly consume these arguments during destruction in
general.

### DIFF
--- a/packages/@glimmer/integration-tests/test/managers/modifier-manager-test.ts
+++ b/packages/@glimmer/integration-tests/test/managers/modifier-manager-test.ts
@@ -234,6 +234,52 @@ abstract class ModifierManagerTest extends RenderTest {
       /You attempted to update `foo` on `.*`, but it had already been used previously in the same computation/
     );
   }
+
+  @test
+  'does not eagerly access arguments during destruction'(assert: Assert) {
+    class Foo extends CustomModifier {}
+
+    let foo = this.defineModifier(Foo);
+
+    let Main = defineComponent(
+      { foo },
+      '{{#if @state.show}}<h1 {{foo @state.bar baz=@state.baz}}>hello world</h1>{{/if}}'
+    );
+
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
     "typescript": "4.0.2"
   },
   "dependencies": {
-    "@handlebars/parser": "^2.0.0",
+    "@handlebars/parser": "~2.0.0",
```
rwjblue, line 46: Why do we need to pin this? This seems not good...

rwjblue (commented):
The other changes seem good, but we should revert the `package.json`/`yarn.lock` changes (they are unrelated to the rest of the PR)...

package.json
pzuraq: For some reason, the change was breaking. I'm not sure what the issue is exactly and have not had time to dig in

rwjblue: sure but since nothing else changed package.json / yarn.lock in this PR you probably can just delete the changes and `yarn` without issue

pzuraq: So I did that, tests failed, this change did result in `yarn.lock` changes and frozen lockfile threw

rwjblue: Running `yarn` on master does not require any `yarn.lock` changes, and `@handlebars/parser` is at 2.0.0. 

I've pushed 60f2dec72018e5e5fae7bd1701b5d9bd98434170 fixing this for you.

rwjblue (approved):

rwjblue:
I pinned @handlebars/parser in #1285 (and created #1286 to track down why it needed to happen), and rebased this.

PR #1277, opened 2021-02-28
[Feat]: equality operators

snewcomer:
RFC: https://github.com/emberjs/rfcs/pull/560

Related pkg - https://github.com/jmurphyau/ember-truth-helpers/blob/master/addon/helpers/equal.js

Some outstanding questions

- How does this behave in loose mode? We need to determine who will "win" in the case a user has ember-truth-helpers installed as well and how to throw a deprecation warning. 
- [change](https://github.com/emberjs/rfcs/pull/560#issuecomment-770002102) `neq` to `not-eq`.  Happy to open an another RFC or amendment to resolve.  This would also align us and allow us to deprecate ember-truth-helpers as well.
- Support for multiple params?


rwjblue:
> We need to determine who will "win" in the case a user has ember-truth-helpers installed as well 

The resolved helper **must** win (we can't assume that a resolved helper named `eq` _is_ from ember-truth-helpers and therefore make assumptions about its internal implementation), but I am not 100% we want to issue a deprecation in all cases where a helper is resolved from user land for one of these names.

> and how to throw a deprecation warning.

RE: when to issue a deprecation: I think we would only want to issue a deprecation if our implementation returns a **different** result than the resolved helper version.

RE: how to issue a deprecation: there is a nice deprecation system already built out, you import the deprecate function from global context and then invoke it. Ember has a hook into those deprecations and can augment the deprecation options (e.g. to add `for`/`since`/`until`/etc).

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/impl.ts
```diff
@@ -34,7 +34,8 @@ class KeywordImpl<
   constructor(
     protected keyword: S,
     type: KeywordType,
-    private delegate: KeywordDelegate<KeywordMatches[K], Param, Out>
+    private delegate: KeywordDelegate<KeywordMatches[K], Param, Out>,
+    private options?: { strictOnly: boolean }
```
snewcomer, line 38: Wasn't sure if this anonymous type was ok with you.

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/equality.ts
```diff
@@ -0,0 +1,67 @@
...
+  let {
+    args: { named, positional },
+  } = node;
+
+  if (named && !named.isEmpty()) {
+    return Err(generateSyntaxError(`(eq) does not take any named arguments`, node.loc));
+  }
+
+  return Ok(positional);
+}
```
rwjblue, line 19: I think we also need to assert if you use more than two positional arguments

https://github.com/emberjs/rfcs/blob/master/text/0560-add-equality-operators.md#eq

> {{eq}}
> Binary operation. Throws an error if not called with exactly two arguments. Equivalent of === This is identical to the eq helper in ember-truth-helpers

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/equality.ts
```diff
@@ -0,0 +1,67 @@
...
+  let {
+    args: { named, positional },
+  } = node;
+
+  if (named && !named.isEmpty()) {
+    return Err(generateSyntaxError(`(neq) does not take any named arguments`, node.loc));
+  }
+
+  return Ok(positional);
+}
```
rwjblue, line 49: Needs to assert RE: more than two positional params

> {{neq}}
> Binary operation. Throws an error if not called with exactly two arguments. Equivalent of !== This is identical to the not-eq helper in ember-truth-helpers, except for the name.

rwjblue (changes requested):

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
```diff
@@ -316,3 +316,29 @@ APPEND_OPCODES.add(Op.Log, (vm) => {
...
+
+APPEND_OPCODES.add(Op.Equal, (vm) => {
+  let { positional } = check(vm.stack.pop(), CheckArguments).capture();
+  let stack = vm.stack;
+
+  stack.push(
+    createComputeRef(() => {
+      let values = reifyPositional(positional);
+
+      return values.every((v) => v === values[0]);
```
rwjblue, line 328: I don't think we can use `.every` (IE11 doesn't have it), additionally we only need to support comparing two positionals.

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
```diff
@@ -316,3 +316,29 @@ APPEND_OPCODES.add(Op.Log, (vm) => {
...
+
+APPEND_OPCODES.add(Op.NotEqual, (vm) => {
+  let { positional } = check(vm.stack.pop(), CheckArguments).capture();
+  let stack = vm.stack;
+
+  stack.push(
+    createComputeRef(() => {
+      let values = reifyPositional(positional);
+
+      return values.some((v) => v !== values[0]);
```
rwjblue, line 341: I don't think we can use `.some` (IE11 doesn't have it), additionally we only need to support comparing two positionals.

rwjblue (changes requested):
I think we also need to update https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/vm/lib/opcodes.toml for these new opcodes.

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
btecu: IE does have `.some` since version 9:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#browser_compatibility

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
btecu: IE does have `.every` since version 9:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every#browser_compatibility

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/equality.ts
snewcomer: I can do that!  I built this in b/c I saw a comment from a certain someone ;). Happy to do whatever you feel best for this initial PR though.

https://github.com/emberjs/rfcs/pull/560#discussion_r385882319

NullVoxPopuli (approved):

MelSumner:
What else does this PR need to land? I would love love love to see this land. <3 Thank you all for working on it!

pzuraq:
@MelSumner part of the reason this has been blocked is because of v4 prep, we didn't want to land more changes in the VM before the last 3.x LTS was fully stable, in case we had to backport more fixes (there have been a few already).

Once that's done, I think the biggest issue is we need to actually change the way this is implemented so that in `loose` mode the resolved helpers win. That is a bit of a bigger change overall, because IIRC keywords always win in loose mode, so we need to implement these as "fallback helpers". I believe we were talking about doing _both_ implementations, because ultimately we do want these to be keywords, and then having the compiler choose which one to use depending on whether or not the template was in loose mode. Also, the outstanding comments need to be addressed.

MelSumner:
@pzuraq perfect, thanks for clarifying!

snewcomer:
> this is implemented so that in loose mode the resolved helpers win

It has been a while so I may have this wrong but I I think one workaround that we talked about and I put into this PR was a flag `strictModeOnly`.  Then in ember.js, I would implement the flags again (for loose mode?).  We can chat on Monday about it though!

packages/@glimmer/integration-tests/test/keywords/equality-test.ts
```diff
@@ -0,0 +1,106 @@
...
+  ['it works eq']() {
+    const AComponent = defineComponent({}, '{{eq 1 1}}');
+    this.renderComponent(AComponent);
+
+    this.assertHTML('true');
+  }
+
+  @test
+  ['it works multiple arguments eq']() {
+    const AComponent = defineComponent({}, '{{eq 1 1 1}}');
```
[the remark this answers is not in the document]
snewcomer: https://github.com/emberjs/rfcs/pull/560#discussion_r385882319

@rwjblue easy enough to support multiple I imagine.  What do you think?

NullVoxPopuli:
What's left here? It's been a couple months

snewcomer:
@NullVoxPopuli I believe nothing.  I'm excited about this but I'll need some help from either @chancancode and/or @rwjblue to review.  I've addressed the comments so far...

NullVoxPopuli:
I'll ping them

packages/@glimmer/runtime/lib/compiled/opcodes/expressions.ts
NullVoxPopuli: IE11 is no longer supported by the VM, so we don't need to care anyway :upside_down_face:

NullVoxPopuli:
does anything need to be done?
can this be merged?

snewcomer:
Sounds like we have a few steps.

1. Just to note, there is a leaf PR that would need to be merged - https://github.com/snewcomer/glimmer-vm/pull/3.  I could merge them all down to this branch if we are generally comfortable with the implementation.  Sounds like we might be but...
2. Pending wycats comments
3. And adding an error for > 2 arguments seems like a safe approach.  I'd be happy to add that to this PR unless somebody has a strong feeling.

snewcomer:
This has been updated with strict 2 arguments to `eq` and `neq`!

bertdeblock (commented):
Looking forward to this 👍

PR #1287, opened 2021-03-22
Adds a utility for parsing template locals

pzuraq:
This can be used by preprocessors and other tools that need to know
what locals could exist within the template

### DIFF
--- a/packages/@glimmer/syntax/lib/get-template-locals.ts
+++ b/packages/@glimmer/syntax/lib/get-template-locals.ts
@@ -1,0 +1,108 @@
+import { isKeyword } from './keywords';
+import { preprocess } from './parser/tokenizer-event-handlers';
+import traverse from './traversal/traverse';
+import * as ASTv1 from './v1/api';
+
+/**
+ * Gets the correct Token from the Node based on it's type
+ */
+function tokensFromType(node: ASTv1.Node, scopedTokens: string[]): string | void {
+  if (node.type === 'PathExpression') {
+    if (node.head.type === 'AtHead' || node.head.type === 'ThisHead') {
+      return;
+    }
+
+    const possbleToken = node.head.name;
+
+    if (scopedTokens.indexOf(possbleToken) === -1) {
+      return possbleToken;
+    }
+  } else if (node.type === 'ElementNode') {
+    const { tag } = node;
+
+    const char = tag.charAt(0);
+
+    if (char === ':' || char === '@') {
+      return;
+    }
+
+    if (tag.substr(0, 5) === 'this.') {
+      return;
+    }
+
+    if (scopedTokens.indexOf(tag) !== -1) {
+      return;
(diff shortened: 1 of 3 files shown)


packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Adds tokens to the tokensSet based on their node.type
+ */
+function addTokens(tokensSet: Set<string>, node: ASTv1.Node, scopedTokens: string[]) {
+  const maybeTokens = tokensFromType(node, scopedTokens);
+
+  (Array.isArray(maybeTokens) ? maybeTokens : [maybeTokens]).forEach((maybeToken) => {
+    if (
+      maybeToken !== undefined &&
+      !maybeToken.startsWith('@') &&
+      maybeToken.match(/^[$A-Z_][0-9A-Z_$]*$/i)
```
rwjblue, line 47: discussed, will remove

packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -0,0 +1,106 @@
...
+      exit({ blockParams }) {
+        blockParams.forEach(() => {
+          scopedTokens.pop();
+        });
+      },
+    },
+
+    All(node) {
+      addTokens(tokensSet, node, scopedTokens);
+    },
```
rwjblue, line 96: Can we move to using `PathExpression`? Are there any other ways to reference a "bound value"?

packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  });
+}
+
+/**
+ * Parses and traverses a given handlebars html template to extract all template locals
+ * referenced that could possible come from the praent scope. Can exclude known keywords
+ * optionally.
+ */
+export function getTemplateLocals(html: string, options?: { excludeKeywords: boolean }): string[] {
```
rwjblue, line 59: Can we default this?

packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -0,0 +1,106 @@
...
+        });
+      },
+    },
+
+    All(node) {
+      addTokens(tokensSet, node, scopedTokens);
+    },
+  });
+
+  let tokens = Array.from(tokensSet);
```
rwjblue, line 99: Can we rely on `Array.from`? It is not in IE11

packages/@glimmer/syntax/lib/get-template-locals.ts
pzuraq: inverted it so by default we exclude keywords

rwjblue (approved):

PR #1289, opened 2021-03-23
Exclude HTML elements from template locals by default

pzuraq:
Currently, all valid identifiers will be included in the output,
including HTML like identifiers. This can cause issues that are not
obvious from the get go, for instance:

```js
test('Supports the on modifier', async (assert) => {
  class MyComponent extends Component {
    static template = hbs`<button {{on "click" this.incrementCounter}}>Count: {{this.count}}</button>`;
    @tracked count = 0;

    @action
    incrementCounter() {
      this.count++;
    }
  }

  const element = document.getElementById('qunit-fixture')!;

  await renderComponent(MyComponent, element);
  assert.strictEqual(
    element.innerHTML,
    `<button>Count: 0</button>`,
    'the component was rendered'
  );

  const button = element.querySelector('button')!;
  button.click();

  await didRender();
  assert.strictEqual(
    element.innerHTML,
    `<button>Count: 1</button>`,
    'the component was rerendered'
  );
});
```

This test fails because `button` is technically an in-scope variable,
declared _after_ `renderComponent` is called. Using the template
transform, it thinks that `button` really exists and is a component
definition, and throws a ReferenceError when it attempts to access it
because it hasn't been defined yet.

This change would allow us to ignore any identifier that is:

1. Only used in angle bracket invocation
2. Has no path segments
3. Is all lower case

Preventing these collisions from happening with common HTML elements
such as `button`, `span`, `div`, etc. This also follows JSX's rules for
determining if a [value in scope is a component](https://reactjs.org/docs/jsx-in-depth.html#html-tags-vs.-react-components),
so at the least this strategy is already used in a real world framework
without much confusion.

### DIFF
--- a/packages/@glimmer/syntax/lib/get-template-locals.ts
+++ b/packages/@glimmer/syntax/lib/get-template-locals.ts
@@ -3,10 +3,19 @@ import { preprocess } from './parser/tokenizer-event-handlers';
 import traverse from './traversal/traverse';
 import * as ASTv1 from './v1/api';
 
+interface GetTemplateLocalsOptions {
+  includeKeywords?: boolean;
+  includeHtmlElements?: boolean;
+}
+
 /**
  * Gets the correct Token from the Node based on it's type
  */
-function tokensFromType(node: ASTv1.Node, scopedTokens: string[]): string | void {
+function tokensFromType(
+  node: ASTv1.Node,
+  scopedTokens: string[],
+  options: GetTemplateLocalsOptions
+): string | void {
   if (node.type === 'PathExpression') {
     if (node.head.type === 'AtHead' || node.head.type === 'ThisHead') {
       return;
@@ -26,6 +35,10 @@ function tokensFromType(node: ASTv1.Node, scopedTokens: string[]): string | void
       return;
     }
 
+    if (!options.includeHtmlElements && tag.indexOf('.') === -1 && tag.toLowerCase() === tag) {
+      return;
+    }
+
     if (tag.substr(0, 5) === 'this.') {
       return;
     }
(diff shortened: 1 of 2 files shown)


packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -56,7 +74,7 @@ function addTokens(tokensSet: Set<string>, node: ASTv1.Node, scopedTokens: strin
  * referenced that could possible come from the praent scope. Can exclude known keywords
  * optionally.
  */
-export function getTemplateLocals(html: string, options?: { includeKeywords: boolean }): string[] {
+export function getTemplateLocals(html: string, options: GetTemplateLocalsOptions = {}): string[] {
```
rwjblue, line 77: I think we should default both of the options, something like:

```suggestion
export function getTemplateLocals(html: string, options: GetTemplateLocalsOptions = { includeKeywords: false, includeHtmlElements: false }): string[] {
```

I believe this will be the common case.

packages/@glimmer/syntax/lib/get-template-locals.ts
```diff
@@ -26,6 +35,10 @@ function tokensFromType(node: ASTv1.Node, scopedTokens: string[]): string | void
       return;
     }
 
+    if (!options.includeHtmlElements && tag.indexOf('.') === -1 && tag.toLowerCase() === tag) {
```
rwjblue, line 38: Just to confirm, block param cases would have returned already right? (If not, this check isn't sufficient)...

Can you add a test for:

```hbs
{{#let (component @thing) as |button|}}
  <button></button>
{{/let}}
```

pzuraq: Yes, this additional check is only for top level items

rwjblue (approved):

PR #1293, opened 2021-04-01
Fix nested calls to helpers in dynamic helpers

pzuraq:
Currently, nested calls to helpers in dynamic helpers cause problems
because the nested helper overwrites the register that we were using to
pass along the helper definition. This is actually a worse way of doing
something that the `Dup` opcode is designed to do anyways, so I updated
to do that, following the same pattern in DynamicModifiers. Added tests
for both dynamic helpers and modifiers with nested helpers.

Original issue: https://github.com/pzuraq/ember-could-get-used-to-this/pull/41

### DIFF
--- a/packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/vm.ts
+++ b/packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/vm.ts
@@ -1,4 +1,4 @@
-import { $v0 } from '@glimmer/vm';
+import { $fp, $v0 } from '@glimmer/vm';
 import {
   Option,
   Op,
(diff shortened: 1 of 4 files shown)


NullVoxPopuli (approved):

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/vm.ts
```diff
@@ -79,13 +79,13 @@ export function CallDynamic(
   named: WireFormat.Core.Hash,
   append?: () => void
 ): void {
-  op(Op.Load, $v0);
   op(MachineOp.PushFrame);
   SimpleArgs(op, positional, named, false);
-  op(Op.DynamicHelper, $v0);
+  op(Op.Dup, $fp, 1);
```
NullVoxPopuli, line 84: is there a reference somewhere to see what each opcode's purpose is?

rwjblue: I think https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/vm/lib/opcodes.toml is probably the best place.

rwjblue (approved):

PR #1295, opened 2021-04-13
Deprecate `@foo={{helper}}`

chancancode:
Per [RFC 496](https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#3-no-implicit-invocation-of-argument-less-helpers), this commit deprecates the implicit invocation of argument-less helpers and requires explicit parenthesis if:

1. Non-strict mode, AND
2. `helper` is a free variable (not `this.helper`, `@helper` or
   a local variable), AND
3. No arguments are provided to the helper, AND
4. It's in a component invocation's named argument position (not
   `<div id={{helper}}>` or `<Foo bar={{helper}}>`), AND
5. Not parenthesized (not `@foo={{(helper)}}`), AND
6. Not interpolated (not `@foo="{{helper}}"`).

The cases in real apps where all of the above are true should be quite rare, as most helpers take at least one argument.

This is probably not the most efficient/minimal changes to the compiler infrastructure, but I chose to optimize for making the change easier to remove/rollback (i.e. adding new nodes instead of changing existing ones) as we won't be needing this for very long.

### DIFF
--- a/packages/@glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts
+++ b/packages/@glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts
@@ -1,0 +1,95 @@
+import { RenderTest, jitSuite, test, defineSimpleHelper } from '../..';
+
+class ArgumentLessHelperParenLessInvokeTest extends RenderTest {
+  static suiteName = 'argument-less helper paren-less invoke';
+
+  @test
+  'invoking an argument-less helper without parens in named argument position is deprecated'(
+    assert: Assert
+  ) {
+    this.registerHelper('is-string', ([value]: readonly unknown[]) => typeof value === 'string');
+
+    this.registerHelper('foo', () => 'Hello, world!');
+    this.registerComponent('TemplateOnly', 'Bar', '[{{is-string @content}}][{{@content}}]');
+
+    this.render('<Bar @content={{foo}} />', { foo: 'Not it!' });
+    this.assertHTML('[true][Hello, world!]');
+    this.assertStableRerender();
+
+    assert.validateDeprecations(
+      new RegExp(
+        /The `foo` helper was used in the `\(unknown template module\)` template as /.source +
+          /`@content={{foo}}`\. This is ambigious between wanting the `@content` argument /.source +
+          /to be the `foo` helper itself, or the result of invoking the `foo` helper /.source +
+          /\(current behavior\)\. This implicit invocation behavior has been deprecated\./.source
+      )
+    );
+  }
+
+  @test
+  'invoking an argument-less helper with parens in named argument position is not deprecated'() {
+    this.registerHelper('is-string', ([value]: readonly unknown[]) => typeof value === 'string');
+
+    this.registerHelper('foo', () => 'Hello, world!');
+    this.registerComponent('TemplateOnly', 'Bar', '[{{is-string @content}}][{{@content}}]');
+
+    this.render('<Bar @content={{(foo)}} />');
+    this.assertHTML('[true][Hello, world!]');
+    this.assertStableRerender();
+  }
+
+  @test
+  'invoking an argument-less helper with quotes in named argument position is not deprecated'() {
+    this.registerHelper('is-string', ([value]: readonly unknown[]) => typeof value === 'string');
+
+    this.registerHelper('foo', () => 'Hello, world!');
+    this.registerComponent('TemplateOnly', 'Bar', '[{{is-string @content}}][{{@content}}]');
+
+    this.render('<Bar @content="{{foo}}" />');
+    this.assertHTML('[true][Hello, world!]');
+    this.assertStableRerender();
+  }
+
+  @test
+  'passing a local helper in named argument position is not deprecated'() {
+    this.registerHelper('is-string', ([value]: readonly unknown[]) => typeof value === 'string');
+
+    const foo = defineSimpleHelper(() => 'Hello, world!');
+
+    this.registerComponent('TemplateOnly', 'Bar', '[{{is-string @content}}][{{@content}}]');
+
+    this.render('{{#let this.foo as |foo|}}<Bar @content={{foo}} />{{/let}}', { foo });
+    this.assertHTML('[false][Hello, world!]');
+    this.assertStableRerender();
+  }
+
+  @test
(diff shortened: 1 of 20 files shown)


lifeart:
wondering, should it be template-lint rule too

packages/@glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const foo = defineSimpleHelper(() => 'Hello, world!');
+
+    this.registerComponent('TemplateOnly', 'Bar', '[{{is-string @content}}][{{@content}}]');
+
+    this.render('{{#let this.foo as |foo|}}<Bar @content={{(foo)}} />{{/let}}', { foo });
+    this.assertHTML('[true][Hello, world!]');
+    this.assertStableRerender();
+  }
+
+  // TODO: this should work, but doesn't
```
chancancode, line 79: This seems to be a bug. It calls `Function.prototype.toString()` on the helper instead of invoking it.

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
```diff
@@ -219,6 +231,7 @@ export namespace Expressions {
     | GetPathFreeAsComponentOrHelperHeadOrThisFallback
     | GetPathFreeAsComponentOrHelperHead
     | GetPathFreeAsHelperHeadOrThisFallback
+    | GetPathFreeAsDeprecatedHelperHeadOrThisFallback
```
chancancode, line 234: Some of the `GetPathFree*` are not actually possible/used. Take `GetPathFreeAsHelperHeadOrThisFallback` for instance – if there is a path (i.e. `foo={{bar.baz}}` then `bar` must be a fallback for `this.bar` and cannot be a helper.

However, the current code, [specifically here](https://github.com/glimmerjs/glimmer-vm/blob/master/packages/@glimmer/compiler/lib/passes/2-encoding/expressions.ts#L93-L95) seems to require/assuming symmetry between `GetFree*` and `GetPathFree*`. Deleting the unused stuff from here seems to cause type issues.

In the end, I ended up abusing/hijacking the "path" portion of `GetPathFreeAsDeprecatedHelperHeadOrThisFallback` to save the argument name for the deprecation message (adding a `string` element to the tuple also breaks types for similar reasons).

chancancode:
> wondering, should it be template-lint rule too

It's not possible to tell this statically since you don't know whether it is a helper or fallback for `this.foo`

rwjblue:
> > wondering, should it be template-lint rule too
> 
> It's not possible to tell this statically since you don't know whether it is a helper or fallback for this.foo

I think we _could_ make it a rule, you just can only use the rule when you are already using `no-implicit-this` (which is in the recommended set now FWIW). 🤔

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
```diff
@@ -79,6 +79,8 @@ export const enum SexpOpcodes {
   GetFreeAsComponentOrHelperHead = 35,
   // a helper or `this` fallback `attr={{x}}`
   GetFreeAsHelperHeadOrThisFallback = 36,
+  // a helper or `this` fallback (deprecated) `@arg={{x}}`
+  GetFreeAsDeprecatedHelperHeadOrThisFallback = 99,
```
rwjblue, line 83: Why 99?

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
rwjblue: > if there is a path (i.e. foo={{bar.baz}} then bar must be a fallback for this.bar and cannot be a helper

Hmm, is that correct? I thought this should work (as of ember-source@3.25+), and is part of "everything is a value" conceptually. So both having a yielded block param for `bar` with a property that is `baz` that contains a `Ember.Helper.helper(() => {})` and having the same thing at `this.bar` **should** invoke that helper. 🤔

chancancode: Sure, but that would also not go through this path as bar is a local variable.

This is only used when bar is a “free variable” (what was called ambient scope in strict mode) _but_ we need to lookup bar as a resolved helper, and failing that, fallback to this.bar, which should never happen as far as I can tell.

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
chancancode: This is temporary and will be removed again soon, so seems somehow safer to use a number that wouldn’t be used to mean something else important anytime soon (though if the WF and compiler version doesn’t match you are going to have a bad time anyway)

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
pzuraq: This is correct I believe, the path stuff for those `GetFree` opcodes was from before the implementation of strict mode, when we were still thinking of imported values as free variables. Now, imported values are in-scope values, which means that they never become any kind of `GetFree`.

packages/@glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts
chancancode: @pzuraq ^ (not super related to this PR, but the test caught the problem)

packages/@glimmer/interfaces/lib/compile/wire-format.d.ts
wycats: @pzuraq is correct

packages/@glimmer/integration-tests/test/syntax/argument-less-helper-paren-less-invoke-test.ts
wycats: Is the issue fixed or do we need to open an issue to track it?

chancancode: needs more work, I can open an issue

wycats (approved):
Looks great to me!

PR #1296, opened 2021-04-20
Correctly handle <:inverse> named block

wagenet:


chancancode:
We will need to make sure it works with the checks for duplicate blocks and stuff (it normally checks that you can’t have two blocks with the same name, but we need to make sure you can’t have both else and inverse here).

wagenet:
@chancancode it did not work, but I added a new assertion.

chancancode (approved):

PR #1298, opened 2021-04-22
[BUGFIX] Ensure `hash` properties are reified lazily

pzuraq:
`(hash)` previously had two different types of behavior, one where it
was used as a reference and could get individual properties lazily, and
one where it was fully reified and exposed in JS. The lazy nature of
hash was very useful as it prevented having to realize the full cost of
each reference immediately, and it also meant that some values could
be accessed in conditionals and run assertions just in time. For
example, if someone was using a component that was loaded lazily via
an engine, you could render the component on a hash within a conditional
and it would not cause an error (because the component did not exist):

```hbs
<SomeComponent as |api|>
  <api.eagerComponent/>

  {{#if this.displayLazy}}
    <api.lazyComponent/>
  {{/if}}
</SomeComponent>
```

This PR restores this ability by having `(hash)` return a proxy rather
than an object, which allows us to access each individual value without
reifying the references immediately. For backwards compatibility, we
also allow users to set values on the proxy, so long as they are not
setting any of the values defined on the proxy initially, since those
are read-only references.

### DIFF
--- a/packages/@glimmer/integration-tests/test/helpers/hash-test.ts
+++ b/packages/@glimmer/integration-tests/test/helpers/hash-test.ts
@@ -156,6 +156,70 @@ class HashTest extends RenderTest {
 
     this.assertHTML('Chad Hietala');
   }
+
+  @test
+  'individual hash values are accessed lazily'(assert: Assert) {
+    class FooBar extends GlimmerishComponent {
+      firstName = 'Godfrey';
+
+      get lastName() {
+        assert.ok(false, 'lastName was accessed');
+
+        return;
+      }
+    }
+
+    this.registerComponent(
+      'Glimmer',
+      'FooBar',
+      `{{yield (hash firstName=@firstName lastName=this.lastName)}}`,
+      FooBar
+    );
+
+    this.render(`<FooBar @firstName="Godfrey" as |values|>{{values.firstName}}</FooBar>`);
+
+    this.assertHTML('Godfrey');
+    this.assertStableRerender();
+  }
+
+  @test
+  'defined hash keys cannot be updated'(assert: Assert) {
+    class FooBar extends GlimmerishComponent {
+      constructor(owner: object, args: { hash: Record<string, unknown> }) {
+        super(owner, args);
+        args.hash.firstName = 'Chad';
+      }
+    }
+
+    this.registerComponent('Glimmer', 'FooBar', `{{yield @hash}}`, FooBar);
+
+    assert.throws(() => {
+      this.render(`
+        <FooBar @hash={{hash firstName="Godfrey"}} as |values|>
+          {{values.firstName}} {{values.lastName}}
+        </FooBar>
+      `);
+    }, /You attempted to set the "firstName" value on an object generated using the \(hash\) helper|Assignment to read-only properties is not allowed in strict mode/);
+  }
+
+  @test
+  'undefined hash keys can be updated'() {
+    class FooBar extends GlimmerishComponent {
+      constructor(owner: object, args: { hash: Record<string, unknown> }) {
+        super(owner, args);
+        args.hash.lastName = 'Chan';
+      }
+    }
+
+    this.registerComponent('Glimmer', 'FooBar', `{{yield @hash}}`, FooBar);
+
+    this.render(
+      `<FooBar @hash={{hash firstName="Godfrey"}} as |values|>{{values.firstName}} {{values.lastName}}</FooBar>`
+    );
+
+    this.assertHTML('Godfrey Chan');
+    this.assertStableRerender();
+  }
 }
 
 jitSuite(HashTest);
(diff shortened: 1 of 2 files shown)


simonihmig:
> so long as they are not setting any of the values defined on the proxy initially, since those are read-only references.

The assertion added here is causing an [addon test](https://github.com/kaliber5/ember-bootstrap/blob/master/tests/integration/components/bs-form-test.js#L410-L426) to fail, when using Ember canary where this change here just landed. 

I don't really understand the reasoning here, at least from a user's perspective. Why shouldn't I be able to change a property of the hash even when it existed before? 

In my mental model using `{{hash foo="bar"}}` should be the same as defining `{ foo: 'bar' }}` on the class and passing this in the template. The latter allows mutating the hash in every possible way.

Even if there is some good reason to do this, this behavior must be documented, and changing it feels very much like a breaking change, which the failing test basically confirms. Not?

pzuraq:
@simonihmig it would be very, very difficult to make this behavior work the way you described. It's really non-trivial, and introduces a lot of issues with timing and syncing the local state of the hash with your changes, to the state of the incoming references that defined the hash in the first place.

We wanted to avoid these extra complexities, which would likely introduce a host of their own small bugs and inconsistencies in our opinion. The hope was that users were treating `hash` as derived state, the way it was meant to be used, and were not using it as a store mutable state at all. If this is a common use case, then we may need to update the behavior to help that, but we really would rather avoid it.

If we do make the change, it is likely we will immediately deprecate the ability to do this, since it'll be much easier and simpler to maintain a `hash` which does not have this behavior.

simonihmig:
I see, thanks for explaining.

So practically I myself am not really affected by this, as we don't have used patterns like this. 

But this case was added to our test suite because of a real use case by @jelhan IIRC, right?

pzuraq:
FWIW, if this is failing tests already in canary then I think we should at least re-evaluate our original decision. I'm going to discuss it with folks next week, some are currently out, we'll see how that goes before we do any backports or anything like that.

jelhan:
My mental modal about `{{hash}}` helper was that it returned an object with the named arguments as properties. Basically I assumed that it could be implemented in user-land like this:

```js
// app/helpers/hash.js

import { helper } from '@ember/component/helper';

function hash(_, namedArgs) {
  return { ...namedArgs };
}

export default helper(hash);
```

I have used it whenever it improved readability to construct an object in the template rather than in the JavaScript file. Basically I considered that either constructing the object in a backing JavaScript class or using `{{hash}}` template helper has the same result.

```js
// app/components/my-component.js

import Component from '@glimmer/component';

export default class MyComponent extends Component {
  get formModel() {
    return {};
  }
}
```

```hbs
<PostForm @model={{this.formModel}} />
```

vs

```hbs
<PostForm @model={{hash}} />
```

I often preferred the `{{hash}}` helper because it makes it often easier to reason about the code.

My mental modal about `{{hash}}` helper and derived state was similar to [`@localCopy` decorator](https://github.com/pzuraq/tracked-toolbox#localcopy) provided by `tracked-toolbox`. The local copy could be modified but would be thrown away if one of the arguments change.

Speaking in the past here because this issue and the discussion changed my mental model. I wasn't aware that `{{hash}}` helper evaluates arguments lazy. I wasn't aware that the returned object should not be mutated. Why is it not frozen in that case? I'm confused. My mental model around `{{hash}}` helper failed apart. Not sure if I'm the only one, who feels like this.

The impact on the apps I'm working on may not be that big though. Often the `{{hash}}` is wrapped inside an [Ember Changeset](https://github.com/poteto/ember-changeset) if used as a model for a form. That should mitigate the affect as the object returned by `{{hash}}` helper is not modified. But I'm sure there are some cases, which would be broken.

It feels like a breaking change for me. Especially as the [API documentation](https://api.emberjs.com/ember/3.26/classes/Ember.Templates.helpers/methods/hash?anchor=hash) does not mention that mutating the returned object is discouraged. To be honest it sounds for more like a chance, which deserves a RFC.

lifeart:
With set on hash it should throw error:

image


https://ember-twiddle.com/7ede4a6b8d3ce50698acc9d7fe57ebaa?openFiles=helpers.h%5C.js%2C

(replace 

 		return this.object = this.object || EmberObject.extend({...hash}).create();

with

 		return this.object = this.object || hash

pzuraq:
@jelhan to be fair, we were also not aware that `hash` worked this way when we made the original refactors/changes to references 😅 this was the issue with the previous reference system, every reference type could customize its own behavior, so you had weird edge cases where one specific type of reference would do something completely differently than every other type. So that was just a behavior we didn't know existed.

I think it's likely we'll have to keep the old behavior since things are failing, but in general I think you should reconsider the strategy with using `hash` as a mutable store of value. At the least, the following are all true:

1. `hash` always returned a _normal_ object, not a TrackedObject or anything like that.
2. So, the only way to update the state on `hash` and make those updates _reactive_ is by using `Ember.get` to read values on it, and using `Ember.set` to write values to it.
3. In the (distant) future, `Ember.get` and `Ember.set` will likely be deprecated. At the least, they are not considered the main way to access/update state in Octane and beyond.

Given that, I think this pattern will have to be re-evaluated either way. If this pattern were to continue in the future, I think we'd have to have `hash` return a TrackedObject (and introduce TrackedObject to Ember directly).

PR #1304, opened 2021-05-11
Remove unnecessary property descriptor assertion from hash

pzuraq:
Removes an unnecessary property descriptor assertion, hash should work
like a normal object (unlike args proxies) and so it should be allowed
to read the property descriptors on it.


lifeart:
may resolve https://github.com/emberjs/ember.js/issues/19130

PR #1308, opened 2021-05-18
Allow all (hash) keys to be set

pzuraq:
Enables setting all keys on a `(hash)`. We do this by creating a
side-effecting cache which syncs the value of the key to the local state
whenever it is first accessed, and checking that cache on all subsequent
accesses. Whenever the upstream value has dirtied, the cache will rerun
and overwrite the local state of the hash.

When setting values on the hash with `Ember.get` and `Ember.set`, these
functions will get and set the tags associated with the _proxy_ for
those keys. This is why `customTagFor` has been updated to get and
combine the tag for the proxy as well, for the time being.


packages/@glimmer/integration-tests/test/helpers/hash-test.ts
```diff
@@ -193,13 +193,36 @@ class HashTest extends RenderTest {
...
+        super(owner, args);
+        args.hash.name = 'Chad';
+      }
+    }
+
+    this.registerComponent('Glimmer', 'FooBar', `{{yield @hash}}`, FooBar);
+
+    this.render(`<FooBar @hash={{hash name=this.name}} as |values|>{{values.name}}</FooBar>`, {
+      firstName: 'Godfrey',
+    });
```
rwjblue, line 217: Where does `this.name` (passed to the hash invocation) come from? The value passed as the second arg of `this.render` is `firstName`, should this be:

```suggestion
    this.render(`<FooBar @hash={{hash name=this.firstName}} as |values|>{{values.name}}</FooBar>`, {
      firstName: 'Godfrey',
    });
```

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
```diff
@@ -193,13 +193,36 @@ class HashTest extends RenderTest {
...
+
+    this.assertHTML('Chad');
+    this.assertStableRerender();
+
+    this.rerender({ name: 'Tom' });
```
rwjblue, line 222: The initial `this.render` call passes a `firstName` field but this one passes a `name` field. Can you either explain why that is intentional, or sync them?

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
```diff
@@ -193,13 +193,36 @@ class HashTest extends RenderTest {
...
+    this.render(
+      `<FooBar @hash={{hash firstName="Godfrey" lastName="Hietala"}} as |values|>{{values.firstName}} {{values.lastName}}</FooBar>`
+    );
+
+    this.assertHTML('Chad Hietala');
+    this.assertStableRerender();
+  }
+
+  @test
+  'defined hash keys are reset whenever the upstream changes'() {
```
rwjblue, line 205: I think the prior behavior was that all manually set hash values would be reset if *any* were reset (because the hash itself was unstable), right?

Could you add another test for that?

rwjblue (changes requested):

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
pzuraq: We cannot do this, unfortunately. This directly contradicts with our goal of making each individual key lazy. The way that was accomplished previously was the fact that the value of a reference was different than the paths of references in templates.

Basically this would put us back to square one.

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
pzuraq: Ah yes, this was a mistake

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
pzuraq: yes, the original render call was a mistake, will update

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
pzuraq: Discussed it further with @rwjblue, what we can do is eagerly entangle anything the first time a value is set on the hash, and then reset everything whenever anything changes. It's a bit annoying, but it's more accurate compared to the original behavior.

rwjblue (approved):

packages/@glimmer/integration-tests/test/helpers/hash-test.ts
pzuraq: Discussed further again, that strategy unfortunately does not work in IE11, so we're back to the original new semantics

PR #1310, opened 2021-05-19
Adds EffectsManager

pzuraq:
This PR finishes up the autotracking refactors that were done
previously, making after-render effects that used to be handled by the
render transaction handled instead by an EffectsManager. The
EffectsManager essentially manages a number of caches, which it keeps
in lists internally. Every time a render completes, these queues are
scheduled to run, with every cache in the queue being checked in order
to see if something has changed. If it has, the effect runs its update.

The biggest change with this refactor is that the
`didCreate`/`didUpdate` hooks used by classic components will now
interleave with modifiers. Previously, they always ran _before_ all
modifiers, regardless of whether the modifiers were children of a given
component. In theory, this shouldn't be an issue, but if it is we can
separate out the component hooks into a separate queue and restore the
previous ordering.

Supercedes #1061


NullVoxPopuli:
What's left?
I'll very happily take this across the finish line

PR #1313, opened 2021-05-21
[Feat]: element keyword

snewcomer:
WIP

Ref polyfill - 
https://github.com/tildeio/ember-element-helper

TODO

- [ ] tests passing
- [ ] curry test
- [ ] test failure with named args
- [ ] verify debug name for DynamicElementManager
- [ ] address limitations and what we should do here - https://github.com/tildeio/ember-element-helper#limitations

### DIFF
--- a/packages/@glimmer/integration-tests/test/keywords/dynamic-element-test.ts
+++ b/packages/@glimmer/integration-tests/test/keywords/dynamic-element-test.ts
@@ -1,0 +1,38 @@
+import { RenderTest, test, jitSuite, syntaxErrorFor, defineComponent, trackedObj } from '../..';
+
+class DynamicElementTest extends RenderTest {
+  static suiteName = '{{element}} keyword';
+
+  @test
+  ['it works element']() {
+    const AComponent = defineComponent({}, '{{#let (element "div") as |Tag|}}<Tag></Tag>{{/let}}');
+    this.renderComponent(AComponent);
+
+    this.assertHTML('<div></div>');
+  }
+
+  @test
+  ['correctly renders when values update element']() {
+    let args = trackedObj({ foo: 'span' });
+
+    const AComponent = defineComponent({}, '{{element @foo}}');
+    this.renderComponent(AComponent, args);
+
+    this.assertHTML('<span></span>');
+
+    args.foo = 'div';
+    this.rerender({ foo: 'div' });
+
+    this.assertHTML('<div></div>');
+  }
+
+  @test
+  ['it errors']() {
+    this.assert.throws(() => {
+      const AComponent = defineComponent({}, '{{element "div" 1}}');
+      this.renderComponent(AComponent);
+    }, syntaxErrorFor('(element) only takes one positional argument - the element tag name', '{{element "div" 1}}', 'an unknown module', 1, 0));
+  }
+}
+
+jitSuite(DynamicElementTest);
--- a/packages/@glimmer/runtime/lib/references/dynamic-element.ts
+++ b/packages/@glimmer/runtime/lib/references/dynamic-element.ts
@@ -1,0 +1,70 @@
+import { InternalComponentCapabilities, InternalComponentManager } from '@glimmer/interfaces';
+import { CurriedType, Dict, Maybe } from '@glimmer/interfaces';
+import { createComputeRef, NULL_REFERENCE, Reference, valueForRef } from '@glimmer/reference';
+import { setInternalComponentManager } from '@glimmer/manager';
+import { curry } from '../curried-value';
+
+const CAPABILITIES: InternalComponentCapabilities = {
+  dynamicLayout: false,
+  dynamicTag: false,
+  prepareArgs: false,
+  createArgs: false,
+  attributeHook: false,
+  elementHook: false,
+  createCaller: false,
+  dynamicScope: false,
+  updateHook: false,
+  createInstance: false,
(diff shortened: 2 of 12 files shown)


NullVoxPopuli:
@snewcomer you still motivated to finish this? (how close is it? should we find help?)

snewcomer:
@NullVoxPopuli Would love to!  I just need some confidence that the work I get done will actually be reviewed and merged though. e.g. https://github.com/glimmerjs/glimmer-vm/pull/1277

packages/@glimmer/integration-tests/test/keywords/dynamic-element-test.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.assertHTML('<div></div>');
+  }
+
+  @test
+  ['it errors']() {
+    this.assert.throws(() => {
+      const AComponent = defineComponent({}, '{{element "div" 1}}');
+      this.renderComponent(AComponent);
+    }, syntaxErrorFor('(element) only takes one positional argument - the element tag name', '{{element "div" 1}}', 'an unknown module', 1, 0));
+  }
```
chancancode, line 35: lets have one for named args too, since I think we want to reserve the space for that to mean attributes

chancancode: also `{{element true}}` etc errors

packages/@glimmer/integration-tests/test/keywords/dynamic-element-test.ts
```diff
@@ -0,0 +1,38 @@
...
+}
```
chancancode, line 36: Let's add some tests for:

```hbs
<Tag>
  Some dynamic content here.
  When "Tag" changes, the content here get torn down and re-rendered.
  You can probably assert that with a helper that does `count++` or something similar.
</Tag>
```
```hbs
According to https://github.com/emberjs/rfcs/pull/620/files#r418378655

{{#let (element "") as |Tag|}}
  <Tag>Hello</Tag>
{{/let}}

behaves like

Hello (no wrapping tag)
```

```hbs
According to https://github.com/emberjs/rfcs/pull/620/files#r418379071

{{#let (element @tagName) as |Tag|}}
  <Tag>Hello</Tag>
{{/let}}

Apparently, when @tagName changes between "div" and null/undefined, the above behaves like:

{{#if @tagName}}
  <div>Hello</div>
{{/if}}
```

packages/@glimmer/runtime/lib/references/dynamic-element.ts
```diff
@@ -0,0 +1,70 @@
...
+  constructor(public tagName: string) {}
+}
+
+setInternalComponentManager(DYNAMIC_ELEMENT_MANAGER, DynamicElementComponentDefinition.prototype);
+
+export default function createDynamicElementRef(tagName: Reference) {
+  let lastValue: Maybe<Dict> | string, curriedDefinition: object | string | null;
+
+  return createComputeRef(() => {
+    let value = valueForRef(tagName) as string;
```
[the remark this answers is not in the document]
chancancode: (Therefore also tests that makes this dynamic instead of a literal)

packages/@glimmer/runtime/lib/references/dynamic-element.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    if (value === lastValue) {
+      return curriedDefinition;
+    }
```
chancancode, line 57: I think the code will be easier to write if you do


```ts
let lastValue = null;
let component = null;

return createComputeRef(() => {
  // ...

  if (value === lastValue) {
    return component;
  }

  lastValue = value;
  component = ...;
  
  return component;
});
```

chancancode: I also think that we shouldn't need the "curry" stuff here anymore. If I understand correctly, things are supposed to Just Work™ when you do `<Tag>` where `Tag` is a component. So I don't think we need to go out of our way to "curry nothing" here.

On a related note I think we also try to hide what the return type is since it can leak into user land JS code an ultimately it will be depended on by app code _somehow_ (in this case the `DynamicElementComponentDefinition` class and the fact that it has a `tagName` property on it will accidentally become public api).

So we can do this dance (or there may be a similar pattern elsewhere in glimmer-vm already?):


```ts
// possibly, we can do less work between DEBUG and production, but I think it doesn't really matter
// since this shouldn't be a super common feature?

// not exported, not public
class DynamicElement {
  constructor(readonly tagName: string) {}
}

const DYNAMIC_ELEMENTS = new WeakMap<object, DynamicElement>();

function makeDynamicElement(tagName: string | null): object | null {
  if (tagName === null) {
    return null;
  }

  let component = new DynamicElement(tagName);

  let opaqueValue = Object.freeze({
    // just so you get something nice to look at when logging to the console
    toString() {
      return `(element "${tagName}")`;
    }
  });
  
  setInternalComponentManager(DYNAMIC_ELEMENT_MANAGER, opaqueValue);

  DYNAMIC_ELEMENTS.set(opaqueValue, component);

  return opaqueValue;
}
```

chancancode:
It seems like this PR implemented the part where `(element ...)` returns a component, but haven't actually implemented the "when we render that component it does the right thing" part. Is that right/is that the part you need help with?

PR #1314, opened 2021-05-22
Consolidate references and tags

pzuraq:


NullVoxPopuli:
@pzuraq what's the goal of this PR?

How could someone best take over?

pzuraq:
@NullVoxPopuli the goal is to consolidate the tracking primitives so that we only have one set of primitives. Previously, we had the system of tags where you manually tracked tags. After a while, we realized we could represent everything pretty much as just Sources and Caches, where Sources represent root state and Caches represent derived state. Furthermore, these can both be implemented by the same underlying class and use the same interface (`getValue` and `setValue` can be called on both), meaning that they can be used interchangeably.

With this refactor, references in templates just become caches or sources, and everywhere we're currently using tags manually can be simplified a lot. This PR does that simplification in Glimmer VM, though it would still need to be done in Ember to integrate.

PR #1318, opened 2021-05-27
Ensure hash works with custom tags

pzuraq:
We need to make sure references exist before we attempt to get
their value and track them.

Fixes: https://github.com/emberjs/ember.js/issues/19562


pzuraq:
Integration PR in Ember w/ tests: https://github.com/emberjs/ember.js/pull/19565

rwjblue (approved):

Issue #950, opened 2019-06-25
Long inserts optimization


lifeart:
@wycats @chancancode 

Case: If we have long list, and append more nodes into it, rendering struggling.

Real life most used case: 

	let users = Array(20);
	
	let moreUsers = Array(20);
	
	let updatedUsers = [].concat(users, moreUsers);
	
How glimmer-vm doing it:

	updatedUsers -> 1..19 - retain nodes, 20-40 -> insert nodes, this is 20 DOM API calls.

How to optimize?

	- batch inserts while we see chain.
	
Example: 

	const insertStack = [];

	node 21 -> if insert -> insertStack.push(createNode(21))
	node 22 -> if insert -> insertStack.push(createNode(22))
	node 23 -> if insert -> insertStack.push(createNode(23))
	node 23 -> if (NOT insert) 
		-> entryForNode(21).appendChild(DocumentFragment(insertStack))
		-> do (NOT insert) case..

stefanpenner:
Ah, i am surprise this is how we are doing it.

lifeart:
https://ember-twiddle.com/edab79a0d0525d8dd8b8de1dbf82c527?openFiles=templates.components.performant-each.hbs%2Ctemplates.components.performant-each.hbs

![image](https://user-images.githubusercontent.com/1360552/60173496-88e6af00-9817-11e9-9014-9ed5877375cf.png)

Using fragments for initial loops rendering can bering 2x performance!

lifeart:
https://ember-twiddle.com/7190940e60cf314aa2c26a708a884f23?openFiles=templates.application.hbs%2Ctemplates.components.performant-each.hbs  - it's already working case (increase initial rendering performance 2x+), but, I don't think it support fastboot and add one more dom tag.
No perf improvements on rerender

lifeart:
Can't figure out how to get this optimization in glimmer-vm, but, we can create component, like

```hbs
{{#super-each items as |item index|}}
<div> hello ! </div>
{{/super-each}}
```

inside super-each template:

```hbs
<div {{ref this "container"}}>
  {{#let (chunk @items 1000) as |arraySlice|}}
    {{#each arraySlice as |arrayFragment sliceIndex|}}
      {{#let (create-fragment) as |fragment|}}
        {{#in-element fragment}}
          {{#each arrayFragment as |item index|}}
            {{yield item (add index (mult sliceIndex 1000))}}
          {{/each}}
        {{/in-element}}
[...]

PR #1322, opened 2021-06-16
[wip] trying to append items in list using fragment

lifeart:
Trying to get perf values for https://github.com/glimmerjs/glimmer-vm/issues/950

Mr created to get tracerbench results

upd:

looks like current benchmark not cover "append items to already rendered list of items", and perf-improvements should be related to this behavour

### DIFF
--- a/packages/@glimmer/integration-tests/lib/suites/each.ts
+++ b/packages/@glimmer/integration-tests/lib/suites/each.ts
@@ -721,27 +721,27 @@ export class EachSuite extends RenderTest {
     }
   }
 
-  @test
-  're-iterating nested arrays works'() {
-    let arr = [
-      [1, 2, 3, 4, 5],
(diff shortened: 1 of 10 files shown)


vlascik:
Maybe you could try to benchmark it with https://github.com/krausest/js-framework-benchmark ? Ember's create row results have been an embarrassment there for years.

NullVoxPopuli:
Is this still worth pursuing?

lifeart:
@NullVoxPopuli I confused on adding "append" benchmark example, I need time or hep here

wagenet:
It looks like this PR is stalled and since it's still in draft I'm going to go ahead and close it. If you do want to pick it up we'd be happy to do what we can to get it over the finish line.

NullVoxPopuli:
There's also been recentist research written somewhere that fragments now slow things down a smidge

PR #1325, opened 2021-07-19
Remove `partial` functionality

nlfurniss:
Remove `{{partial}}` functionality from Glimmer as it's being removed in Ember 4.0.

Please note, I'm deeply unfamiliar with Glimmer so if I've missed something big or small please let me know 😊

`npm test` passes locally


packages/@glimmer/interfaces/lib/serialize.d.ts
```diff
@@ -23,17 +23,17 @@
...
  *
  * 1. To turn handles created by the `CompileTimeLookup` into live helpers,
- *    modifiers, components, and partials.
- * 2. To resolve dynamic components and partials at runtime that come from
- *    calls to `{{component dynamic}}` or `{{partial dynamic}}`.
+ *    modifiers, and components.
+ * 2. To resolve dynamic components at runtime that come from
+ *    calls to `{{component dynamic}}`.
  *
  * The `CompileTimeLookup` and `RuntimeResolver` must maintain symmetry
```
nlfurniss, line 38: Not sure what to do with this with partials gone, would love some guidance

chadhietala:
Thanks for taking the first pass at this. IIRC there is this concept called `evalScope` which is meant to track the scope in which a partial is used. Its sort of confusing because partial is also referred to `eval` within the code base as well and I would suggest looking to see if the `{{partial}}` implementation was the only consumer of certain APIs.

@pzuraq or @rwjblue might able to provide more guidance here.

nlfurniss:
Thanks Chad. After digging into `evalScope`, I find myself deleting more and more code and running into a lack of knowledge when it comes to getting things working again. I'm going to close this and leave it to the people who actually know what they're doing 🙃

PR #1330, opened 2021-08-29
Remove deprecated support for mutation after consumption during certain manager hooks

snewcomer:
This PR removes supporting infrastructure for allowing, but deprecating read after write in components and modifiers.  Going forward, this will error with a message that you attempted to read then write in the same transaction.

Ref Ember.js issue: https://github.com/emberjs/ember.js/issues/19617

Ember.js PR: https://github.com/emberjs/ember.js/pull/19735

### DIFF
--- a/packages/@glimmer/validator/lib/debug.ts
+++ b/packages/@glimmer/validator/lib/debug.ts
@@ -1,6 +1,6 @@
 import { Tag } from './validators';
 import { DEBUG } from '@glimmer/env';
-import { deprecate, assert } from '@glimmer/global-context';
+import { assert } from '@glimmer/global-context';
 
 export let beginTrackingTransaction:
   | undefined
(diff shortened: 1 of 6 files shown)


packages/@glimmer/validator/lib/debug.ts
```diff
@@ -218,31 +194,23 @@ if (DEBUG) {
...
-      } catch (e) {
-        if (e.stack) {
-          let updateStackBegin = e.stack.indexOf('Stack trace for the update:');
-
-          if (updateStackBegin !== -1) {
-            let start = nthIndex(e.stack, '\n', 1, updateStackBegin);
-            let end = nthIndex(e.stack, '\n', 4, updateStackBegin);
-            e.stack = e.stack.substr(0, start) + e.stack.substr(end);
-          }
+    // This hack makes the assertion message nicer, we can cut off the first
```
NullVoxPopuli, line 197: not related to this PR, but I wonder what is going on here -- why are we cutting of parts of the error message? are those parts ever needed?

NullVoxPopuli (approved):
noice 
image

rwjblue (approved):
Thank you!

PR #1340, opened 2021-09-24
Perf tests

lifeart:
list of some codebase improvements (want to get trasertbench results on it)


packages/@glimmer/util/lib/object-utils.ts
```diff
@@ -31,19 +31,11 @@ function assignFn(obj: any) {
 export let assign = Object.assign ?? assignFn;
 
 export function fillNulls<T>(count: number): T[] {
```
rwjblue, line 33: I'd rather just remove this helper. It seems worse than just using `new Array(999).fill(null)` in each location.

packages/@glimmer/util/lib/object-utils.ts
```diff
@@ -31,19 +31,11 @@ function assignFn(obj: any) {
...
+  return new Array(count).fill(null);
 }
 
-export function values<T>(obj: { [s: string]: T }): T[] {
-  const vals = [];
-  for (const key in obj) {
-    vals.push(obj[key]);
-  }
-  return vals;
+export const objectValues = Object.values ?? objectValuesFn;
```
rwjblue, line 37: Hmm, why can't we assume that `Object.values` is present? I assume that if we drop IE11 we can drop it..

lifeart: we have CI for IE11, and it will fail

lifeart: lol, same with Array.fill

NullVoxPopuli: Ie 11 support is dropped. Can this be updated? (Does it need to be?)

lifeart: @NullVoxPopuli let me try.

NullVoxPopuli:
What's left on this PR?

lifeart:
closing prior to https://github.com/glimmerjs/glimmer-vm/pull/1428

PR #1343, opened 2021-09-29
Remove deprecated versions of modifier/component manager capabilities from error messages

nlfurniss:
Error messages were incorrect, plus they're causing tests in ember.js to fail that check the error messages

### DIFF
--- a/packages/@glimmer/manager/lib/public/component.ts
+++ b/packages/@glimmer/manager/lib/public/component.ts
@@ -121,7 +121,7 @@ export class CustomComponentManager<O extends Owner, ComponentInstance>
       if (DEBUG && !FROM_CAPABILITIES!.has(delegate.capabilities)) {
         // TODO: This error message should make sense in both Ember and Glimmer https://github.com/glimmerjs/glimmer-vm/issues/1200
         throw new Error(
-          `Custom component managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.4' | '3.13')\` (imported via \`import { capabilities } from '@ember/component';\`). Received: \`${JSON.stringify(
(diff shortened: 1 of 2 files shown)


packages/@glimmer/manager/lib/public/component.ts
```diff
@@ -121,7 +121,7 @@ export class CustomComponentManager<O extends Owner, ComponentInstance>
       if (DEBUG && !FROM_CAPABILITIES!.has(delegate.capabilities)) {
         // TODO: This error message should make sense in both Ember and Glimmer https://github.com/glimmerjs/glimmer-vm/issues/1200
         throw new Error(
-          `Custom component managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.4' | '3.13')\` (imported via \`import { capabilities } from '@ember/component';\`). Received: \`${JSON.stringify(
+          `Custom component managers must have a \`capabilities\` property that is the result of calling the \`capabilities('3.13')\` (imported via \`import { capabilities } from '@ember/component';\`). Received: \`${JSON.stringify(
             delegate.capabilities
```
snewcomer, line 125: RE: somewhere closer to the top of the file.  Do we need to change this to `3.22`?

https://github.com/glimmerjs/glimmer-vm/blob/3234f05f7a53b22d6e30bc745fd50fc99d4b8856/packages/%40glimmer/manager/lib/public/component.ts#L45

snewcomer: Oh wait.  That is just modifiers.  So components aren't in lockstep with the modifiers?

snewcomer (approved):
Great job!

packages/@glimmer/manager/lib/public/component.ts
nlfurniss: https://deprecations.emberjs.com/v3.x/#toc_manager-capabilities-components-3-4

Seems like they’re not

rwjblue (approved):
Thank you!

PR #1345, opened 2021-10-01
Cache `hbsPosFor(offset)` in `Source`

chriskrycho:
`hbsPosFor` is currently the hottest path in the current implementation of the template compiler in terms of self time. Adding this cache for it improves throughput by about half a second in [one rough benchmark][bench] (against a collection of about 1,500 templates totaling about 54,000 lines of Handlebars), run 5 times against each version of the compiler:

[bench]: https://github.com/brendenpalmer/repro-ember-3-25-template-regression/pull/2

|           variant            | average | min  | max  |
| ---------------------------- | ------- | ---- | ---- |
| @glimmer/compiler@0.65.3     | 7002.2  | 6976 | 7031 |
| @glimmer/compiler@master     | 9181.4  | 9159 | 9202 |
| @glimmer/compiler@experiment | 8445.2  | 8400 | 8507 |

Additionally, `hbsPosFor` *stops* being the largest self time; in the same rough benchmark, it dropped from ~560ms to ~97ms of self time.

### DIFF
--- a/packages/@glimmer/syntax/lib/source/source.ts
+++ b/packages/@glimmer/syntax/lib/source/source.ts
@@ -31,11 +31,19 @@ export class Source {
     });
   }
 
+  private _hbsPosesFor = new Map<number, SourcePosition | null>();
+
   hbsPosFor(offset: number): Option<SourcePosition> {
+    const current = this._hbsPosesFor.get(offset);
+    if (current !== undefined) {
+      return current;
+    }
+
     let seenLines = 0;
     let seenChars = 0;
 
     if (offset > this.source.length) {
+      this._hbsPosesFor.set(offset, null);
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


chriskrycho:
Note: it appears we should probably pick #1344 over this, and not land both, because that one subsumes this one. (Runtimes here are lower than in the table above because I switched to running the benchmark in sequence to avoid CPU contention interfering. That was also the cause of the wild swings observed back in #1341.)

## `syntax-cache-head` (#1344)

|            variant             | average | min  | max  |
| ------------------------------ | ------- | ---- | ---- |
| `@glimmer/compiler@0.65.3`     | 2339.2  | 2306 | 2362 |
| `@glimmer/compiler@master`     | 3700.4  | 3680 | 3739 |
| `@glimmer/compiler@experiment` | 3052.8  | 3028 | 3082 |

## `syntax-cache-hbsPos` (#1355)

|            variant             | average | min  | max  |
| ------------------------------ | ------- | ---- | ---- |
| `@glimmer/compiler@0.65.3`     | 2381.2  | 2328 | 2483 |
| `@glimmer/compiler@master`     | 3750.6  | 3674 | 3939 |
| `@glimmer/compiler@experiment` | 3156.6  | 3137 | 3200 |

## combined

|            variant             | average | min  | max  |
| ------------------------------ | ------- | ---- | ---- |
| `@glimmer/compiler@0.65.3`     | 2326.6  | 2320 | 2333 |
| `@glimmer/compiler@master`     | 3710    | 3692 | 3742 |
| `@glimmer/compiler@experiment` | 3040.4  | 3002 | 3064 |

PR #1346, opened 2021-10-11
Fix comma splice in `on` modifier error message.

muziejus:


muziejus:
Thanks for your patience. I changed the tests, too.

locks:
Thanks again :)

PR #1348, opened 2021-10-16
Implementation for RFC #756

NullVoxPopuli:
Implementation of https://github.com/emberjs/rfcs/pull/756

Todo:
- [x] PR to the RFC to add an example of existing behavior today where consuming a tracked value in the helper body entangles with helper execution

### DIFF
--- a/packages/@glimmer/manager/lib/internal/defaults.ts
+++ b/packages/@glimmer/manager/lib/internal/defaults.ts
@@ -1,6 +1,10 @@
 import { buildCapabilities } from '../util/capabilities';
 
-import type { CapturedArguments as Arguments, HelperCapabilities } from '@glimmer/interfaces';
+import type {
+  CapturedArguments as Arguments,
+  HelperCapabilities,
+  HelperManagerWithValue,
+} from '@glimmer/interfaces';
 
 type FnArgs<Args extends Arguments = Arguments> =
   | [...Args['positional'], Args['named']]
@@ -20,11 +24,11 @@ export class FunctionHelperManager implements HelperManagerWithValue<State> {
     hasScheduledEffect: false,
   }) as HelperCapabilities;
 
-  createHelper(fn: AnyFunction, args: Arguments) {
+  createHelper(fn: AnyFunction, args: Arguments): State {
     return { fn, args };
   }
 
-  getValue({ fn, args }: State) {
+  getValue({ fn, args }: State): unknown {
     if (Object.keys(args.named).length > 0) {
       let argsForFn: FnArgs<Arguments> = [...args.positional, args.named];
 
@@ -34,7 +38,7 @@ export class FunctionHelperManager implements HelperManagerWithValue<State> {
     return fn(...args.positional);
   }
 
-  getDebugName(fn: AnyFunction) {
+  getDebugName(fn: AnyFunction): string {
     if (fn.name) {
       return `(helper function ${fn.name})`;
     }
--- a/packages/@glimmer/manager/lib/internal/defaults.ts
+++ b/packages/@glimmer/manager/lib/internal/defaults.ts
@@ -13,7 +13,7 @@ interface State {
   args: Arguments;
 }
 
-export class FunctionHelperManager {
+export class FunctionHelperManager implements HelperManagerWithValue<State> {
   capabilities = buildCapabilities({
     hasValue: true,
     hasDestroyable: false,
--- a/packages/@glimmer/manager/lib/internal/defaults.ts
+++ b/packages/@glimmer/manager/lib/internal/defaults.ts
@@ -1,0 +1,44 @@
+import { buildCapabilities } from '../util/capabilities';
+
+import type { CapturedArguments as Arguments, HelperCapabilities } from '@glimmer/interfaces';
+
+type FnArgs<Args extends Arguments = Arguments> =
+  | [...Args['positional'], Args['named']]
+  | [...Args['positional']];
+
+type AnyFunction = (...args: any[]) => unknown;
+
+interface State {
+  fn: AnyFunction;
+  args: Arguments;
+}
+
+export class FunctionHelperManager {
+  capabilities = buildCapabilities({
+    hasValue: true,
+    hasDestroyable: false,
+    hasScheduledEffect: false,
+  }) as HelperCapabilities;
+
+  createHelper(fn: AnyFunction, args: Arguments) {
+    return { fn, args };
+  }
+
+  getValue({ fn, args }: State) {
+    if (Object.keys(args.named).length > 0) {
+      let argsForFn: FnArgs<Arguments> = [...args.positional, args.named];
+
+      return fn(...argsForFn);
+    }
+
+    return fn(...args.positional);
+  }
+
+  getDebugName(fn: AnyFunction) {
+    if (fn.name) {
+      return `(helper function ${fn.name})`;
+    }
+
+    return '(anonymous helper function)';
+  }
+}
--- a/packages/@glimmer/manager/lib/internal/index.ts
+++ b/packages/@glimmer/manager/lib/internal/index.ts
@@ -7,6 +7,7 @@ import {
   Owner,
 } from '@glimmer/interfaces';
 import { CustomHelperManager } from '../public/helper';
+import { FunctionHelperManager } from './defaults';
 
 type InternalManager =
   | InternalComponentManager
@@ -119,6 +120,8 @@ export function setInternalHelperManager<T extends object, O extends Owner>(
   return setManager(HELPER_MANAGERS, manager, definition);
 }
 
+const DEFAULT_MANAGER = new CustomHelperManager(() => new FunctionHelperManager());
+
 export function getInternalHelperManager(definition: object): CustomHelperManager | Helper;
 export function getInternalHelperManager(
   definition: object,
@@ -138,7 +141,13 @@ export function getInternalHelperManager(
     );
   }
 
-  const manager = getManager(HELPER_MANAGERS, definition);
+  let manager = getManager(HELPER_MANAGERS, definition);
+
+  // Functions are special-cased because functions are defined
+  // as the "default" helper, per: https://github.com/emberjs/rfcs/pull/756
+  if (manager === undefined && typeof definition === 'function') {
+    manager = DEFAULT_MANAGER;
+  }
 
   if (manager) {
     return manager;
--- a/packages/@glimmer/manager/test/managers-test.ts
+++ b/packages/@glimmer/manager/test/managers-test.ts
@@ -183,9 +183,24 @@ module('Managers', () => {
       assert.ok(typeof instance === 'object', 'manager is an internal manager');
       assert.ok(
         typeof instance.getHelper({}) === 'function',
-        'manager can generatew helper function'
+        'manager can generate helper function'
       );
       assert.equal(instance['factory'], factory, 'manager has correct delegate factory');
+    });
+
+    test('it determines the default manager', (assert) => {
+      let myTestHelper = () => 0;
+      let instance = getInternalHelperManager(myTestHelper) as CustomHelperManager<object>;
+
+      assert.ok(typeof instance === 'object', 'manager is an internal manager');
+      assert.ok(
+        typeof instance.getHelper({}) === 'function',
+        'manager can generate helper function'
+      );
+      assert.strictEqual(
+        instance['factory']({})?.getDebugName?.(myTestHelper),
+        '(helper function myTestHelper)'
+      );
     });
 
     test('it works with internal helpers', (assert) => {
--- a/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
+++ b/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
@@ -32,6 +32,153 @@ class HelperManagerTest extends RenderTest {
     this.assertHTML('hello');
   }
 
+  @test '(Default Helper Manager) plain functions work as helpers'(assert: Assert) {
+    let count = 0;
+
+    const hello = () => {
+      count++;
+      return 'plain function';
+    };
+
+    const Main = defineComponent({ hello }, '{{hello}}');
+
+    this.renderComponent(Main);
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+
+    this.rerender();
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+  }
+
+  @test '(Default Helper Manager) plain functions track positional args'(assert: Assert) {
+    let count = 0;
+
+    let obj = (x: string) => {
+      count++;
+      return x;
+    };
+    let args = trackedObj({ value: 'hello', unused: 'unused' });
+
+    this.renderComponent(defineComponent({ obj }, '{{obj @value @unused}}'), args);
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('hello');
+
+    args.value = 'there';
+    this.rerender();
+
+    assert.equal(count, 2, 'rendered twice');
+    this.assertHTML('there');
+
+    args.unused = 'unused2';
+    this.rerender();
+
+    assert.equal(count, 3, 'rendered thrice');
+    this.assertHTML('there');
+  }
+
+  @test '(Default Helper Manager) plain functions entangle with any tracked data'(assert: Assert) {
+    let count = 0;
+    let trackedState = trackedObj({ value: 'hello' });
+
+    let obj = () => {
+      count++;
+      return trackedState.value;
+    };
+
+    this.renderComponent(defineComponent({ obj }, '{{obj}}'));
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('hello');
+
+    trackedState.value = 'there';
+    this.rerender();
+    this.assertHTML('there');
+    assert.equal(count, 2, 'rendered twice');
+  }
+
+  @test '(Default Helper Manager) plain functions do not track unused named args'(assert: Assert) {
+    let count = 0;
+
+    let obj = (x: string, _options: Record<string, unknown>) => {
+      count++;
+      return x;
+    };
+    let args = trackedObj({ value: 'hello', unused: 'unused' });
+
+    this.renderComponent(defineComponent({ obj }, '{{obj @value namedOpt=@unused}}'), args);
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('hello');
+
+    args.unused = 'unused2';
+    this.rerender();
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('hello');
+  }
+
+  @test '(Default Helper Manager) plain functions tracked used named args'(assert: Assert) {
+    let count = 0;
+
+    let obj = (_x: string, options: Record<string, unknown>) => {
+      count++;
+      return options.namedOpt;
+    };
+
+    let args = trackedObj({ value: 'hello', used: 'used' });
+
+    this.renderComponent(defineComponent({ obj }, '{{obj @value namedOpt=@used}}'), args);
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('used');
+
+    args.used = 'there';
+    this.rerender();
+
+    assert.equal(count, 2, 'rendered twice');
+    this.assertHTML('there');
+  }
+
+  @test '(Default Helper Manager) plain function helpers can have default values (missing data)'(
+    assert: Assert
+  ) {
+    let count = 0;
+    let obj = (x = 'default value') => {
+      count++;
+      return x;
+    };
+
+    let args = trackedObj({});
+
+    this.renderComponent(defineComponent({ obj }, 'result: {{obj}}'), args);
+    this.assertHTML('result: default value');
+    assert.equal(count, 1, 'rendered once');
+  }
+
+  @test '(Default Helper Manager) plain function helpers can have overwritten default values'(
+    assert: Assert
+  ) {
+    let count = 0;
+    let obj = (x = 'default value') => {
+      count++;
+      return x;
+    };
+
+    let args = trackedObj({ value: undefined });
+
+    this.renderComponent(defineComponent({ obj }, 'result: {{obj @value}}'), args);
+    this.assertHTML('result: default value');
+    assert.equal(count, 1, 'rendered once');
+
+    args.value = 'value';
+    this.rerender();
+
+    this.assertHTML('result: value');
+    assert.equal(count, 2, 'rendered twice');
+  }
+
   @test 'tracks changes to named arguments'(assert: Assert) {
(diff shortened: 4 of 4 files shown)


NullVoxPopuli:
> Error: Could not find module `@glimmer/interfaces` imported from `@glimmer/manager`

should interfaces be a dep in manager?

NullVoxPopuli:
@rwjblue ready for review / added some tests. Not sure how many test scenarios you want / feel necessary, so if you think of more that are needed, lemme know

packages/@glimmer/manager/test/managers-test.ts
```diff
@@ -183,11 +183,18 @@ module('Managers', () => {
...
+        'manager can generate helper function'
       );
       assert.equal(instance['factory'], factory, 'manager has correct delegate factory');
     });
 
+    test('it determines the default manager', (assert) => {
+      let helper = () => 0;
+      let instance = getInternalHelperManager(helper);
+
+      assert.ok(instance, helper as any);
```
snewcomer, line 195: assert.equal?

NullVoxPopuli: whoops

NullVoxPopuli: fixed

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,37 @@
...
+    let argsForFn: FnArgs<Arguments> = [
+      ...args.positional,
+      Object.keys(args.named).length > 0 ? args.named : {},
+    ];
+
+    return fn(...argsForFn);
+  }
+
+  getDebugName(fn: State['fn']) {
+    return fn.name || '(anonymous helper function)';
```
snewcomer, line 35: Does anything in the build compression or magnification perhaps make `name` not what we think it is?

NullVoxPopuli: Minification could change the function name. (If in module space) But folks shouldn't rely on function name stability?

NullVoxPopuli: Ie doesn't have function names, it seems.

Can we drop support for ie?

chancancode:
This one probably requires more focused attention, will come back to it next week

NullVoxPopuli:
sg,  rebased

NullVoxPopuli:
@chancancode should be green - linting seems to have failed for random networking reasons

lifeart (approved):

krisselden (approved):

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  getValue({ fn, args }: State) {
+    let argsForFn: FnArgs<Arguments> = [
+      ...args.positional,
+      Object.keys(args.named).length > 0 ? args.named : {},
+    ];
+
+    return fn(...argsForFn);
+  }
```
chancancode, line 32: What is this used for and why do we need it? What happens if you don't have this?

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -138,7 +139,18 @@ export function getInternalHelperManager(
     );
   }
 
-  const manager = getManager(HELPER_MANAGERS, definition)!;
+  let manager = getManager(HELPER_MANAGERS, definition)!;
```
chancancode, line 142: This `!` seems suspect (we still check for `undefined`, etc later on) – what is it doing for us here?

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -138,7 +139,18 @@ export function getInternalHelperManager(
...
+
+  // Functions are special-cased because functions are defined
+  // as the "default" helper, per: https://github.com/emberjs/rfcs/pull/756
+  if (manager === undefined && typeof definition === 'function') {
+    setInternalHelperManager(
+      new CustomHelperManager(() => new FunctionHelperManager()),
+      definition
+    );
+
+    manager = getManager(HELPER_MANAGERS, definition)!;
```
chancancode, line 152: Do we really need to set the manager, as opposed to just returning it (well, assigning it to the variable here and 167 will do the returning)? Is this effectively just caching the lookup?

Also, is there a way to avoid making a new instance each time, since the manager seems entirely stateless?

packages/@glimmer/manager/lib/internal/index.ts
NullVoxPopuli: this was pre-existing, but it's because at the end of the function `return manager` expects non-undefined. If manager ever is undefined, we could throw an error and get away from both `!`, but is that the correct behavior? happy to add if it is.

chancancode:
I also can't tell from the RFC:

```js
class MyComponent {
  @tracked name = '...';

  greet(message) {
    return `${message}, ${this.name}`; // <- do we call this function again when `this.name` changes?
  }
}
```

```hbs
{{this.greet "hello"}}
```

Personally I think the answer is yes (and there probably isn't other options given how the system works), but 1. we should have test for it; 2. we should update the RFC to state it explicitly

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  createHelper(fn: State['fn'], args: State['args']) {
+    return { fn, args };
+  }
+
+  getValue({ fn, args }: State) {
+    let argsForFn: FnArgs<Arguments> = [
+      ...args.positional,
+      Object.keys(args.named).length > 0 ? args.named : {},
+    ];
```
chancancode, line 29: This seems unnecessary – if we got here by definition we don't have arguments so we just need to call the function. However, there is the open question of whether it should be `fn()` vs `fn({})`.

packages/@glimmer/manager/lib/internal/index.ts
NullVoxPopuli: oh nice! I hadn't thought of that. updated

packages/@glimmer/manager/lib/internal/defaults.ts
chancancode: We probably shouldn't try to use `fn.name` at all, or at least not by itself, because of minification precisely (so it'll probably just be `"zw"` or something cryptic in prod). We could do something like:

```js
if (fn.name) {
  return `(helper function ${fn.name})`;
} else {
  return '(anonymous helper function)';
}
```

Is there an existing format/convention from the helper manager that we can steal/align with?

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: I think it should be `fn({})`, because at the call site, it's a common pattern (that I've seen anyway (outside of ember)), to do like:
```js
let options = args[args.length - 1];
// or
let [options] = args.reverse();
```
giving a stable type is good DX, imo, and reduces the amount of surprise when arguments aren't passed.

packages/@glimmer/manager/lib/internal/index.ts
chancancode: Is typescript not able to follow the code somehow? It seems like we branched in a way that I would expect TS to be able to figure it out, since we did check for undefined.

But anyway, we can structure it like this:

```ts
let manager = ...;

if (manager) {
  return manager;
} else if (typeof definition === 'function') {
  return DEFAULT_MANAGER;
} else if (isOptional === true) { // === true seems a bit extra but whatev
  return null;
} else {
  throw new Error(...);
}
```

NullVoxPopuli:
> i.e. should {{this.myFunc}} be called as this.myFunc.call(undefined) or this.myFunc.call(undefined, {})?

I think it should be `myFunc({})`, because at the call site, it's a common pattern (that I've seen anyway (outside of ember)), to do like:
```js
let options = args[args.length - 1];
// or
let [options] = args.reverse();
```
giving a stable type is good DX, imo, and reduces the amount of surprise when arguments aren't passed.

> .call(undefined, {})?

why call? normal invocation preserves the `this` of the function (if any), and allows people to entangle with the helper function (an existing behavior of today's helper functions).

> I also can't tell from the RFC:
> the answer is yes (and there probably isn't other options given how the system works), 

yes, autotracking is maintained with helpers today. 
but 

> 1. we should have test for it; 

there already is a test, `(Default Helper Manager) plain functions track positional args` and `(Default Helper Manager) plain functions do not track unused named args`

> 2. we should update the RFC to state it explicitly

Sounds reasonable (esp since our RFCs tend to be the source of truth for docs until an edition lands).
I'll PR that tomorrow (hopefully)

packages/@glimmer/manager/lib/internal/defaults.ts
chancancode: If that's the case then you can't ever write something like this:

```js
class MyComponent {
  @tracked name = '...';

  greet(message = 'hello') { // <- this will always be filled with `{}`
    return `${message}, ${this.name}`;
  }
}
```

More importantly, you can't import and use a lot of library functions that uses patterns like this either (it doesn't have to be the only/first argument – as long as it takes a variable number of positional arguments, i.e. any function with at least one default positional argument – then the result would be wrong.

Which seems like a big deal considering this is supposed to make it possible to just import an existing function and use it in the template directly.

chancancode:
> I think it should be myFunc({})

See above. Not saying it should definitely be `myFunc()` or that it wasn't without its own problems, but I think `myFunc({})` has pretty obvious problems that should require us to rethink this a bit.

> Why `.call(undefined, {})?`

I was just being clear/precise/pedantic, this already is what's happening. When you write `{{this.foo}}`, we get `fn = this[foo]` as a value, and by the time you do `fn()` the `this` is already lost. To make it a method call you would need `@action` or `this.foo.bind(this)`, in which case the `.call(undefined, ...)` does not affect it as the `this` is already "baked in" to the value.

> yes, autotracking is maintained with helpers today

I am specifically asking about when the tracked state does NOT come from an argument (i.e. from `this` and closure captures). I don't think that was explicit in the RFC nor were there tests for it.

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: yeah, that's true. :thinking: 
default arguments are enough to have me switch the implementation. Next commit

chancancode: Let's make sure we test it too, and try to give this some visibility so people other than us can chime in (a PR to update the RFC is probably the easiest to get attention).

This is just one possible path, but there is also the more generic versions, such as `{{(someFunction)}}`. What happens in that case?

chancancode: > This seems unnecessary – if we got here by definition we don't have arguments so we just need to call the function.

Sorry – reading the code more, I think I am wrong about this (doesn't affect the `()` vs `({})` discussion but we should probably make that its own thread). I thought `getValue()` is only called in the `{{someFn}}` case, but that doesn't seem like how it works and is just the part of the helper manager api that mans "call the helper". So I think your args-handling code is actually needed.

NullVoxPopuli:
> tracked state does NOT come from an argument 

yes, I know. that's still how it works today, you access `this.trackedThing` in a helper defined on your component, and it entangles properly.

```js
export default class Demo extends Component {
  @tracked foo = 3;
  
  myHelper = helper(() => this.foo * 2); // properly entangles
}
```

NullVoxPopuli:
> I was just being clear/precise/pedantic, this already is what's happening.

gotchya, thanks!

> has pretty obvious problems that should require us to rethink this a bit.

yeah, I'm on your side now -- working on a test for this

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: > could do something like:

sounds good to me!

> Is there an existing format/convention from the helper manager that we can steal/align with?

I didn't see one as I was browsing around

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  getValue({ fn, args }: State) {
+    let namedKeys = Object.keys(args.named).length;
+
+    if (args.positional.length === 0 && namedKeys === 0) {
+      return fn();
+    }
+
+    let argsForFn: FnArgs<Arguments> = [...args.positional, namedKeys > 0 ? args.named : {}];
```
chancancode, line 32: I think this doesn't work when you have positional but not named. You'll still be passing the empty object in that case. I think you need something like

```js
let args = [...args.positional];

if (Object.keys(args.named).length > 0) {
  args.push(args.named);
}

return fn(...args);
```

NullVoxPopuli: nay,
image

default arguments are still allowed with the tailing `{}` arg

chancancode: This is what I am saying won't work: `function greeting(name, message = "hello")`

chancancode:
> yes, I know. that's still how it works today, you access this.trackedThing in a helper defined on your component, and it entangles properly.

IMO we should still add tests for this case for the default manager

> just PRing to the RFC an example showing current behavior of auto-tracking local tracked data accessed from within the helpers

Sounds good. I think this is good to clarify, but probably secondary to/not as important as clarifying whether there is or is not an automatic `{}` in the args when no named arguments are passed. I think that is an important design question that was missed in the original RFC, so it's good to make sure we are arriving at the right conclusion. I think we are both leaning no at this point, but we started with opposite intuitions so that's probably good indication that it's not a trivial thing.

If you are able to open the PR, I can bring it up for discussion on Friday to give it some attention, then we can see if we decided to just merge that PR, or tweet that PR for 1-week FCP, etc.

Thanks for working on this!

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: ahhhh yeah, I gotchya

chancancode:
> IMO we should still add tests for this case for the default manager

I think the easiest way to do that in tests would be something like

```
// Probably try to make this a slightly more real
class Greeter {
  @tracked name = '...';
}

let greeter = new Greeter();

let helper = () => return `Hello ${greeter.name}`;

// render, assert

greeter.name = '...something else...';

// assert rerendered
```

Not sure what is available in the test harness here, alternatively you may be able to use a "tracked object" thing, but you can probably find examples in tests covering this behavior in regular helpers

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,50 @@
+import { buildCapabilities } from '../util/capabilities';
+
+import type { CapturedArguments as Arguments, HelperCapabilities } from '@glimmer/interfaces';
+
+type FnArgs<Args extends Arguments = Arguments> =
+  | [...Args['positional'], Args['named']]
+  | [...Args['positional'], {}];
```
chancancode, line 7: I think this is wrong now (if we decided to go with not always passing the `{}`). At this point, isn't it just `type AnyFunction =  (...unknown[]): unknown`?

NullVoxPopuli: yup. updates incoming :partying_face:

NullVoxPopuli: fixed

NullVoxPopuli:
> I think the easiest way to do that in tests would be something like

added test: `(Default Helper Manager) plain functions entangle with any tracked data`

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+type FnArgs<Args extends Arguments = Arguments> =
+  | [...Args['positional'], Args['named']]
+  | [...Args['positional']];
+
+interface FunctionHelperState<Args extends Arguments = Arguments> {
+  fn: <Return>(...args: FnArgs<Args>) => Return;
+  args: Args;
+}
+
+export class FunctionHelperManager<State extends FunctionHelperState> {
```
chancancode, line 14: @chriskrycho @dfreeman does Glint consume these types in any way? If not,  have some doubts about the utility of having this generic here. Inside the VM, we never know what the types of the user supplied objects are so things just get filled with `unknown`. If this is not useful to Glint either I think we can just remove them to keep things simple. It would be different if this is subclassable.

NullVoxPopuli: >  utility of having this generic here

it was mostly for me to type less.
Happy with whatever makes the most sense though

chancancode: Assuming the Glint team has no use for it, you can probably just do:

```
type AnyFunction = (...args: unknown[]) => unknown;

interface FunctionHelperState {
  fn: AnyFunction;
  args: Argument;
}
```

You can keep keep using `State['fn']` and `State['args']` if you prefer, or you can just replace them with `AnyFunction` and `Arguments` at that point

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+  capabilities = buildCapabilities({
+    hasValue: true,
+    hasDestroyable: false,
+    hasScheduledEffect: false,
+  }) as HelperCapabilities;
+
+  createHelper(fn: State['fn'], args: State['args']) {
```
chancancode, line 21:
```suggestion
  createHelper(fn: State['fn'], args: State['args']): State {
```

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+    return { fn, args };
+  }
+
+  getValue({ fn, args }: State) {
```
chancancode, line 25:
```suggestion
  getValue({ fn, args }: State): unknown {
```

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (Object.keys(args.named).length > 0) {
+      let argsForFn: FnArgs<Arguments> = [...args.positional, args.named];
+
+      return fn(...argsForFn);
+    }
+
+    return fn(...args.positional);
+  }
+
+  getDebugName(fn: State['fn']) {
```
chancancode, line 35:
```suggestion
  getDebugName(fn: State['fn']): string {
```

packages/@glimmer/manager/lib/internal/defaults.ts
chriskrycho: I’ll let Dan confirm but I don’t think we’d end up using these. I will note though that the type should probably be `(...args: any[]) => unknown` because otherwise subtyping is impossible due to the way that variance works with function parameters.

dfreeman: Glint has no awareness of the managers that back template-invokable entities (as the effective interface of the entity in a template isn't robustly derivable from the type of its manager), so keeping the types simple here seems like a good call to me 👍

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: (otherwise I feel these changes are kinda superficial)

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,42 @@
+import { buildCapabilities } from '../util/capabilities';
+
+import type { CapturedArguments as Arguments, HelperCapabilities } from '@glimmer/interfaces';
+
+type FnArgs<Args extends Arguments = Arguments> =
+  | [...Args['positional'], Args['named']]
+  | [...Args['positional']];
+
+interface FunctionHelperState<Args extends Arguments = Arguments> {
+  fn: <Return>(...args: FnArgs<Args>) => Return;
```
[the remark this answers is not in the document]
NullVoxPopuli: Thanks for the info! (and link)!
That's a great rule to follow, I like it.

NullVoxPopuli:
Types are now greatly simplified. Thanks @chancancode, @chriskrycho, and @dfreeman !!!

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: latest push has simplifications. thanks all!

packages/@glimmer/manager/lib/internal/defaults.ts
chancancode: Hm, there may be a disconnect, it's not about not using interfaces, I am commenting on using explicit return types.

packages/@glimmer/manager/lib/internal/defaults.ts
```diff
@@ -0,0 +1,44 @@
...
+  | [...Args['positional']];
+
+type AnyFunction = (...args: any[]) => unknown;
+
+interface State {
+  fn: AnyFunction;
+  args: Arguments;
+}
+
+export class FunctionHelperManager {
```
chancancode, line 16:
```suggestion
export class FunctionHelperManager implements HelperManagerWithValue<State> {
```

Otherwise we can make a mistake in here (or the requirements might have changed), but you will get the error in a very far away location (probably where you try to pass it into the constructor of `CustomHelperManager`)

packages/@glimmer/manager/lib/internal/defaults.ts
NullVoxPopuli: > interface
> inference

these words are annoyingly similar lol

NullVoxPopuli: there is an eslint-todo system out there somewhere, yeah? been meaning to look into one for work.

NullVoxPopuli: but anywho, if you're open to a lint, I can figure that out, too, and just add these changes

chancancode: lol sorry about that. Yeah, I don't think we are "against inference" in general (e.g. it's fine to use them when declaring variables), but definitely not in return positions, since they are part of the function/method's API (public or private), so you don't want a refactor in the function body to _accidentally_ change the API of the function. It's fine if you did it intentionally and had to change the return type, but that would be a good thing to see in the diff during a PR review (for example).

So I think roughly the rule is don't use inference for things that affects your consumers, and use them all you want inside your private implementation details. Mostly that applies to return types but probably a good practice for things like exported `const`s, etc.

Arguably that should probably apply for function parameters too (when defaulting), but usually we seem ok with it, probably because it would be pretty obvious in the PR when you made a change in the parameter list, whereas the return type can change due to any changes in the function body which is much less obvious.

sandstrom:
@NullVoxPopuli Awesome! 🎉

PR #1351, opened 2021-10-21
Support lexical scope in loose mode

ef4:
This allows you to pass lexically scoped values into templates even when those templates are not in strict mode. The whole system is already designed to make this work correctly, all that is required is to not actively stop it from working.

This use case is hugely useful as a compile target for embroider because embroider already does built-time resolution of components and helpers in existing (non-strict) templates. Today our compiler output does:

```js
import a0 from './path/to/some-component';
import { precompileTemplate } from '@ember/template-compilation';
window.define('your-app/components/some-component', function() { return a0 });
export default precompileTemplate("<SomeComponent />");
```

With this feature it can become:
```js
import SomeComponent from './path/to/some-component';
import { precompileTemplate } from '@ember/template-compilation';
export default precompileTemplate("<SomeComponent />", { scope: () => ({ SomeComponent }), strictMode: false });
```

which means we skip over a large amount of runtime registration and lookup.

### DIFF
--- a/packages/@glimmer/integration-tests/test/lexical-scope-test.ts
+++ b/packages/@glimmer/integration-tests/test/lexical-scope-test.ts
@@ -1,0 +1,17 @@
+import { RenderTest, test, jitSuite, defineComponent } from '..';
+
+class LexicalScopeTest extends RenderTest {
+  static suiteName = 'loose mode: lexical scope';
+
+  @test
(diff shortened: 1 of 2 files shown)


ef4:
So it turns out the test coverage here is extremely unrealistic and this is insufficient to make it work for actual precompiled templates. Followup PR coming.

PR #1362, opened 2021-11-05
Fix issues with "loose free variable" lookup by dropping unneeded `GetFreeAsFallback` op code

ef4:
The `GetFreeAsFallback` opcode got broken by 179a388d69af6fcd1bc906de396bde55e4925933. As far as I can tell, it's never correct to call `withPath` by itself, it's always supposed to come after an opcode that leaves a value on the stack that needs path support.

Rather than fix it, I'm going to try to cut the rest of it out, since it exists only to support `this` fallback.

This still needs to get tested against Ember. I am unable to get Ember and glimmer-vm working correctly together locally, with or without these changes.

### DIFF
--- a/packages/@glimmer/compiler/lib/builder/builder.ts
+++ b/packages/@glimmer/compiler/lib/builder/builder.ts
@@ -327,7 +327,7 @@ function buildKeyword(
       return [Op.If, expect(params, 'if requires params')[0], block, inverse];
     case 'each':
       let keyExpr = normalized.hash ? normalized.hash['key'] : null;
-      let key = keyExpr ? buildExpression(keyExpr, 'Generic', symbols) : null;
+      let key = keyExpr ? buildExpression(keyExpr, 'Strict', symbols) : null;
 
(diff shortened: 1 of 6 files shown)


packages/@glimmer/compiler/lib/builder/builder.ts
```diff
@@ -583,12 +582,7 @@ export function buildCallHead(
 }
 
 export function buildGetPath(head: NormalizedPath, symbols: Symbols): Expressions.GetPath {
-  return buildVar(
-    head.path.head,
-    VariableResolutionContext.LooseFreeVariable,
-    symbols,
-    head.path.tail
-  );
+  return buildVar(head.path.head, VariableResolutionContext.Strict, symbols, head.path.tail);
```
snewcomer, line 585: 👍

snewcomer (approved):
Ah ha!  Generic -> Strict.  I'm glad you got to the bottom of this!

rwjblue (approved):
Thanks for working on this @ef4!

PR #1364, opened 2021-11-17
Correct grammar in immutable args error

jamescdavis:
The semicolon was just because I was here, but the singular linking verb for a plural subject was activating me. 😆

### DIFF
--- a/packages/@glimmer/manager/lib/util/args-proxy.ts
+++ b/packages/@glimmer/manager/lib/util/args-proxy.ts
@@ -150,7 +150,7 @@ if (HAS_NATIVE_PROXY) {
         throw new Error(
           `You attempted to set ${String(
             prop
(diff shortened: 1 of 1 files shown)


packages/@glimmer/manager/lib/util/args-proxy.ts
```diff
@@ -150,7 +150,7 @@ if (HAS_NATIVE_PROXY) {
         throw new Error(
           `You attempted to set ${String(
             prop
-          )} on the arguments of a component, helper, or modifier. Arguments are immutable and cannot be updated directly, they always represent the values that is passed down. If you want to set default values, you should use a getter and local tracked state instead.`
+          )} on the arguments of a component, helper, or modifier. Arguments are immutable and cannot be updated directly; they always represent the values that are passed down. If you want to set default values, you should use a getter and local tracked state instead.`
```
snewcomer, line 153: Another option is `updated directly.  They always...`

snewcomer (approved):

chancancode (approved):

PR #1368, opened 2022-01-04
Add docs for `setComponentManager`

chriskrycho:
I am of course happy to elaborate here further as is useful!


chriskrycho:
…I’m an idiot; wrong repo.

PR #1381, opened 2022-02-12
chore: use prettier as default code formatter to avoid formatting conflicts

lifeart:
use prettier as default code formatter to avoid formatting conflicts in vscode


NullVoxPopuli (approved):
Seems good to me 👍

wycats:
This has been adopted (in a slightly different way) in a recent commit. Thanks for the suggestion!

Issue #1383, opened 2022-03-08
Feature flag for default helper manager


chancancode:
We landed support for default helper manager in #1348. However, we forgot to include a way to turn the feature off. This is needed because every new feature landing in Ember need to be flagged. The general idea is:

1. In here, we need to add a boolean flag for the feature (either directly on the interface, or nested under `features` or something). please the export: https://github.com/glimmerjs/glimmer-vm/blob/e00baec72b4de60487a290265c818cb53bfb9c35/packages/%40glimmer/global-context/index.ts#L138

2. Then in the new code introduced in #1348, we can add a conditional that import and checks that flag prior to returning the default helper manager.

3. Then we will add an Ember feature flag on the Ember side, and reference it in here: https://github.com/emberjs/ember.js/blob/f89dd90290e993ce8bc9ea5f7f103a82687a2be9/packages/%40ember/-internals/glimmer/lib/environment.ts#L21

4. Release these as 0.84.1 and tweak https://github.com/emberjs/ember.js/pull/20013 accordingly

cc @wagenet @NullVoxPopuli (though I don't think either one of them has the time to do it, so feel free to pick this up)

locks:
@chancancode do you know of a feature flag example in glimmer-vm that a contributor could base themselves on?

PR #1384, opened 2022-03-12
Add a `DEFAULT_HELPER_MANAGER` feature flag

Windvis:
I gave adding a feature flag a shot based on the information in #1383. I went for a similar approach as the Ember feature flagging system (albeit a bit simpler). Let me know if I should do things differently!

Closes https://github.com/glimmerjs/glimmer-vm/issues/1383

### DIFF
--- a/packages/@glimmer/global-context/index.ts
+++ b/packages/@glimmer/global-context/index.ts
@@ -17,6 +17,10 @@ import { DEBUG } from '@glimmer/env';
 
 //////////
 
+export let FEATURE_DEFAULT_HELPER_MANAGER = true;
+
+//////////
+
 /**
  * Interfaces
  *
@@ -154,6 +158,9 @@ export interface GlobalContext {
       id: string;
     }
   ) => void;
+  FEATURES?: {
+    DEFAULT_HELPER_MANAGER?: boolean;
+  };
 }
 
 let globalContextWasSet = false;
@@ -179,6 +186,10 @@ export default function setGlobalContext(context: GlobalContext) {
   warnIfStyleNotTrusted = context.warnIfStyleNotTrusted;
   assert = context.assert;
   deprecate = context.deprecate;
+
+  if (typeof context.FEATURES?.DEFAULT_HELPER_MANAGER === 'boolean') {
+    FEATURE_DEFAULT_HELPER_MANAGER = context.FEATURES.DEFAULT_HELPER_MANAGER;
+  }
 }
 
 export let assertGlobalContextWasSet: (() => void) | undefined;
--- a/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
+++ b/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
@@ -10,6 +10,9 @@ import {
   TestHelperManager,
 } from '../..';
 import { Arguments, Owner } from '@glimmer/interfaces';
+import { FEATURE_DEFAULT_HELPER_MANAGER } from '@glimmer/global-context';
+
+const SKIP_DEFAULT_HELPER_MANAGER_TESTS = !FEATURE_DEFAULT_HELPER_MANAGER;
 
 class HelperManagerTest extends RenderTest {
   static suiteName = 'Helper Managers';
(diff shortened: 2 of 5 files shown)


packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
```diff
@@ -10,423 +10,426 @@ import {
   TestHelperManager,
 } from '../..';
 import { Arguments, Owner } from '@glimmer/interfaces';
+import { FEATURES } from '@glimmer/global-context';
 
-class HelperManagerTest extends RenderTest {
-  static suiteName = 'Helper Managers';
+if (FEATURES.DEFAULT_HELPER_MANAGER) {
```
Windvis, line 15: It seems I disabled too many tests here. I'll try to limit it to the needed tests tomorrow.

packages/@glimmer/global-context/index.ts
```diff
@@ -179,6 +189,8 @@ export default function setGlobalContext(context: GlobalContext) {
   warnIfStyleNotTrusted = context.warnIfStyleNotTrusted;
   assert = context.assert;
   deprecate = context.deprecate;
+
+  assign(FEATURES, context.FEATURES);
```
Windvis, line 193: I ran into the Object.assign error that was mentioned here as well: https://github.com/glimmerjs/glimmer-vm/pull/1338#issuecomment-924446617 

Do we still support IE11 here or can we just [remove the code that triggers the error](https://github.com/glimmerjs/glimmer-vm/blob/4f1bef0d9a8a3c3ebd934c5b6e09de4c5f6e4468/test/index.html#L42-L46)?

packages/@glimmer/global-context/index.ts
```diff
@@ -14,6 +14,13 @@
  */
 
 import { DEBUG } from '@glimmer/env';
+import { assign } from '@glimmer/util';
+
+//////////
+
+export const FEATURES = {
```
Windvis, line 21: I just placed it near the top of the file, let me know if I should put it somewhere else.

packages/@glimmer/global-context/index.ts
```diff
@@ -14,6 +14,13 @@
...
+  DEFAULT_HELPER_MANAGER: false,
+};
```
chancancode, line 23: Let's make this a top level `let` variable:

```suggestion
export let FEATURE_DEFAULT_HELPER_MANAGER = false;
```

The `const` here only mean the object cannot be reassigned, but the POJO can be mutated by whoever imports it, which introduces a "backdoor" for changing the value of the flags without going through the `setGlobalContext` API. (Using an object in that API is fine though.)

packages/@glimmer/global-context/index.ts
chancancode: We don't need to support IE11, but anyway with the change proposed above this is no longer relevant.

packages/@glimmer/global-context/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.84.0",
   "repository": "https://github.com/glimmerjs/glimmer-vm/tree/master/packages/@glimmer/global-context",
   "dependencies": {
-    "@glimmer/env": "^0.1.7"
+    "@glimmer/env": "^0.1.7",
+    "@glimmer/util": "0.84.0"
```
chancancode, line 7: Can be removed now (see above)

packages/@glimmer/global-context/index.ts
chancancode: Also let's default this to `true` by default.

packages/@glimmer/integration-tests/lib/modes/env.ts
```diff
@@ -148,6 +148,10 @@ setGlobalContext({
       actualDeprecations.push(msg);
     }
   },
+
+  FEATURES: {
+    DEFAULT_HELPER_MANAGER: true,
+  },
```
chancancode, line 154: With the feature being true by default we won't need this

chancancode (commented):
Thanks for picking this up! Mostly looks good to me, just a few minor changes

Windvis:
@chancancode Thanks for the quick review! I've pushed the requested changes :+1:.

chancancode (approved):

chancancode:
Thank you! Released as 0.84.1!

PR #1388, opened 2022-04-04
Failing test for printer bug

ef4:
This PR is a failing test to illustrate a bug.

If you parse to ast and immediately print back out, it turns:

```hbs
<Link></Link>
```

into

```hbs
<Link>
```

This behavior is special for `<Link>`. Other names don't do this.

I suspect it's because, in strict HTML spec terms `<Link>` is the same as `<link>`, and `<link>` is supposed to not have a closing tag. But in Glimmer/Ember terms, `<Link>` is a component and dropping the closing tag introduces a syntax error.

This was found in https://github.com/embroider-build/embroider/issues/1176

### DIFF
--- a/packages/@glimmer/syntax/test/generation/print-test.ts
+++ b/packages/@glimmer/syntax/test/generation/print-test.ts
@@ -21,6 +21,7 @@ let templates = [
   '<Foo />',
   '<Foo as |bar|>{{bar}}</Foo>',
   '{{#in-element this.someElement}}Content here{{/in-element}}',
+  '<Link></Link>',
 
   // void elements
   '<br>',


Windvis:
This seems to be caused by [this line](https://github.com/glimmerjs/glimmer-vm/blob/07db0655c8d4b5c71792bf49ba3b547c196e460a/packages/%40glimmer/syntax/lib/generation/printer.ts#L248). The element is never closed because [`link` is a void element](https://github.com/glimmerjs/glimmer-vm/blob/07db0655c8d4b5c71792bf49ba3b547c196e460a/packages/%40glimmer/syntax/lib/generation/printer.ts#L9) and `Link` will be lowercased to `link` in that check.

I'm not sure what the correct fix is though. The printer doesn't seem to "know" about components. Should we add a simple `PascalCase` check to consider an element a component? I assume it's more complex than that?

ef4:
I suspect we just need to accept that we diverge from HTML in this spot. It's not like we don't already diverge in others (like block yields `as |thing|` syntax.

PR #1389, opened 2022-04-09
Fix a problem where plain functions can't be used as helper if they are not passed through the strict mode context object

Windvis:
While working on https://github.com/emberjs/ember.js/pull/20052 I noticed that an [error was thrown](https://github.com/glimmerjs/glimmer-vm/blob/07db0655c8d4b5c71792bf49ba3b547c196e460a/packages/%40glimmer/runtime/lib/compiled/opcodes/content.ts#L59) when using a function that was stored on a component class property. The [`hasInternalHelperManager` util](https://github.com/glimmerjs/glimmer-vm/blob/07db0655c8d4b5c71792bf49ba3b547c196e460a/packages/%40glimmer/manager/lib/internal/index.ts#L220) is used there but it doesn't pass through the default helper manager fallback behavior code so it returns false and the error is shown.

It seems functions passed as strict mode context values follow a different code path than component arguments and class properties so the error wasn't shown in the tests of the original PR.

### DIFF
--- a/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
+++ b/packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
@@ -3,6 +3,7 @@ import {
   RenderTest,
   test,
   jitSuite,
+  GlimmerishComponent,
   tracked,
   defineComponent,
   trackedObj,
@@ -45,6 +46,58 @@ class HelperManagerTest extends RenderTest {
     };
 
     const Main = defineComponent({ hello }, '{{hello}}');
+
+    this.renderComponent(Main);
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+
+    this.rerender();
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+  }
+
+  @test({ skip: SKIP_DEFAULT_HELPER_MANAGER_TESTS })
+  '(Default Helper Manager) plain functions passed as component arguments work as helpers'(
+    assert: Assert
+  ) {
+    let count = 0;
+
+    const hello = () => {
+      count++;
+      return 'plain function';
+    };
+
+    const Main = defineComponent({}, '{{(@hello)}}');
+
+    this.renderComponent(Main, {
+      hello,
+    });
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+
+    this.rerender();
+
+    assert.equal(count, 1, 'rendered once');
+    this.assertHTML('plain function');
+  }
+
+  @test({ skip: SKIP_DEFAULT_HELPER_MANAGER_TESTS })
+  '(Default Helper Manager) plain functions stored on component class properties work as helpers'(
+    assert: Assert
+  ) {
+    let count = 0;
+
+    const Main = defineComponent({}, '{{(this.hello)}}', {
+      definition: class extends GlimmerishComponent {
+        hello = () => {
+          count++;
+          return 'plain function';
+        };
+      },
+    });
 
     this.renderComponent(Main);
 
--- a/packages/@glimmer/manager/lib/internal/index.ts
+++ b/packages/@glimmer/manager/lib/internal/index.ts
@@ -213,13 +213,36 @@ export function getInternalComponentManager(
 ///////////
 
 export function hasInternalComponentManager(definition: object): boolean {
-  return getManager(COMPONENT_MANAGERS, definition) !== undefined;
+  return (
+    hasDefaultComponentManager(definition) ||
+    getManager(COMPONENT_MANAGERS, definition) !== undefined
+  );
 }
 
 export function hasInternalHelperManager(definition: object): boolean {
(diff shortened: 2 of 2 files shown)


Windvis:
I'm not sure what the performance check faillure is about. :thinking:

packages/@glimmer/integration-tests/test/managers/helper-manager-test.ts
```diff
@@ -57,6 +58,58 @@ class HelperManagerTest extends RenderTest {
     this.assertHTML('plain function');
   }
 
+  @test({ skip: SKIP_DEFAULT_HELPER_MANAGER_TESTS })
```
Windvis, line 61: These new tests fail without the change in the `hasInternalHelperManager` util.

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -217,7 +217,7 @@ export function hasInternalComponentManager(definition: object): boolean {
 }
 
 export function hasInternalHelperManager(definition: object): boolean {
-  return getManager(HELPER_MANAGERS, definition) !== undefined;
+  return Boolean(getInternalHelperManager(definition, true));
```
chancancode, line 220: Hm, I don't think we can do this exactly, for one thing `getInternalHelperManager` would throw errors which would be problematic. Maybe we add something like this:

```ts
function hasDefaultManager(map: WeakMap<object, object>, definition: object): boolean {
  if (FEATURE_DEFAULT_HELPER_MANAGER) {
    if (map === HELPER_MANAGERS) {
      return typeof object === 'function';
    }
  }
  
  return false;
}
```

Then each of these `has*` functions can be updated to `hasDefaultManager(..., definition) || getManager(..., definition) !== undefined` and we can update this when we add other default managers.

I think it is safe (and faster) to do the default check first in all cases since you can't "unset" the default manager as far as I know, so if it has a default manager, then regardless of whether it had a more specific one, the answer is still going to be true.

Does that seem good to you?

chancancode:
Thanks for catching this!

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -213,13 +213,32 @@ export function getInternalComponentManager(
...
-  return getManager(MODIFIER_MANAGERS, definition) !== undefined;
+  return (
+    hasDefaultManager(MODIFIER_MANAGERS, definition) ||
+    getManager(MODIFIER_MANAGERS, definition) !== undefined
+  );
+}
+
+function hasDefaultManager(managerMap: WeakMap<object, object>, definition: object): boolean {
+  if (FEATURE_DEFAULT_HELPER_MANAGER) {
+    if (managerMap === HELPER_MANAGERS) {
```
Windvis, line 238: @chancancode I implemented it as you suggested but the map is only needed to determine the type of object we're dealing with. Since the utils that call this function already know the type, I think adding 3 dedicated helpers (ex. `hasDefaultHelperManager`) or moving the logic into the `has*` utils itself might be "better"?

packages/@glimmer/manager/lib/internal/index.ts
Windvis: > Then each of these has* functions can be updated to ...

I wasn't sure if you meant that I should do that as part of this PR so I added it like that just in case. :smile:  Let me know if it should only be added to the helper version for now.

packages/@glimmer/manager/lib/internal/index.ts
chancancode: Yeah, splitting it into three sounds fine too, feel free to make the change 👍🏼 The other two can just be `return false` for now, but it will mark and obvious spot that we need to update when we implement something for the other two.

Having separate `hasDefault*Manager` functions shouldn't add much in terms of cost (probably will get inlined), and it seems clearer than doing everything in `hasInternal*Manager`.

chancancode:
The perf test also passed so I think it was just a random failure last time. I'll check back tomorrow to give you some time to make the change you wanted. If not I will probably merge it anyway since it can always be refactored later.

packages/@glimmer/manager/lib/internal/index.ts
```diff
@@ -213,13 +213,36 @@ export function getInternalComponentManager(
...
 }
 
 export function hasInternalModifierManager(definition: object): boolean {
-  return getManager(MODIFIER_MANAGERS, definition) !== undefined;
+  return (
+    hasDefaultModifierManager(definition) || getManager(MODIFIER_MANAGERS, definition) !== undefined
+  );
+}
+
+function hasDefaultComponentManager(_definition: object): boolean {
```
Windvis, line 234: VS Code suggested to underscore the argument name since it's not used yet. I'm not sure if it's actually needed though, the lint script doesn't show any warnings when I don't prefix it so it just seems a VS Code only thing?

chancancode: I think we usually do that anyway, seems fine to me

Windvis:
@chancancode Thank you for the quick reviewing!

chancancode:
@Windvis published v0.84.2 with this

PR #1390, opened 2022-04-27
Allow components to be named the same as void tags

NullVoxPopuli:
Closes: https://github.com/glimmerjs/glimmer-vm/pull/1388
Much thanks to @ef4 and @Windvis 

tl;dr: `<Link></Link>` should be an allowed component invocation.

## Why? ლ(ಥ Д ಥ )ლ

Background: https://github.com/embroider-build/embroider/issues/1176
(though, @runspired isn't the first to run in to this, I've noticed this issue myself for the longest time, and just had a hard time understanding what was going on, because I was new into my "how do I even debug the things"-with-embroider-journey :sweat_smile: 

This is caused by [this line](https://github.com/glimmerjs/glimmer-vm/blob/07db0655c8d4b5c71792bf49ba3b547c196e460a/packages/%40glimmer/syntax/lib/generation/printer.ts#L248).

Though, a [bit of concern](https://github.com/glimmerjs/glimmer-vm/pull/1388#issuecomment-1088667419)

> The printer doesn't know about components.

I think long-term this is something that will fizzle out as a class of problem, because of strict mode (but who knows how long it'll be until everything is strict mode).

## The drawbacks ┌༼▀̿ Ĺ̯▀̿༽┐

Anyone using `Link`, `LINK`, `LiNk`, `lINK`, `liNK`, `linK` as their style/asset loader in the their glimmer-parsed template will now _not_ be a void element.

_how many people are doing this tho_? :upside_down_face: 

## The importance (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

Not being able to have a `<Link>` component with embroider-using apps is a blocker for _many many many many_ apps / addons I maintain from being able to adopt embroider.


acorncom:
Per https://github.com/embroider-build/embroider/issues/1176#issuecomment-1088741861 it looks like there might be some other cases that might be worth testing / handling here?

NullVoxPopuli:
Makes sense to me! Updates pushed!

lifeart (approved):

packages/@glimmer/syntax/lib/generation/printer.ts
```diff
@@ -245,7 +245,7 @@ export default class Printer {
   }
 
   CloseElementNode(el: ASTv1.ElementNode): void {
-    if (el.selfClosing || voidMap[el.tag.toLowerCase()]) {
+    if (el.selfClosing || voidMap[el.tag]) {
```
ef4, line 248: I don't think this really matches Glimmer's rule about component naming, which are based solely on whether the first letter is capital. `<lINK>` is still a void tag.

Something like

```js
function isVoidTag(tag: string): boolean {
  return voidMap[tag.toLowerCase()] && tag[0].toLowerCase() === tag[0];
}

CloseElementNode(el: ASTv1.ElementNode): void {
     if (el.selfClosing || isVoidTag(el.tag)) {
```

NullVoxPopuli: it doesn't look like `<lINK>` is supported even before these changes -- errors occur elsewhere.
I do like this function tho, so I'll add this.

NullVoxPopuli: pushed

chancancode (approved):

Alonski:
Hey,
Has this been released in a version of Ember by chance?

Alonski:
Is it possible to also hotfix this for older versions as well? Specifically 3.28?

chriskrycho:
Almost certainly not. This is not a new bug, and it's definitely not a security bug (3.28 is in the security-only phase of its LTS cycle). I understand that it's annoying, and we will definitely get it onto 4.8 and *possibly* onto 4.4, but we will need to discuss even the 4.4 case. It would be one thing if this were a regression, but it is not, as far as I can tell, and we don't make a habit out of doing significant backports for general bug fixes, else we would end up having to do it for basically *every* bug fix.

Don't get me wrong: I am sympathetic and it is annoying! But there are also costs to a backport: if we mess something up, we then have to revert or try to fix *that*, which causes yet more churn and difficulty for our users! And: there are only so many hours in the day.

Alonski:
Yes makes sense. @bertdeblock do you think we can add a disclaimer for the lint rule or something? It's somewhat unusable right now no?

bertdeblock:
@Alonski To fix the lint rule, I think we just need a new release here and then update the `@glimmer/*` packages in `ember-template-recast`, specifically `@glimmer/syntax`. We could also bump `ember-template-recast` in `ember-template-lint` afterwards (+ release), to force the fix for end users.

Alonski:
Oh that makes sense. I was thinking we had to update Ember source as well but what you said is that we don't. Thanks!

bertdeblock:
Would it be possible for someone to already release this? Getting it in the appropriate Ember versions can still be done / discussed afterwards I believe? Having this available in a new release would already allow us to fix the issue over in `ember-template-lint` I think.

chriskrycho:
I pinged a couple folks with permissions in Discord but I believe one is on vacation and the other had a sick kid this week.

ef4:
Comment via @wycats: there may still be a remaining bug here because if the tokenizer is not scope-aware, not-capitalized cases could still get wrong handling:

```hbs
<Thing as |link|>
  <link></link>
</Thing>
```

PR #1399, opened 2022-11-20
Chore: upgrade to Yarn 3

chriskrycho:
## What it does

- bump the pinned version
- gitignore the cache (no 'zero-install' at least now)
- update the lock file
- (automatically) sort all the `package.json` dependencies

## Notes

The other good move here would be to switch to pnpm, but this is a good first move regardless: Yarn v1 is unmaintained, and Volta only currently has preview support for pnpm but has full support for Yarn v3.


wycats:
Superseded by #1418, which migrated to pnpm.

PR #1400, opened 2022-11-20
Update to TS 4.9

chriskrycho:
> **Note** This currently sits on top of #1399 for convenience, though it does not strictly depend on it; I can rebase this if we choose not to land that one.

---

Update to TS 4.9

- Update TypeScript itself to 4.9, and make the following changes to the code to handle the significant improvements to strictness which landed during the 4.x series:

  - improvements to `Promise` type inference
  - switching the default type of a `catch` argument to `unknown`
  - general improvements to strictness, where TS falls back to `any` in fewer places
  - a change where TS rejects direct abstract class construction
  - changing to default to `unknown` rather than `{}` in unconstrained type parameters

- Update broccoli-typescript-compiler, which means we no longer need to provide `resolutions` for TypeScript since it's now properly treated as a peer dependency by broccoli-typescript-compiler.


wycats:
Superseded by #1418, which updates the codebase to TS 5.0.4.

PR #1406, opened 2023-01-18
Fix getTemplateLocals for yielded block params

simonihmig:
Given this template:

```hbs
<template>
  <HeadlessForm @data={{data}} as |form|>
    <form.field @name="firstName" as |field|>
      <field.label>First Name</field.label>
      <field.input data-test-first-name />
    </form.field>
    <form.field @name="lastName" as |field|>
      <field.label>Last Name</field.label>
      <field.input data-test-last-name />
    </form.field>
  </HeadlessForm>
</template>
  ```

... `eslint-plugin-ember` would complain about `form` and `field` violating `no-undef` (see https://github.com/ember-cli/eslint-plugin-ember/issues/1747). The root cause turned out to be that the scope function (after template-imports have been compiled away) would _falsely_ include those block params (like `scope: () => ['HeadlessForm', 'data', 'form', 'field']`), which are only local to the inner scopes of the template, not the outer/JS scope.

This PR should fix the root cause of those wrong locals returned by `getTemplateLocals()`.

### DIFF
--- a/packages/@glimmer/syntax/test/template-locals-test.ts
+++ b/packages/@glimmer/syntax/test/template-locals-test.ts
@@ -35,6 +35,13 @@ QUnit.test('it works', function (assert) {
 
     <this.dynamicAngleComponent>
     </this.dynamicAngleComponent>
+
+    <ComponentYieldingContextual as |hash|>
+      <hash.some as |some|>
+        <some.other/>
+        {{some.value}}
+      </hash.some>
+    </ComponentYieldingContextual>
   `);
 
   assert.deepEqual(locals, [
@@ -45,6 +52,7 @@ QUnit.test('it works', function (assert) {
     'global-value',
     'some',
     'someOther',
+    'ComponentYieldingContextual',
   ]);
 });
 
--- a/packages/@glimmer/syntax/test/template-locals-test.ts
+++ b/packages/@glimmer/syntax/test/template-locals-test.ts
@@ -37,7 +37,9 @@ QUnit.test('it works', function (assert) {
     </this.dynamicAngleComponent>
 
     <ComponentYieldingContextual as |hash|>
+      <InsideHash />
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/@glimmer/syntax/test/template-locals-test.ts [hunk 1]
```diff
@@ -35,6 +35,13 @@ QUnit.test('it works', function (assert) {
 
     <this.dynamicAngleComponent>
     </this.dynamicAngleComponent>
+
+    <ComponentYieldingContextual as |hash|>
+      <hash.some as |some|>
+        <some.other/>
```
chriskrycho, line 41: While we're here, let's add another couple that *should* be:

```suggestion
      <InsideHash />
      <hash.some as |some|>
        {{inside-another-hash}}
        <some.other/>
```

Then we can include these in the locals as well. Basically: I want to make sure that our block-yielding behavior handles *both* the yielded content and the regular content (and that it doesn't regret).

packages/@glimmer/syntax/test/template-locals-test.ts
```diff
@@ -45,6 +52,7 @@ QUnit.test('it works', function (assert) {
     'global-value',
     'some',
     'someOther',
+    'ComponentYieldingContextual',
```
chriskrycho, line 55:
```suggestion
    'ComponentYieldingContextual',
    'InsideHash',
    'inside-another-hash',
```

I *think* this is the order it should show up in ☝🏼

chriskrycho (changes requested):
This looks good, let's just add a tiny bit more coverage while we're mucking with this!

chriskrycho:
The failing CI run for floating deps is because of Node 12 having gone out of support for a bunch of things (we should, separately, see about updating that).

chriskrycho (approved):

Issue #1410, opened 2023-03-21
Values used as modifiers from nested function invocations need to have associateDestroyableChild called on them


NullVoxPopuli:
Reproduction here: https://github.com/ember-modifier/ember-modifier/issues/613#issuecomment-1468823388

The situation is when a modifier is conditionally applied:
```hbs
<div {{ (if condition someModifier) }}>
```
when condition becomes false, `someModifier` is not destroyed

PR #1411, opened 2023-03-21
Add more tests for dynamic modifier invocation

NullVoxPopuli:
While investigating https://github.com/glimmerjs/glimmer-vm/issues/1410
Which is https://github.com/ember-modifier/ember-modifier/issues/613
I found that glimmer-vm, at least, seems to be handling this scenario correctly.
So while I debug elsewhere, I figured it may be good to have the extra tests, since I didn't see anything covering the this use case.


wycats:
Modifier cleanup should definitely run reliably whenever their associated element is removed from the DOM.

PR #1414, opened 2023-03-27
Change volta and C.I. to use Node 14

NullVoxPopuli:
This fixes the "floating dependencies", as some of our dependencies have dropped support for node 12, so we should as well.
Node 12 is not supported: https://github.com/nodejs/release#release-schedule
Node 14 is EOL'd next month.


bertdeblock (approved):

NullVoxPopuli:
Now that https://github.com/glimmerjs/glimmer-vm/pull/1418 has landed, we can close this. :tada:

PR #1421, opened 2023-05-04
Improve lexical scope

wycats:
# Improved lexical scope

This is an initial implementation of an improved API for compiling a
template with lexical scope.

Instead of passing a list of locals to the precompileTemplate function,
you can pass a function that takes a variable name and returns a boolean
to indicate whether the variable is in scope.

## Details

This fixes an inconsistency with how HTML tags and keywords
behave when they refer to a Handlebars local variable vs. when they
refer to a JavaScript lexical variable.

```jsx
const component = <template><div>{{yield}}</div></template>;

<template>
  <component />
</template>
```

After this change, the `<component>` invocation correctly refers to the
surrounding scope variable, even though `component` is a keyword.

This makes it consistent with:

```jsx
const variable = <template><div>{{yield}}</div></template>;

<template>
{{#let variable as |component|}}
  <component />
{{/let}}
</template>
```

These semantics are an important future-proofing measure: they allow us
to add additional keywords in the future, since local variables (either
lexical JavaScript variables of Handlebars variables) will win over the
newly defined keywords.

If we allowed keywords to win over lexical JavaScript variables, every
new keyword would be a breaking change. The semantics we used for
`as |div|` and `as |component|` were carefully designed to avoid this
problem, and this change helps us implement the same behavior for
lexical JavaScript variables.

### DIFF
--- a/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
+++ b/packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
@@ -1,7 +1,7 @@
 import { precompile } from '@glimmer/compiler';
 import { preprocess } from '../..';
 import { module } from '../support';
-import { unwrapTemplate, assert as glimmerAssert } from '@glimmer/util';
+import { unwrapTemplate, assert as glimmerAssert, expect } from '@glimmer/util';
 import { SexpOpcodes, WireFormat } from '@glimmer/interfaces';
 import { TemplateWithIdAndReferrer } from '@glimmer/opcode-compiler';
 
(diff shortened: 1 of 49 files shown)


packages/@glimmer/integration-tests/test/compiler/compile-options-test.ts
```diff
@@ -26,6 +26,40 @@ module('[glimmer-compiler] precompile', ({ test }) => {
     assert.strictEqual(wire.moduleName, 'my/module-name', 'Template has correct meta');
   });
 
+  test('lexicalScope is used if present', (assert) => {
```
NullVoxPopuli, line 29: can we add a test (or does one already exist somewhere else?) for `<x.someProperty />`?

ef4:
As soon as this is in a release I'm ready to use it.

NullVoxPopuli:
should be in 0.85.3

PR #1423, opened 2023-05-07
Migrate to Vite

NullVoxPopuli:
wip -- seeing where CI is at

The main thing blocking this work is that we need a CLI tool listen to qunit as it runs tests.

Options for inspiration / potential integration:
 - https://github.com/larsthorup/mocha-vite-puppeteer/blob/main/cli.js (the very manual approach, but gives us the most control)
   - tl;dr: 
     - use puppeteer for managing the browser launching
       - could also use a more dedicated browser launcher, such as: https://www.npmjs.com/package/@httptoolkit/browser-launcher
     - wire puppeteer to a custom low-level vite API, createServer
     - use custom endpoint / server-API for handling QUnit sending messages of test status back to the CLI
   - **this may be the best long-term option -- we could make a vite-qunit plugin for vite** 
 - [testcafe](https://testcafe.io/) ? might be possible, but their docs don't make finding where the integrations points would be easily found
 - other interesting things
   - https://github.com/bahmutov/start-server-and-test
   - https://modern-web.dev/docs/test-runner/test-frameworks/write-your-own/
     - https://modern-web.dev/docs/test-runner/browser-launchers/overview/#configuring-browser-launchers 
     - I don't think this allows for connecting to a vite server -- it could load dist-output, but that leads to slower dev-feedback loop
   - https://node-tap.org/docs/reporting/


wycats:
I got a decent setup going with puppeteer, but it doesn't (yet) do type tests, browserstack tests or perf tests. These are high-priority follow-up improvements.

PR #1430, opened 2023-05-26
Fix source checking when parsing {{~!~}} and the like.

navels:
This fixes a warning that derives from several of these statements in `ember-prism`, which are intended to strip whitespace without otherwise doing anything:
```
{{~! ~}}
```
Presently these result in a compiler warning when building `ember-prism` or any addon which consumes it:
```
unexpectedly found "! ~" when slicing source, but expected " "
unexpectedly found "! ~" when slicing source, but expected " "
unexpectedly found "! ~" when slicing source, but expected " "
unexpectedly found "! ~" when slicing source, but expected " "
```

Fixes one cause of https://github.com/emberjs/ember.js/issues/19392

When fixing this I decided some tests would be nice to add even though they don't cover this fix.  Not really thinking it is worth a test specifically targeting this case, though.

### DIFF
--- a/packages/@glimmer/compiler/test/compiler-test.ts
+++ b/packages/@glimmer/compiler/test/compiler-test.ts
@@ -317,6 +317,8 @@ test('top-level comments', `<!-- {{foo}} -->`, c` {{foo}} `);
 
 test('handlebars comments', `<div>{{! Better not break! }}content</div>`, ['<div>', [s`content`]]);
 
+test('handlebars comments with whitespace removal', '<div>  {{~! some comment ~}}  content</div>', ['<div>', [s`content`]]);
+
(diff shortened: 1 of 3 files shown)


acorncom:
I've been helping a client with some upgrades recently and found this message went away once the app was upgraded to 5.12. Probably can close now?

boris-petrov:
I am on 5.12 and I still very much have these messages.

boris-petrov:
@NullVoxPopuli, @wycats - could you please take a look at this PR if it really fixes the issue at hand? And also [the other one](https://github.com/glimmerjs/glimmer-vm/pull/1431)? This has been an issue for quite a while now...

navels:
Covered by #1717

PR #1431, opened 2023-05-27
Skip checking the source for synthetic attributes to prevent warning.

navels:
This fixes a warning caused when template parsing encounters a synthetic attribute (i.e., inserted by a plugin).  For example, building the `ember-maybe-in-element` addon outputs these warnings:

```
unexpectedly found "<div id=\"original-p" when slicing source, but expected "@destinationElement"
unexpectedly found "<div id=\"origi" when slicing source, but expected "@renderInPlace"
unexpectedly found "    \n      <div id=" when slicing source, but expected "@destinationElement"
unexpectedly found "    \n      <di" when slicing source, but expected "@renderInPlace"
```

and the same will happen for any app that uses the `ember-basic-dropdown` addon (which uses the above addon).

These should fix the remaining causes of https://github.com/emberjs/ember.js/issues/19392 that are not already fixed by https://github.com/glimmerjs/glimmer-vm/pull/1430.

I tried to add a test that would be even vaguely useful and couldn't really come up with anything.

### DIFF
--- a/packages/@glimmer/syntax/lib/v2/normalize.ts
+++ b/packages/@glimmer/syntax/lib/v2/normalize.ts
@@ -8,6 +8,7 @@ import {
   preprocess,
 } from '../parser/tokenizer-event-handlers';
 import type { SourceLocation } from '../source/location';
+import { SYNTHETIC_LOCATION } from '../source/location';
 import { SourceSlice } from '../source/slice';
 import type { Source } from '../source/source';
 import type { SourceSpan } from '../source/span';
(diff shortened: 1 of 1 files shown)


acorncom:
I've been helping a client with some upgrades recently and found this message went away once the app was upgraded to 5.12. Probably can close now? You can see that other folks handled this by patching temporarily using `pnpm patch` options

navels:
Might want to leave this open based on this comment: https://github.com/glimmerjs/glimmer-vm/pull/1430#issuecomment-2473793233

acorncom:
@navels thanks, will do

PR #1441, opened 2023-09-01
expose visitor-keys for eslint plugin 

### LINKED ISSUE #1436: expose visitor keys in index
So we can import them normally.
Now i have to do something like this: 
https://github.com/ember-cli/eslint-plugin-ember/blob/050185cec885f087c0fb27f2efce04d949f428f1/lib/parsers/gjs-parser.js#L6

patricklx:
resolves #1436


patricklx:
fixed by #1467

PR #1447, opened 2023-10-12
fix: capture render tree can fail if args cannot be obtained

patricklx:
this currently happens on https://meta.discourse.org/ when checking with ember-inspector and not being logged in.

### DIFF
--- a/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
+++ b/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
@@ -138,19 +138,29 @@ class DebugRenderTreeTest extends RenderTest {
 
   @test 'emberish curly components'() {
     this.registerComponent('Curly', 'HelloWorld', 'Hello World');
+    let error: Error|null = null;
+    const obj = {
+      get getterWithError() {
+        error = new Error('error');
+        throw error;
+      }
+    }
 
     this.render(
-      `<HelloWorld @arg="first"/>{{#if this.showSecond}}<HelloWorld @arg="second"/>{{/if}}`,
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
hello! can a test be added? thanks!

patricklx:
Hi, where can I add that?

NullVoxPopuli:
maybe in here? https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer-workspace/integration-tests/test/debug-render-tree-test.ts

patricklx:
added the tests for this.
Could this also be backported?

patricklx:
@NullVoxPopuli

NullVoxPopuli:
unsure, what version would you want to backport to?

NullVoxPopuli (approved):

NullVoxPopuli:
for  others, a repro of the problem: https://stackblitz.com/edit/github-pgxsyq?file=app%2Fcomponents%2Fwelcome.gts

(open the ember inspector, ensure the `throw` is happening, and see that the component view is blank)

NullVoxPopuli:
glimmer-vm doesn't have an LTS tho :sweat_smile: 

an issue is that even if we fix ember-LTS, all the versions in between won't be fixed, because glimmer-vm is bundled with ember-source right now.

patricklx:
i think we need to revert this.
it affects also other parts of glimmer...

NullVoxPopuli:
How do you mean? Can we add more tests to prevent whatever it in you've discovered?

NullVoxPopuli:
Looks like #1460 expnains

patricklx:
for example here it is used: https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%40glimmer/runtime/lib/helpers/fn.ts#L87.
in application this change would have caught errors, and the developer would not see it in the app...

NullVoxPopuli:
Oofta, that tells me we don't have sufficient failure-case testing

PR #1450, opened 2023-10-23
Use workspace protocol everywhere so release-it doesn't cause a diff,…

NullVoxPopuli:
… breaking C.I.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1143,10 +1143,10 @@ importers:
   packages/@glimmer/wire-format:
     dependencies:
       '@glimmer/interfaces':
-        specifier: 0.84.3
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1143,10 +1143,10 @@ importers:
   packages/@glimmer/wire-format:
     dependencies:
       '@glimmer/interfaces':
-        specifier: 0.84.3
```
NullVoxPopuli, line 1146: these changing due to release-it bumps is what caused CI to fail on `master` / the last tag (v0.85.0),

hopefully with some followup here: https://github.com/release-it-plugins/workspaces/pull/110#discussion_r1369310157

we can get something more robust implemented

PR #1453, opened 2023-10-23
Setup dispatch / manual re-running of release workflow

NullVoxPopuli:
So I can stop publishing basically no-op tags to try a new release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,12 @@
 name: Release
 
 on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        required: true
+        type: string
+        description: 'Which tag do we want to retry? this should begin with v, and match exactly one of the tags present on the repo'
   push:
     tags:
       - 'v*'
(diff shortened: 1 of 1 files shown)


chancancode (approved):

.github/workflows/release.yml
```diff
@@ -12,6 +18,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.events.inputs.tag || github.ref }}
```
NullVoxPopuli, line 22: ref is:
> 

  | The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed. For workflows triggered by pull_request, this is the pull request merge branch. For workflows triggered by release,  this is the release tag created. For other triggers, this is the branch  or tag ref that triggered the workflow run. This is only set if a  branch or tag is available for the event type. The ref given is  fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1.
-- | --


from: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

PR #1459, opened 2023-10-24
Revert "fix: capture render tree can fail if args cannot be obtained"

patricklx:
Reverts glimmerjs/glimmer-vm#1447

this also affects other parts, unrelated to capture render tree


NullVoxPopuli:
Thank you!

PR #1460, opened 2023-10-24
fix capture render tree fails when errors in args

patricklx:
better alternative to #1447  .

i''m actually not sure anymore how [meta.discource.](https://meta.discourse.org/) was in a state where it could render the components, but would fail when using inspector.
having errors in the args would cause rendering to fail as well.

### DIFF
--- a/packages/@glimmer/interfaces/lib/runtime/arguments.d.ts
+++ b/packages/@glimmer/interfaces/lib/runtime/arguments.d.ts
@@ -60,3 +60,9 @@ export interface Arguments {
   positional: readonly unknown[];
   named: Record<string, unknown>;
 }
+
+export interface ArgumentsDebug {
+  positional: readonly unknown[];
+  named: Record<string, unknown>;
+  errors: Record<string, Error>;
+}
--- a/packages/@glimmer/runtime/lib/debug-render-tree.ts
+++ b/packages/@glimmer/runtime/lib/debug-render-tree.ts
@@ -7,7 +7,7 @@ import type {
 } from "@glimmer/interfaces";
 import { assign, expect, Stack } from '@glimmer/util';
 
-import { reifyArgs } from './vm/arguments';
+import { reifyArgsDebug } from './vm/arguments';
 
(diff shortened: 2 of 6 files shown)


packages/@glimmer/vm-babel-plugins/index.ts
```diff
@@ -21,7 +21,7 @@ export default function generateVmPlugins(
   return [
     [
       __loadPlugins
-        ?
```
patricklx, line 24: auto save... 
well lets cleanup this space

packages/@glimmer/runtime/lib/debug-render-tree.ts
```diff
@@ -181,7 +181,7 @@ export default class DebugRenderTreeImpl<TBucket extends object>
     let template = this.captureTemplate(node);
     let bounds = this.captureBounds(node);
     let children = this.captureRefs(refs);
-    return { id, type, name, args: reifyArgs(args), instance, template, bounds, children };
+    return { id, type, name, args: reifyArgsDebug(args), instance, template, bounds, children };
```
NullVoxPopuli, line 184: Does this always happen? Do we want this to occur un production or dev only?

patricklx: i think its should be for both

packages/@glimmer/interfaces/lib/runtime/arguments.d.ts
```diff
@@ -58,5 +58,6 @@ export interface CapturedNamedArguments extends Record<string, Reference> {
 
 export interface Arguments {
   positional: readonly unknown[];
+  errors: Record<string, Error>;
```
NullVoxPopuli, line 61: Where all is this interface used? If it's public, i'm worried about adding to it.

If it's internal, than it's fine

patricklx: it would be used for the inspector, to distinguish if an arg throws an error, or if it return an Error.
as it could also just be a `return new Error()`

NullVoxPopuli (approved):

PR #1461, opened 2023-10-24
Declare @glimmer/local-debug-flags

NullVoxPopuli:
Ran in to this while trying to upgrade the vm in ember: https://github.com/emberjs/ember.js/pull/20561

We can't import packages without declaring them in `dependencies` or `peerDependencies`. Since the local-debug-flags package is entirely wrapped in `import.meta.env` guards, then we can include it and rely on build to remove / svelte away the unused blocks.


NullVoxPopuli:
I think I'm going to take a different approach with this actually.

gonna add a custom plugin to rollup to remove references to this package

PR #1467, opened 2023-10-27
visitorKeys is needed for prettier

NullVoxPopuli:
Resolves: https://github.com/glimmerjs/glimmer-vm/issues/1465 and https://github.com/glimmerjs/glimmer-vm/issues/1436

PR #1468, opened 2023-10-27
Use a Set for checking if an element is a void element, which is faster than object access

NullVoxPopuli:
When it comes to raw speed, for small datasets, Map.get is faster than Object[..] is faster than Set.has

[benchmark - Map#get - key exists](https://perf.link/#eyJpZCI6InJ4YmdkMGthcWtmIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHZvaWRNYXAgPSBPYmplY3QuY3JlYXRlKG51bGwpO1xuXG5jb25zdCB2b2lkVGFnTmFtZXMgPSBbJ2FyZWEnLCAnYmFzZScsICdicicsICdjb2wnLCAnY29tbWFuZCcsICdlbWJlZCcsICdocicsICdpbWcnLCAnaW5wdXQnLCAna2V5Z2VuJywgJ2xpbmsnLCAnbWV0YScsICdwYXJhbScsICdzb3VyY2UnLCAndHJhY2snLCAnd2JyJ107XG5cbnZvaWRUYWdOYW1lcy5mb3JFYWNoKCh0YWdOYW1lKSA9PiB7XG4gIHZvaWRNYXBbdGFnTmFtZV0gPSB0cnVlO1xufSk7XG5cbmNvbnN0IHZvaWRUYWdBcnJheSA9IE9iamVjdC5rZXlzKHZvaWRNYXApO1xuY29uc3Qgdm9pZFRhZ1NldCA9IG5ldyBTZXQodm9pZFRhZ0FycmF5KTtcbmNvbnN0IHZvaWRUYWdNYXAgPSBuZXcgTWFwKE9iamVjdC5lbnRyaWVzKHZvaWRNYXApKTsiLCJ0ZXN0cyI6W3sibmFtZSI6Ik1hcCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdNYXAuZ2V0KCdsaW5rJykpIHtcblxufSIsInJ1bnMiOls0NzYzMDAwLDY0ODgwMDAsMjQ4MTUwMDAsNDY3NjAwMCwxMTU4MDAwLDEzMjcwMDAsMzY1MzAwMCwzNzgwMDAwLDIzODEwMDAsNDU2NzAwMCwxMTMwNDAwMCwxNDc1MDAwMCw1Mjg0MDAwLDc3NTEwMDAsMjA2ODEwMDAsNTYzNDAwMCwyMzI1ODAwMCwxNTE2MDAwLDYxNDkwMDAsODAzMjAwMCwxOTAxNTAwMCwxNDkyMDAwLDEyODA1MDAwLDY0MTMwMDAsNjM5NjAwMCwxNTY5ODAwMCwzMjI4MDAwLDY5NjkwMDAsMTQ4NzIwMDAsMTI5NTIwMDAsMTY3MzQwMDAsNDU4NTAwMCwxNzIwMzAwMCwxNzQwMDAwLDE1NjExMDAwLDU2ODEwMDAsMTM4MjAwMCwxMjA1MjAwMCw4MTIwMDAwLDc3ODIwMDAsNjI4MDAwMCwyMDQ4NzAwMCwxMTIwNzAwMCwxODMyNTAwMCw2OTI4MDAwLDExMDczMDAwLDIxMjIzMDAwLDMzNTMwMDAsMTAwMTgwMDAsMTkzNTcwMDAsMzA4NTAwMCwxMTY1NzAwMCw3Njg5MDAwLDU4NjQwMDAsMTA5NTgwMDAsNzA4ODAwMCw0MTIzMDAwLDE5OTA2MDAwLDc2MDAwMCwxOTY0MTAwMCwzNTQxMDAwLDMwMTAwMDAsOTA5MDAwLDExNTUzMDAwLDEyMTYwMDAwLDE2OTcwMDAsODUwNTAwMCw1MDEwMDAwLDU2OTcwMDAsNTg0ODAwMCw2NzU1MDAwLDM4ODYwMDAsNjQwNjAwMCw4NzI4MDAwLDc1MjYwMDAsMTU3OTQwMDAsNjQxNjAwMCwxODU0MTAwMCw2NTU2MDAwLDcxMjUwMDAsMTA1OTQwMDAsMTkwOTEwMDAsMjUwODMwMDAsNzQzNzAwMCwyMjU0MTAwMCw0MDAwMCwyNDE2NDAwMCwyNjkzMzAwMCwyMzQ1MzAwMCwyNTE5MDAwLDE0MDI3MDAwLDEzODIwMDAsMTQyNTQwMDAsMTY0MTQwMDAsOTAyMDAwMCwxMDY1NDAwMCwyNDk2MDAwLDQwMjYwMDAsMzUyNzAwMCw1NzMxMDAwXSwib3BzIjo5NjA3NjgwfSx7Im5hbWUiOiJvYmplY3QgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkTWFwWydsaW5rJ10pIHtcblxufSIsInJ1bnMiOls3MTE0MDAwLDIwMzYyMDAwLDIxOTg2MDAwLDUzNTYwMDAsMTQ2MTAwMDAsNzU5NDAwMCwxNDk1NDAwMCwzMzU2MDAwLDE4MTM3MDAwLDIwMDAwMCwyOTAyMDAwLDQ2MzEwMDAsMzQzNTAwMCw0MDg2MDAwLDExMTA0MDAwLDM3MDAwLDExMzUxMDAwLDIyNjQxMDAwLDIzMzYwMDAsMTEzODEwMDAsNzcxODAwMCwxMTU2ODAwMCw2NDg1MDAwLDI2NzMwMDAsMzI1OTAwMCwxNDQ1MzAwMCwzMDE1MDAwLDE2MTgzMDAwLDY5NjMwMDAsNzQ1MjAwMCw3MDE5MDAwLDE1NDQwMDAsOTc3NDAwMCwxNzQwMDAsMTAzMTgwMDAsMjE1MTEwMDAsMTY2MDkwMDAsNzQzMjAwMCwxNTQ0MDAwLDM2OTcwMDAsMzgzMjAwMCw4MTU1MDAwLDI5MzEwMDAsMTEzODQwMDAsMzA2NzAwMCwyNjM4MDAwLDEyMzQ0MDAwLDE2NTgwMDAsNDcxNzAwMCwxMTMxNjAwMCwxMDEyMDAwLDY0MjMwMDAsNDUyNTAwMCwxODA0MDAwLDMwNzcwMDAsNDE0MTAwMCwzNjQ5MDAwLDE0Njk3MDAwLDczNTAwMCwxMjQ0NTAwMCwxNTQ0MDAwLDMwMTUwMDAsMTUyMDEwMDAsMTc2NzAwMCwxNjAwMDAwLDE4MzI5MDAwLDcyMTcwMDAsMzU4MDAwMCwxNTQ0MDAwLDQ1MjQwMDAsNTMzMzAwMCwzMTE3MDAwLDU0OTcwMDAsNDM3NTAwMCwzOTM1MDAwLDQ4NDcwMDAsMzU3ODAwMCwzNDIxMDAwLDMzMjYwMDAsMjI5OTAwMCwxMjA5MzAwMCwxNDY3NTAwMCwxODM3MDAwMCwzMDUxMDAwLDQ4ODMwMDAsMTkwOTAwMCwxMzA1MDAwMCwxMTIxMDAwMCwyMzgwNDAwMCwxNTE1MDAwLDEyNjI1MDAwLDIwOTcwMDAsNTY5OTAwMCwxNDA0MTAwMCwyMDAwMDAwLDQwMTMwMDAsMTA1NTAwMCwyMTA1MDAwLDIyOTgwMDAsNjk1MDAwXSwib3BzIjo3MTA3NTEwfSx7Im5hbWUiOiJzZXQgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkVGFnU2V0LmhhcygnbGluaycpKSB7XG5cbn0iLCJydW5zIjpbMTA5NjAwMCwxMzgyMDAwLDEwMTY4MDAwLDE4OTAwMDAsMTcyNDAwMCw1Mzg5MDAwLDExMjQ3MDAwLDEyNjM2MDAwLDEzOTQwMDAsMTExOTIwMDAsMjc0NjUwMDAsMjk3NzAwMCwxNDI2MjAwMCw2NDgwMDAsMTU5MjIwMDAsNzQ4MzAwMCw2NjU5MDAwLDE4MDg4MDAwLDkwNjIwMDAsNDIxNDAwMCwyMDM5MDAwMCwyODA3MDAwLDExNDMwMDAsMTM2MDAwMCw2MjIwMDAsMTIwNjIwMDAsOTY0MDAwLDE4Mjg5MDAwLDI2NzEwMDAsNzUwODAwMCwxOTg4MDAwLDk1NzYwMDAsMTgzODAwMCwxMjkzMjAwMCwxNTg2OTAwMCwxMzg5MjAwMCw1NDA5MDAwLDE0MDAwMCw5MzEwMDAwLDcyMTAwMCwyMDA0NzAwMCwxOTU0MDAwMCwxNTE3NjAwMCwyMTc0MDAwLDE5NzU4MDAwLDIwOTU0MDAwLDI5ODkwMDAsMTM4MjAwMCwyMDgyMDAwLDY1NjYwMDAsMTIxNTMwMDAsMTEyMTAwMDAsMzMzMDAwLDE0MDM0MDAwLDE5MjkzMDAwLDEzODIwMDAsMTcwOTEwMDAsMjc1ODAwMCwxMzgyMDAwLDEzODIwMDAsMTYwNjAwMCwxMzgyMDAwLDE1OTkxMDAwLDE0NjU5MDAwLDI4MTAwMCwxMTg1MTAwMCwzMzQxMDAwLDc3MTAwMCw4NTEwMDAsMTEwNTAwMCwxODIwMDAsMTA0ODAwMCw5NTIwMDAsMTMxNjAwMCwxODU4MjAwMCwxMzgyMDAwLDgwMDAwLDEzNTgzMDAwLDIzMDAwMDAsMTQ2MDYwMDAsMTMwODIwMDAsNDU3MTAwMCwxMDYwMzAwMCwxOTU1ODAwMCw1MDQ3MDAwLDEzNjA2MDAwLDYwMTUwMDAsNTA0ODAwMCwyMDYzMDAwMCwxOTY3MzAwMCw5MDM3MDAwLDIwNzgzMDAwLDEzNjE0MDAwLDY1MzMwMDAsNjE3MDAwLDE2NDYwMDAsMTM4MjAwMCwxMzgyMDAwLDEyMDQwMDAsOTM2NzAwMF0sIm9wcyI6Nzg1MzQyMH1dLCJ1cGRhdGVkIjoiMjAyMy0xMC0yN1QxNzowNzoxNC43NTRaIn0%3D)

[benchmark - Map#get - key does not exist](https://perf.link/#eyJpZCI6InJ4YmdkMGthcWtmIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHZvaWRNYXAgPSBPYmplY3QuY3JlYXRlKG51bGwpO1xuXG5jb25zdCB2b2lkVGFnTmFtZXMgPSBbJ2FyZWEnLCAnYmFzZScsICdicicsICdjb2wnLCAnY29tbWFuZCcsICdlbWJlZCcsICdocicsICdpbWcnLCAnaW5wdXQnLCAna2V5Z2VuJywgJ2xpbmsnLCAnbWV0YScsICdwYXJhbScsICdzb3VyY2UnLCAndHJhY2snLCAnd2JyJ107XG5cbnZvaWRUYWdOYW1lcy5mb3JFYWNoKCh0YWdOYW1lKSA9PiB7XG4gIHZvaWRNYXBbdGFnTmFtZV0gPSB0cnVlO1xufSk7XG5cbmNvbnN0IHZvaWRUYWdBcnJheSA9IE9iamVjdC5rZXlzKHZvaWRNYXApO1xuY29uc3Qgdm9pZFRhZ1NldCA9IG5ldyBTZXQodm9pZFRhZ0FycmF5KTtcbmNvbnN0IHZvaWRUYWdNYXAgPSBuZXcgTWFwKE9iamVjdC5lbnRyaWVzKHZvaWRNYXApKTsiLCJ0ZXN0cyI6W3sibmFtZSI6Ik1hcCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdNYXAuZ2V0KCdkb2VzLW5vdC1leGl0JykpIHtcblxufSIsInJ1bnMiOls1NjY1MDAwLDE4NjAwMDAsMjEwNTAwMCw0MDMwMDAsNjE4NDAwMCw3MDIwMDAwLDEwMzE4MDAwLDEwOTYxMDAwLDc2MTUwMDAsOTMxNTAwMCwxMTEzNjAwMCw3Mjc2MDAwLDE1NDQ1MDAwLDYzNTQwMDAsMTY3MzIwMDAsNzQ2NzAwMCwxMjkyNDAwMCw0OTk2MDAwLDI0ODEwMDAsMTYwNzAwMCwyMTczMDAwMCwxNDQzNzAwMCwyMTQzMDAwLDE2NzkwMDAwLDI5MjkwMDAsMjg4NDAwMCwxMzg2ODAwMCw5OTU1MDAwLDYwODIwMDAsNjc5MDAwLDI1MjAwMCwxMjAwNDAwMCwzOTAwMDAwLDEwNDExMDAwLDE1MDI1MDAwLDkwMDAsMjM5NzAwMCwxMjA4MDAwMCw0NTAwMDAwLDE2Njk0MDAwLDExNzUwMDAsMTcxNTAwMDAsMTkyNTAwMCwxMjI0MjAwMCw5ODAwMCwxMTgyNDAwMCwyNDExMDAwLDEwMzYzMDAwLDk2NzEwMDAsMTEyNzAwMDAsMTYyODQwMDAsNzM2MDAwLDQ5OTgwMDAsMTYwMDAwLDEyMjA1MDAwLDEwMDAwLDYwNzYwMDAsMTM5OTMwMDAsNzkxNTAwMCw1OTU4MDAwLDc2MDIwMDAsMTI1OTMwMDAsNTQ2MTAwMCwxNDI0MTAwMCw3OTAxMDAwLDE0MTE2MDAwLDEzNzY2MDAwLDg4ODAwMCw0MzU0MDAwLDE3OTEwMDAsMTg1ODcwMDAsODMyMDAwLDE0MjM1MDAwLDE1ODAyMDAwLDExODkxMDAwLDE4MTMwMDAsMTA1MzMwMDAsNjg5MDAwLDEyMDYyMDAwLDYzOTAwMDAsMTI2OTQwMDAsMTMyMzAwMCw3NjQ4MDAwLDE2NDYwMDAsOTQ3OTAwMCwxMzg1MDAwMCwxMzk3MjAwMCw5OTk4MDAwLDExMTcwMDAsMTI0MDAwLDkyNjUwMDAsNTIxMzAwMCw1NDkyMDAwLDEyMDAwLDExNzEzMDAwLDk4MDgwMDAsMTMxMDIwMDAsMTIzNTAwMCw1NjYwMDAwLDEzMDA5MDAwXSwib3BzIjo3ODUwMDQwfSx7Im5hbWUiOiJvYmplY3QgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkTWFwWydkb2VzLW5vdC1leGl0J10pIHtcblxufSIsInJ1bnMiOlszNzk0MDAwLDg0NDQwMDAsMTIxMjcwMDAsMzY3MDAwMCwxMDY5NzAwMCw1NDM4MDAwLDEyODEzMDAwLDk5ODUwMDAsNjc0OTAwMCw5ODE5MDAwLDE2MzU0MDAwLDUwNjcwMDAsMTMyMjIwMDAsNzc1MTAwMCwxMjAyODAwMCw1OTE0MDAwLDE1MjM1MDAwLDE4NDQzMDAwLDE1Nzc4MDAwLDU0MDAwMCw3MTkzMDAwLDgzNTMwMDAsNzU0MDAwLDEyMDk1MDAwLDk4MDAwLDY1NzAwMDAsOTcyMTAwMCwyNDg5MDAwLDI5ODYwMDAsMTY1NDQwMDAsOTgwMDAsMTU5MjIwMDAsMzcyNTAwMCw2MjI1MDAwLDE1MTgxMDAwLDEzNzI4MDAwLDE5MDMwMDAsNjI3ODAwMCwzNTA1MDAwLDEyNzM0MDAwLDMyMjkwMDAsMTY5NTUwMDAsMTc5MDAwMCwxMDUzOTAwMCwxNTk1NTAwMCw5ODY3MDAwLDEwMjUwMDAwLDQxMTEwMDAsODQzNTAwMCw5NDk0MDAwLDEwMjA3MDAwLDE4MTgwMDAwLDU2MDUwMDAsOTE2MDAwLDkwMTkwMDAsMTQ3MzQwMDAsMTI1NDIwMDAsMTM1NDgwMDAsODAzMDAwMCwzMzQ4MDAwLDE3MDU3MDAwLDEwMjI4MDAwLDczNDYwMDAsMTUyMzYwMDAsMjMzNDAwMCwxMzQ4NzAwMCw1NTY4MDAwLDcwODIwMDAsNjY1MDAwMCwxMTYwMDAwLDE1NDg0MDAwLDE2ODAwMDAsNTQxNTAwMCwxNTc2NjAwMCwxODkwMDAsMTkwMTgwMDAsMTI3NjYwMDAsNjE3NjAwMCwxMDA4MDAwMCwzNjAxMDAwLDk4MDAwLDE1NDAxMDAwLDE1MDc0MDAwLDEyOTQ1MDAwLDE1OTYxMDAwLDY2MzIwMDAsMTYzMTkwMDAsNzc2NjAwMCwxNDI0MDAwMCwxNjYwMDAwLDMyMzYwMDAsMTQ2NjMwMDAsMjM2MDAwMCwxNTIzMjAwMCwxMTg4MjAwMCwxMDMxMzAwMCwxNzMzNjAwMCwxMTU5MzAwMCw1MDM1MDAwLDY1ODAwMDBdLCJvcHMiOjkwNzM3MzB9LHsibmFtZSI6InNldCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdTZXQuaGFzKCdkb2VzLW5vdC1leGl0JykpIHtcblxufSIsInJ1bnMiOls5MDMyMDAwLDIzODkwMDAsMjYxNTAwMCw1NjYzMDAwLDM5NDQwMDAsMzA1MzAwMCwxMDAwNTAwMCw3NjA0MDAwLDM4NzIwMDAsODA1OTAwMCwxMDM1OTAwMCwxMzI0MzAwMCwxMjUwMDAwMCw1NTc0MDAwLDg4NDEwMDAsNDU0OTAwMCwxMzEyMzAwMCwxNjM5OTAwMCwxMzU3NzAwMCw5ODAwMCwxOTQ5MTAwMCw5MzQ3MDAwLDczMTYwMDAsMTE1MjgwMDAsMzE1MjAwMCw4ODQwMDAwLDEwOTQ5MDAwLDk4MDAwLDI3OTgwMDAsMTQxNDcwMDAsMTIyMDIwMDAsMTQ5NTcwMDAsMTg1MjAwMCw0NDU5MDAwLDEyNzcwMDAwLDExNTQ1MDAwLDIyMTkwMDAsNTI5NDAwMCwyNzEyMDAwLDEzMjg5MDAwLDEyOTAwMCwxMzQwMTAwMCwyNTgwMDAsMTAyNDQwMDAsMTk0NDQwMDAsODQ0NDAwMCw0NzU4MDAwLDE5NDIwMDAsMTY0NjAwMCw4ODk3MDAwLDk1NzgwMDAsMTM1MTQwMDAsMTMwNzAwMCwxMTUzMTAwMCw5NDcxMDAwLDEzMDMzMDAwLDkwMjUwMDAsMTM4OTIwMDAsMzE5MDAwLDIyMjQwMDAsNzkyNDAwMCwxMTAxMjAwMCw3NjgxMDAwLDExNzg1MDAwLDIxNzYwMDAsMTEyMzYwMDAsMzM1MzAwMCw4OTYwMDAsNTczOTAwMCw0NTQxMDAwLDE1MTIzMDAwLDExNDAwMCwxMDc4MDAwMCwxMzk1NTAwMCwxNDM1MTAwMCwyMDY0ODAwMCw5MjIwMDAwLDczNDcwMDAsNjI4NjAwMCwxMzU2MzAwMCwxMTIwNjAwMCwxMzU1OTAwMCw4Mzk0MDAwLDkwMDAsMzUxMzAwMCw1MzI2MDAwLDEwMTIwMDAsNDYzNTAwMCwxNzAxMDAwMCwxMDYyMzAwMCw3MjUwMDAwLDgyMzMwMDAsOTE5MDAwLDEyMjA4MDAwLDk4MDAwLDg5NjEwMDAsMTIyNjQwMDAsOTM2MjAwMCwzMzYzMDAwLDU5MTAwMDBdLCJvcHMiOjc5ODEwNjB9XSwidXBkYXRlZCI6IjIwMjMtMTAtMjdUMTc6MjE6NTguMTI2WiJ9)

builds on top of: https://github.com/glimmerjs/glimmer-vm/pull/1467

[benchmark - Map#has -  key does not exist](https://perf.link/#eyJpZCI6InJ4YmdkMGthcWtmIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHZvaWRNYXAgPSBPYmplY3QuY3JlYXRlKG51bGwpO1xuXG5jb25zdCB2b2lkVGFnTmFtZXMgPSBbJ2FyZWEnLCAnYmFzZScsICdicicsICdjb2wnLCAnY29tbWFuZCcsICdlbWJlZCcsICdocicsICdpbWcnLCAnaW5wdXQnLCAna2V5Z2VuJywgJ2xpbmsnLCAnbWV0YScsICdwYXJhbScsICdzb3VyY2UnLCAndHJhY2snLCAnd2JyJ107XG5cbnZvaWRUYWdOYW1lcy5mb3JFYWNoKCh0YWdOYW1lKSA9PiB7XG4gIHZvaWRNYXBbdGFnTmFtZV0gPSB0cnVlO1xufSk7XG5cbmNvbnN0IHZvaWRUYWdBcnJheSA9IE9iamVjdC5rZXlzKHZvaWRNYXApO1xuY29uc3Qgdm9pZFRhZ1NldCA9IG5ldyBTZXQodm9pZFRhZ0FycmF5KTtcbmNvbnN0IHZvaWRUYWdNYXAgPSBuZXcgTWFwKE9iamVjdC5lbnRyaWVzKHZvaWRNYXApKTsiLCJ0ZXN0cyI6W3sibmFtZSI6Ik1hcCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdNYXAuaGFzKCdkb2VzLW5vdC1leGl0JykpIHtcblxufSIsInJ1bnMiOls2MTkwMDAsNDA3MDAwLDI0NDgwMDAsODc3MzAwMCwyMTQ4OTAwMCw0NTgzMDAwLDQxMTYwMDAsNTY5MTAwMCwxNzc0NjAwMCw3NjcwMDAwLDExNjQ0MDAwLDQyMDkwMDAsMjQ1NjAwMDAsODgxMDAwMCw3Njg0MDAwLDQyNTYwMDAsNjU2MDAwMCw5MDAwLDkwMDAsMTM5NjEwMDAsODA0MzAwMCwxNjQ2MDAwLDU4MDAwMDAsNzU0MTAwMCwxOTAxODAwMCwxMDAwMCwxNTAxMDAwLDk4MDAwLDM3ODEwMDAsMTUyNTAwMCwxMzc2ODAwMCwxOTIwMDAsMTc5MDkwMDAsMzkwNzAwMCwxMTg4NzAwMCw5MDAwLDE5Nzk4MDAwLDg4NDYwMDAsNDYzMjAwMCw4Mzc2MDAwLDk5ODAwMCwzODI1MDAwLDYxOTIwMDAsMTI5MDAwLDYxMTQwMDAsNzk5ODAwMCwxMTQyNjAwMCwxMzA5NzAwMCw2MTc5MDAwLDEwMDQ1MDAwLDEwMDAwLDk1MjQwMDAsMjQwOTYwMDAsMjc3NTAwMCwxMDczODAwMCw0MzYwMDAwLDI5MDQwMDAsMTA3MTYwMDAsNjY1MTAwMCwxMjY0NjAwMCwyNDAxMTAwMCw0MzE1MDAwLDE2ODQwMDAsNzU5NzAwMCwxMzMwMDAwLDUzODcwMDAsODkxMDAwMCw3NDA4MDAwLDQzNTEwMDAsOTgwMDAsMjMyNzAwMCwxMTYxODAwMCwxMDUzMTAwMCwxMzc1NTAwMCwxMTM2ODAwMCwzMzIzMDAwLDE3NzUwMDAsNTc3MjAwMCwxMDY1NDAwMCwxMTc2NjAwMCw3MTU1MDAwLDQzNDEwMDAsNDgxNzAwMCw2NjI2MDAwLDI3MTI1MDAwLDgzMzUwMDAsMTYwNzkwMDAsMTI0NjYwMDAsOTc5ODAwMCwxMjAyNTAwMCw0NjM2MDAwLDg0NTYwMDAsOTk3OTAwMCw1ODQwMDAsMTI5MjYwMDAsNTQ3NDAwMCwxNTU3MDAwLDE2MzAwMCwxMjA5NjAwMCw4OTA5MDAwXSwib3BzIjo3NjU0ODEwfSx7Im5hbWUiOiJvYmplY3QgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkTWFwWydkb2VzLW5vdC1leGl0J10pIHtcblxufSIsInJ1bnMiOls1MDUwMDAsMTQyNTMwMDAsMTM1MjUwMDAsOTM1NDAwMCwyMjg0NDAwMCw0OTc3MDAwLDEwMDAwLDQ0MzAwMCwxMjc4MDAwMCw4OTI1MDAwLDEwNjExMDAwLDI4NjYwMDAsMjMwODcwMDAsNTA5NTAwMCwzNjU3MDAwLDg0OTUwMDAsMjMwNTAwMCwxNDY4NjAwMCw2MDA0MDAwLDY5MjUwMDAsMTc0NDcwMDAsODk4ODAwMCw1MDY5MDAwLDg0NjEwMDAsMTYxNzUwMDAsMTI0OTkwMDAsNjIxNTAwMCwyOTYyMDAwLDk0NjgwMDAsMzQyMzAwMCw1MDQ4MDAwLDc0ODYwMDAsMjA3NTgwMDAsODIxMTAwMCw2OTQ3MDAwLDg5MjEwMDAsNjI0MDAwMCw2MTA1MDAwLDk4ODkwMDAsNjQ3MDAwMCwxOTQyODAwMCw0MDc5MDAwLDkzMDUwMDAsMjU4MTAwMCwyODIyMDAwLDc3ODEwMDAsMTAwMzcwMDAsMTMwMDcwMDAsNTg3OTAwMCwxMDgyMTAwMCwxMTEyMDAwMCw2MTgzMDAwLDIyODQzMDAwLDExMDcxMDAwLDEwNjQ3MDAwLDk3MDgwMDAsMTY0NjAwMCw1MzYzMDAwLDM4NTkwMDAsOTkwMzAwMCwyNDc3NDAwMCwxNTY3MTAwMCwxNDg4MDAwLDkwODEwMDAsMTAwMDAsMTAxMzEwMDAsNTYyNTAwMCw3MTIwMDAwLDM4NTcwMDAsMjExODAwMCw0MzgwMDAsMTAyNTkwMDAsMTEwMzUwMDAsMzI4MzAwMCw3MzEwMDAsNDEzMjAwMCw2ODYzMDAwLDE2NDcwMDAsODU3NzAwMCw2NTY2MDAwLDEyNDc0MDAwLDMyMDEwMDAsNjc3NDAwMCw1MTE4MDAwLDI2MTMwMDAwLDE2NDYwMDAsMjAxNDAwMCwxMjIyOTAwMCw5MDQ5MDAwLDEzMTIwMDAwLDE1MjI2MDAwLDEwNzgzMDAwLDgwNjEwMDAsMzM0MjAwMCwxMjcxMTAwMCwxMzY3NDAwMCwyNjE1MDAwLDg0ODAwMCwyMTU2MDAwLDE0MTA3MDAwXSwib3BzIjo4MzQ4OTYwfSx7Im5hbWUiOiJzZXQgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkVGFnU2V0LmhhcygnZG9lcy1ub3QtZXhpdCcpKSB7XG5cbn0iLCJydW5zIjpbMTYyMjAwMCw5NjcwMDAsNDc5NDAwMCw2NjQ3MDAwLDIxMTE1MDAwLDYzODUwMDAsMTcwNjAwMCw2NTI5MDAwLDExNTAyMDAwLDQ4ODEwMDAsNjYzOTAwMCwyNjE0MDAwLDIwMTE2MDAwLDE1NzAwMDAsNDUwMzAwMCwyNDY4MDAwLDE2NjMwMDAsMTAwMzcwMDAsNTAxNzAwMCw3NzE3MDAwLDEyMzkwMDAwLDI1MTEwMDAsMjI1NzAwMCw0Mjc4MDAwLDE3NTYwMDAwLDkwODAwMDAsNjUwMTAwMCwxMjQwOTAwMCwyOTg1MDAwLDEyNzE1MDAwLDE0NDM5MDAwLDYxNTQwMDAsMTUyMDgwMDAsMzI1OTAwMCw4MjQ2MDAwLDM1NjMwMDAsMjQyNzEwMDAsMzA1OTAwMCwzNjg1MDAwLDg0MzEwMDAsMTY4MTYwMDAsOTM2NjAwMCwyNTYxMDAwLDc4NjUwMDAsMzQyMDAwMCw1ODg2MDAwLDc4NzEwMDAsOTgxMTAwMCw1MzE3MDAwLDcwMjQwMDAsODUxNjAwMCw1Mzk2MDAwLDIxOTEzMDAwLDkyNjgwMDAsODUwNTAwMCwyMzUyMDAwLDEyOTQwMDAsNjI3NDAwMCw0OTI0MDAwLDExNjU4MDAwLDI1MjA2MDAwLDM4MTAwMDAsMTY0NzgwMDAsMzQ4MjAwMCwyNTgwMDAsNzEwNDAwMCw0NjEwMDAwLDc2NDMwMDAsMzE1MjAwMCwzMTY2MDAwLDkwMDAsODMxNjAwMCw5MTEyMDAwLDEyMjE5MDAwLDU4MzMwMDAsNDM2MjAwMCwxNDQyMDAwLDQ4MzAwMCw1MzU2MDAwLDU4OTUwMDAsOTgwMDAsMjg0MjAwMCw1NzE0MDAwLDE2NjYwMDAsMjIzMDYwMDAsMjM3NjAwMCw0NzYwMDAsMTA5OTkwMDAsNDczNjAwMCwxMjIwMjAwMCw3OTcyMDAwLDM2NzQwMDAsNTY0MTAwMCw4MjU5MDAwLDg5OTIwMDAsODA1MDAwMCwxNDY0MTAwMCw5MTMwMDAsMTg2MjAwMCwxMjAwNDAwMF0sIm9wcyI6NzIwODg5MH1dLCJ1cGRhdGVkIjoiMjAyMy0xMC0yN1QxNzoyNjozOS4xOTFaIn0%3D)

[benchmark - Map#has - key exists](https://perf.link/#eyJpZCI6InJ4YmdkMGthcWtmIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHZvaWRNYXAgPSBPYmplY3QuY3JlYXRlKG51bGwpO1xuXG5jb25zdCB2b2lkVGFnTmFtZXMgPSBbJ2FyZWEnLCAnYmFzZScsICdicicsICdjb2wnLCAnY29tbWFuZCcsICdlbWJlZCcsICdocicsICdpbWcnLCAnaW5wdXQnLCAna2V5Z2VuJywgJ2xpbmsnLCAnbWV0YScsICdwYXJhbScsICdzb3VyY2UnLCAndHJhY2snLCAnd2JyJ107XG5cbnZvaWRUYWdOYW1lcy5mb3JFYWNoKCh0YWdOYW1lKSA9PiB7XG4gIHZvaWRNYXBbdGFnTmFtZV0gPSB0cnVlO1xufSk7XG5cbmNvbnN0IHZvaWRUYWdBcnJheSA9IE9iamVjdC5rZXlzKHZvaWRNYXApO1xuY29uc3Qgdm9pZFRhZ1NldCA9IG5ldyBTZXQodm9pZFRhZ0FycmF5KTtcbmNvbnN0IHZvaWRUYWdNYXAgPSBuZXcgTWFwKE9iamVjdC5lbnRyaWVzKHZvaWRNYXApKTsiLCJ0ZXN0cyI6W3sibmFtZSI6Ik1hcCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdNYXAuaGFzKCdsaW5rJykpIHtcblxufSIsInJ1bnMiOlsyMjA0MDAwLDEwNTA5MDAwLDIyMDg5MDAwLDIwODk3MDAwLDE3MTYwMDAsMjEwMDAsMjAxMDEwMDAsMTYyOTAwMCwxNDAzODAwMCw2OTUwMDAsMTA3MzUwMDAsMzYxMjAwMCwxMzE2MjAwMCw3NDMzMDAwLDUyNTUwMDAsMjAwMzkwMDAsNjI1NTAwMCwyMDg0NzAwMCwyNDU1OTAwMCwyNzAzNjAwMCw0MTE3MDAwLDI2NzAwMDAsMTYxMDEwMDAsNjcyMzAwMCw3ODc4MDAwLDQ3NTkwMDAsMTM4MjAwMCwxNzg3NDAwMCwzMDAwMDAwLDEwMDAxMDAwLDIwNzU4MDAwLDU2NTgwMDAsMTcyNTAwMCw2NDk4MDAwLDEwMjUxMDAwLDg5ODEwMDAsOTk0MDAwLDE0MzU5MDAwLDEzODIwMDAsMjA0NzIwMDAsNjA5MDAwLDIyMTUzMDAwLDI5MjIwMDAsMTMwNjYwMDAsNjI1MTAwMCw0NjM5MDAwLDcwOTYwMDAsMjI2MTQwMDAsNTc5NTAwMCwyNzk3MDAwLDE3NjM1MDAwLDIyMTcwMDAsNTIxMjAwMCwxOTExNDAwMCwyMDQzNDAwMCw5NDUyMDAwLDE1NDg1MDAwLDE0NTY0MDAwLDE0OTQ0MDAwLDUzODkwMDAsMTE1MzMwMDAsMjYxMTAwMCw2MTg4MDAwLDE5MDE1MDAwLDg3MTMwMDAsNjM4ODAwMCwzMjA2MDAwLDIyNzI1MDAwLDI3NjMwMDAsMTk4NjUwMDAsOTE0NDAwMCw2NTgwMDAwLDE4ODE2MDAwLDYyNzUwMDAsMjExMjcwMDAsODEyMDAwMCwzODEyMDAwLDE2Mjk4MDAwLDIwMjkwMDAsNTczMTAwMCwzMTMwMDAwLDI0NDIwMDAwLDk2NDAwMCw1MjUzMDAwLDE1NzkxMDAwLDM2NTMwMDAsNTY0OTAwMCw4NzQxMDAwLDEyMTQwMDAsMTI4NjcwMDAsMjI3NjcwMDAsMTgzMzMwMDAsMTMwNTAwMCwxNTU2ODAwMCw3ODIzMDAwLDEzODIwMDAsMzczNDAwMCwxOTA0MDAwMCwzMjMxMDAwLDExMzMwMDAwXSwib3BzIjo5OTc5MzcwfSx7Im5hbWUiOiJvYmplY3QgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkTWFwWydsaW5rJ10pIHtcblxufSIsInJ1bnMiOls1NDc4MDAwLDIyMTg0MDAwLDIyODIwMDAsNTgwMDAwLDE4MzQ1MDAwLDE4NzUyMDAwLDU4MTAwMDAsMTk5MzAwMCwxMjAwNTAwMCwyMTI2MDAwLDEwMzg2MDAwLDIzMzMxMDAwLDQzNzMwMDAsMTcwNDAwMCwzNjk3MDAwLDEwMzEzMDAwLDQ3NjMwMDAsMTI5NjQwMDAsMTA5NTMwMDAsMjAyODQwMDAsMjI4OTAwMCwyMjcwMDAwLDQ5ODAwMDAsMzc4MDAwLDEwMzk5MDAwLDUzMzYwMDAsMTI2MzAwMCwxMDg4OTAwMCwxNTQ0MDAwLDg5OTkwMDAsODE3NzAwMCwzNzUwMDAwLDM4MzIwMDAsMjQ0ODAwMCw3MzM0MDAwLDY1NDQwMDAsMTIwNTgwMDAsMTI1NjcwMDAsMjMwODAwMCwxNTI5ODAwMCw3NjcwMDAsMTA4MjgwMDAsMTcxNTYwMDAsMzc2MjAwMCwyMzI3MDAwLDE1MTk5MDAwLDEzMDgwMDAsMTI4NzMwMDAsMzEzMTAwMCwxOTQwODAwMCw5ODk4MDAwLDM4MDcwMDAsNjQxMzAwMCwzMzU2MDAwLDE2NjM4MDAwLDY4MDAwMDAsODY5NzAwMCw0MDg4MDAwLDEwNTUyMDAwLDE1NjM5MDAwLDEwMjQ5MDAwLDE1NDQwMDAsNDY3NDAwMCwxMzkzMzAwMCwxMDE0NjAwMCw2NDIzMDAwLDE2MjAyMDAwLDE1NDg5MDAwLDIzMDAwMCw4Njc3MDAwLDMyOTkwMDAsMTU0NDAwMCw3NDQ0MDAwLDEzMjIwMDAsNzU0MDAwMCwyNjM3MDAwLDE0Njg0MDAwLDc0NDcwMDAsMTMwMDAwLDUzNTgwMDAsMTY5MTAwMCwxNzE1NDAwMCwzODIyMDAwLDQ5NTQwMDAsOTkzMDAwMCw0OTIwMDAwLDIxMDUwMDAsNzA2MjAwMCwyNDc5NTAwMCwxMjY4NzAwMCw0Mzc1MDAwLDkwNDkwMDAsMTgwMDEwMDAsMTEyMzYwMDAsMzI1NDAwMCwxNTM3NDAwMCwxMTg1MDAwLDExNTQzMDAwLDMyNDAwMCw5NzA5MDAwXSwib3BzIjo3OTk3NzUwfSx7Im5hbWUiOiJzZXQgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkVGFnU2V0LmhhcygnbGluaycpKSB7XG5cbn0iLCJydW5zIjpbNjYyMDAwMCw1NjI4MDAwLDM3NTgwMDAsMTY3MTIwMDAsNDQ2MjAwMCw2NDQ3MDAwLDU4ODQwMDAsMzExNjAwMCwzMTk4MDAwLDE0NDc1MDAwLDQwMTcwMDAsMjEzOTYwMDAsMjIzNzAwMCw2MzAwMDAsMTUxMDEwMDAsNDIyOTAwMCwxODc0OTAwMCwxNzMwOTAwMCwxODc0OTAwMCwyNDMyMDAwLDE2OTk3MDAwLDIwNzIyMDAwLDI5MDAwMDAsMTkwNjYwMDAsMTkwMDAsMzUyMDAwLDk5MjgwMDAsNTQ5NjAwMCwxNDE5OTAwMCwxODAzOTAwMCw2MDQ5MDAwLDEyNjgwMDAsMTM4MjAwMCwyODEwMDAsMTk4NTAwMCwyNTE3MDAwLDEzMjkwMDAsNTI0NTAwMCwxMzMyOTAwMCw2MDcyMDAwLDIwNjUxMDAwLDQwMjQwMDAsMTQ2MjIwMDAsNTI3NDAwMCw2MjcwMDAsNTgzMDAwLDIwNjc1MDAwLDE3NzgwMDAsMTQ0OTQwMDAsNjMwODAwMCwyODEwMDAsMTY4NTEwMDAsNDU2MDAwMCwxMDczMDAwLDY5OTgwMDAsMTU5NDMwMDAsMTE0ODAwMCwyNDQ3MTAwMCwzNzkxMDAwLDE4NTY0MDAwLDM4NzYwMDAsMjI3MDAwLDE3OTUyMDAwLDIwMTg0MDAwLDY0NzAwMCwxMzgyMDAwLDE2NDYwMDAsMjgxOTAwMCwyMDM2NTAwMCwxNzExMDAwLDEzODMwMDAsMTkyODcwMDAsMTcwNDAwMCwxODk1NTAwMCwyNzU3MDAwLDc2MzEwMDAsMTIxNTcwMDAsMjMyOTAwMCwxNDc0MjAwMCw0MTE1MDAwLDE5ODc3MDAwLDgwNjYwMDAsNDYyMDAwLDEwMTcwMDAwLDI3MTUwMDAsMjM3MzAwMCw2MTIwMDAsMzExNjAwMCwxOTM1NzAwMCw0NzMwMDAwLDIxODE3MDAwLDcxNTcwMDAsNjY5MzAwMCwxNDAwMDAsNTQ1MDAwLDE4OTg5MDAwLDExNTUxMDAwLDUwNTgwMDAsMjE0MzUwMDAsMzIwODAwMF0sIm9wcyI6ODI4OTgwMH1dLCJ1cGRhdGVkIjoiMjAyMy0xMC0yN1QxNzozMDowOS4wMjlaIn0%3D)

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/printer.ts
+++ b/packages/@glimmer/syntax/lib/generation/printer.ts
@@ -55,7 +55,7 @@ export interface PrinterOptions {
  *  - Link (component)
  */
 export function isVoidTag(tag: string): boolean {
-  return !!(voidMap.has(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
+  return voidMap.has(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0];
 }
 
 export default class Printer {
--- a/packages/@glimmer/syntax/lib/generation/printer.ts
+++ b/packages/@glimmer/syntax/lib/generation/printer.ts
@@ -4,15 +4,13 @@ import { escapeAttrValue, escapeText, sortByLoc } from './util';
 /**
   * Not for future observers,
   * Object access for small objects is significantly faster than using Set.has
+  * And Map access is even faster than Object.
   */
-export const voidMap: {
-  [tagName: string]: boolean;
-} = Object.create(null);
-
+export const voidMap = new Map<string, boolean>();
 export const voidTagNames = ['area', 'base', 'br', 'col', 'command', 'embed', 'hr', 'img', 'input', 'keygen', 'link', 'meta', 'param', 'source', 'track', 'wbr'];
 
 voidTagNames.forEach((tagName) => {
-  voidMap[tagName] = true;
+  voidMap.set(tagName, true);
 });
 
 const NON_WHITESPACE = /^\S/u;
@@ -46,7 +44,7 @@ export interface PrinterOptions {
  *  - Link (component)
  */
 function isVoidTag(tag: string): boolean {
-  return !!(voidMap[tag.toLowerCase()] && tag[0]?.toLowerCase() === tag[0]);
+  return !!(voidMap.get(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
 }
 
 export default class Printer {
--- a/packages/@glimmer/syntax/lib/generation/printer.ts
+++ b/packages/@glimmer/syntax/lib/generation/printer.ts
@@ -44,7 +44,7 @@ export interface PrinterOptions {
  *  - Link (component)
  */
 function isVoidTag(tag: string): boolean {
-  return !!(voidMap.get(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
+  return !!(voidMap.has(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
 }
 
 export default class Printer {
--- a/packages/@glimmer/syntax/lib/generation/printer.ts
+++ b/packages/@glimmer/syntax/lib/generation/printer.ts
@@ -1,17 +1,28 @@
 import type * as ASTv1 from '../v1/api';
 import { escapeAttrValue, escapeText, sortByLoc } from './util';
 
-/**
-  * Not for future observers,
-  * Object access for small objects is significantly faster than using Set.has
-  * And Map access is even faster than Object.
-  */
-export const voidMap = new Map<string, boolean>();
-export const voidTagNames = ['area', 'base', 'br', 'col', 'command', 'embed', 'hr', 'img', 'input', 'keygen', 'link', 'meta', 'param', 'source', 'track', 'wbr'];
+export const voidMap = new Set([
+  'area',
+  'base',
+  'br',
+  'col',
+  'command',
+  'embed',
+  'hr',
+  'img',
+  'input',
+  'keygen',
+  'link',
+  'meta',
+  'param',
+  'source',
+  'track',
+  'wbr',
+]);
 
-voidTagNames.forEach((tagName) => {
-  voidMap.set(tagName, true);
-});
+export function getVoidTags() {
+  return [...voidMap.keys()];
+}
 
 const NON_WHITESPACE = /^\S/u;
 
@@ -43,7 +54,7 @@ export interface PrinterOptions {
  * Examples when false:
  *  - Link (component)
  */
-function isVoidTag(tag: string): boolean {
+export function isVoidTag(tag: string): boolean {
   return !!(voidMap.has(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
 }
 
--- a/packages/@glimmer/syntax/index.ts
+++ b/packages/@glimmer/syntax/index.ts
@@ -33,7 +33,7 @@ export { default as Walker } from './lib/traversal/walker';
 export * as ASTv1 from './lib/v1/api';
 export { default as builders } from './lib/v1/public-builders';
 export { default as visitorKeys } from './lib/v1/visitor-keys';
-export { voidTagNames } from './lib/generation/printer';
+export { getVoidTags, isVoidTag } from './lib/generation/printer';
 export * as ASTv2 from './lib/v2/api';
 export { normalize } from './lib/v2/normalize';
 export { node } from './lib/v2/objects/node';
--- a/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
+++ b/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
@@ -106,7 +106,7 @@ export class TokenizerEventHandlers extends HandlebarsNodeVisitors {
         );
       }
 
-      if (voidMap.get(tag.name) || tag.selfClosing) {
+      if (voidMap.has(tag.name) || tag.selfClosing) {
         this.finishEndTag(true);
       }
     } else if (tag.type === 'EndTag') {
@@ -263,7 +263,7 @@ export class TokenizerEventHandlers extends HandlebarsNodeVisitors {
     element: ASTv1.ElementNode,
     selfClosing: boolean
   ): void {
-    if (voidMap.get(tag.name) && !selfClosing) {
+    if (voidMap.has(tag.name) && !selfClosing) {
       // EngTag is also called by StartTag for void and self-closing tags (i.e.
       // <input> or <br />, so we need to check for that here. Otherwise, we would
       // throw an error for those cases.
--- a/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
+++ b/packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
@@ -106,7 +106,7 @@ export class TokenizerEventHandlers extends HandlebarsNodeVisitors {
         );
       }
 
-      if (voidMap[tag.name] || tag.selfClosing) {
+      if (voidMap.get(tag.name) || tag.selfClosing) {
         this.finishEndTag(true);
       }
     } else if (tag.type === 'EndTag') {
@@ -263,7 +263,7 @@ export class TokenizerEventHandlers extends HandlebarsNodeVisitors {
     element: ASTv1.ElementNode,
     selfClosing: boolean
   ): void {
-    if (voidMap[tag.name] && !selfClosing) {
+    if (voidMap.get(tag.name) && !selfClosing) {
       // EngTag is also called by StartTag for void and self-closing tags (i.e.
       // <input> or <br />, so we need to check for that here. Otherwise, we would
       // throw an error for those cases.
@@ -299,8 +299,8 @@ export class TokenizerEventHandlers extends HandlebarsNodeVisitors {
         } else {
           throw generateSyntaxError(
             `An unquoted attribute value must be a string or a mustache, ` +
-              `preceded by whitespace or a '=' character, and ` +
-              `followed by whitespace, a '>' character, or '/>'`,
+            `preceded by whitespace or a '=' character, and ` +
+            `followed by whitespace, a '>' character, or '/>'`,
             span
           );
         }
--- a/packages/@glimmer/syntax/test/public-api-test.ts
+++ b/packages/@glimmer/syntax/test/public-api-test.ts
@@ -26,7 +26,8 @@ QUnit.test('exports are not accidentally removed', function(assert) {
   assert.ok(syntax.ASTv1);
   assert.ok(syntax.builders);
   assert.ok(syntax.visitorKeys);
-  assert.ok(syntax.voidTagNames);
+  assert.ok(syntax.getVoidTags);
+  assert.ok(syntax.isVoidTag);
   assert.ok(syntax.ASTv2);
   assert.ok(syntax.normalize);
   assert.ok(syntax.node);


fisker:
It's slower on my machine.
You should also test non-exists tags in the benchmark, I think.

NullVoxPopuli:
good idea.

with [non-existent-keys](https://perf.link/#eyJpZCI6InJ4YmdkMGthcWtmIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHZvaWRNYXAgPSBPYmplY3QuY3JlYXRlKG51bGwpO1xuXG5jb25zdCB2b2lkVGFnTmFtZXMgPSBbJ2FyZWEnLCAnYmFzZScsICdicicsICdjb2wnLCAnY29tbWFuZCcsICdlbWJlZCcsICdocicsICdpbWcnLCAnaW5wdXQnLCAna2V5Z2VuJywgJ2xpbmsnLCAnbWV0YScsICdwYXJhbScsICdzb3VyY2UnLCAndHJhY2snLCAnd2JyJ107XG5cbnZvaWRUYWdOYW1lcy5mb3JFYWNoKCh0YWdOYW1lKSA9PiB7XG4gIHZvaWRNYXBbdGFnTmFtZV0gPSB0cnVlO1xufSk7XG5cbmNvbnN0IHZvaWRUYWdBcnJheSA9IE9iamVjdC5rZXlzKHZvaWRNYXApO1xuY29uc3Qgdm9pZFRhZ1NldCA9IG5ldyBTZXQodm9pZFRhZ0FycmF5KTtcbmNvbnN0IHZvaWRUYWdNYXAgPSBuZXcgTWFwKE9iamVjdC5lbnRyaWVzKHZvaWRNYXApKTsiLCJ0ZXN0cyI6W3sibmFtZSI6Ik1hcCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdNYXAuZ2V0KCdkb2VzLW5vdC1leGl0JykpIHtcblxufSIsInJ1bnMiOls1NjY1MDAwLDE4NjAwMDAsMjEwNTAwMCw0MDMwMDAsNjE4NDAwMCw3MDIwMDAwLDEwMzE4MDAwLDEwOTYxMDAwLDc2MTUwMDAsOTMxNTAwMCwxMTEzNjAwMCw3Mjc2MDAwLDE1NDQ1MDAwLDYzNTQwMDAsMTY3MzIwMDAsNzQ2NzAwMCwxMjkyNDAwMCw0OTk2MDAwLDI0ODEwMDAsMTYwNzAwMCwyMTczMDAwMCwxNDQzNzAwMCwyMTQzMDAwLDE2NzkwMDAwLDI5MjkwMDAsMjg4NDAwMCwxMzg2ODAwMCw5OTU1MDAwLDYwODIwMDAsNjc5MDAwLDI1MjAwMCwxMjAwNDAwMCwzOTAwMDAwLDEwNDExMDAwLDE1MDI1MDAwLDkwMDAsMjM5NzAwMCwxMjA4MDAwMCw0NTAwMDAwLDE2Njk0MDAwLDExNzUwMDAsMTcxNTAwMDAsMTkyNTAwMCwxMjI0MjAwMCw5ODAwMCwxMTgyNDAwMCwyNDExMDAwLDEwMzYzMDAwLDk2NzEwMDAsMTEyNzAwMDAsMTYyODQwMDAsNzM2MDAwLDQ5OTgwMDAsMTYwMDAwLDEyMjA1MDAwLDEwMDAwLDYwNzYwMDAsMTM5OTMwMDAsNzkxNTAwMCw1OTU4MDAwLDc2MDIwMDAsMTI1OTMwMDAsNTQ2MTAwMCwxNDI0MTAwMCw3OTAxMDAwLDE0MTE2MDAwLDEzNzY2MDAwLDg4ODAwMCw0MzU0MDAwLDE3OTEwMDAsMTg1ODcwMDAsODMyMDAwLDE0MjM1MDAwLDE1ODAyMDAwLDExODkxMDAwLDE4MTMwMDAsMTA1MzMwMDAsNjg5MDAwLDEyMDYyMDAwLDYzOTAwMDAsMTI2OTQwMDAsMTMyMzAwMCw3NjQ4MDAwLDE2NDYwMDAsOTQ3OTAwMCwxMzg1MDAwMCwxMzk3MjAwMCw5OTk4MDAwLDExMTcwMDAsMTI0MDAwLDkyNjUwMDAsNTIxMzAwMCw1NDkyMDAwLDEyMDAwLDExNzEzMDAwLDk4MDgwMDAsMTMxMDIwMDAsMTIzNTAwMCw1NjYwMDAwLDEzMDA5MDAwXSwib3BzIjo3ODUwMDQwfSx7Im5hbWUiOiJvYmplY3QgYWNjZXNzIiwiY29kZSI6ImlmICh2b2lkTWFwWydkb2VzLW5vdC1leGl0J10pIHtcblxufSIsInJ1bnMiOlszNzk0MDAwLDg0NDQwMDAsMTIxMjcwMDAsMzY3MDAwMCwxMDY5NzAwMCw1NDM4MDAwLDEyODEzMDAwLDk5ODUwMDAsNjc0OTAwMCw5ODE5MDAwLDE2MzU0MDAwLDUwNjcwMDAsMTMyMjIwMDAsNzc1MTAwMCwxMjAyODAwMCw1OTE0MDAwLDE1MjM1MDAwLDE4NDQzMDAwLDE1Nzc4MDAwLDU0MDAwMCw3MTkzMDAwLDgzNTMwMDAsNzU0MDAwLDEyMDk1MDAwLDk4MDAwLDY1NzAwMDAsOTcyMTAwMCwyNDg5MDAwLDI5ODYwMDAsMTY1NDQwMDAsOTgwMDAsMTU5MjIwMDAsMzcyNTAwMCw2MjI1MDAwLDE1MTgxMDAwLDEzNzI4MDAwLDE5MDMwMDAsNjI3ODAwMCwzNTA1MDAwLDEyNzM0MDAwLDMyMjkwMDAsMTY5NTUwMDAsMTc5MDAwMCwxMDUzOTAwMCwxNTk1NTAwMCw5ODY3MDAwLDEwMjUwMDAwLDQxMTEwMDAsODQzNTAwMCw5NDk0MDAwLDEwMjA3MDAwLDE4MTgwMDAwLDU2MDUwMDAsOTE2MDAwLDkwMTkwMDAsMTQ3MzQwMDAsMTI1NDIwMDAsMTM1NDgwMDAsODAzMDAwMCwzMzQ4MDAwLDE3MDU3MDAwLDEwMjI4MDAwLDczNDYwMDAsMTUyMzYwMDAsMjMzNDAwMCwxMzQ4NzAwMCw1NTY4MDAwLDcwODIwMDAsNjY1MDAwMCwxMTYwMDAwLDE1NDg0MDAwLDE2ODAwMDAsNTQxNTAwMCwxNTc2NjAwMCwxODkwMDAsMTkwMTgwMDAsMTI3NjYwMDAsNjE3NjAwMCwxMDA4MDAwMCwzNjAxMDAwLDk4MDAwLDE1NDAxMDAwLDE1MDc0MDAwLDEyOTQ1MDAwLDE1OTYxMDAwLDY2MzIwMDAsMTYzMTkwMDAsNzc2NjAwMCwxNDI0MDAwMCwxNjYwMDAwLDMyMzYwMDAsMTQ2NjMwMDAsMjM2MDAwMCwxNTIzMjAwMCwxMTg4MjAwMCwxMDMxMzAwMCwxNzMzNjAwMCwxMTU5MzAwMCw1MDM1MDAwLDY1ODAwMDBdLCJvcHMiOjkwNzM3MzB9LHsibmFtZSI6InNldCBhY2Nlc3MiLCJjb2RlIjoiaWYgKHZvaWRUYWdTZXQuaGFzKCdkb2VzLW5vdC1leGl0JykpIHtcblxufSIsInJ1bnMiOls5MDMyMDAwLDIzODkwMDAsMjYxNTAwMCw1NjYzMDAwLDM5NDQwMDAsMzA1MzAwMCwxMDAwNTAwMCw3NjA0MDAwLDM4NzIwMDAsODA1OTAwMCwxMDM1OTAwMCwxMzI0MzAwMCwxMjUwMDAwMCw1NTc0MDAwLDg4NDEwMDAsNDU0OTAwMCwxMzEyMzAwMCwxNjM5OTAwMCwxMzU3NzAwMCw5ODAwMCwxOTQ5MTAwMCw5MzQ3MDAwLDczMTYwMDAsMTE1MjgwMDAsMzE1MjAwMCw4ODQwMDAwLDEwOTQ5MDAwLDk4MDAwLDI3OTgwMDAsMTQxNDcwMDAsMTIyMDIwMDAsMTQ5NTcwMDAsMTg1MjAwMCw0NDU5MDAwLDEyNzcwMDAwLDExNTQ1MDAwLDIyMTkwMDAsNTI5NDAwMCwyNzEyMDAwLDEzMjg5MDAwLDEyOTAwMCwxMzQwMTAwMCwyNTgwMDAsMTAyNDQwMDAsMTk0NDQwMDAsODQ0NDAwMCw0NzU4MDAwLDE5NDIwMDAsMTY0NjAwMCw4ODk3MDAwLDk1NzgwMDAsMTM1MTQwMDAsMTMwNzAwMCwxMTUzMTAwMCw5NDcxMDAwLDEzMDMzMDAwLDkwMjUwMDAsMTM4OTIwMDAsMzE5MDAwLDIyMjQwMDAsNzkyNDAwMCwxMTAxMjAwMCw3NjgxMDAwLDExNzg1MDAwLDIxNzYwMDAsMTEyMzYwMDAsMzM1MzAwMCw4OTYwMDAsNTczOTAwMCw0NTQxMDAwLDE1MTIzMDAwLDExNDAwMCwxMDc4MDAwMCwxMzk1NTAwMCwxNDM1MTAwMCwyMDY0ODAwMCw5MjIwMDAwLDczNDcwMDAsNjI4NjAwMCwxMzU2MzAwMCwxMTIwNjAwMCwxMzU1OTAwMCw4Mzk0MDAwLDkwMDAsMzUxMzAwMCw1MzI2MDAwLDEwMTIwMDAsNDYzNTAwMCwxNzAxMDAwMCwxMDYyMzAwMCw3MjUwMDAwLDgyMzMwMDAsOTE5MDAwLDEyMjA4MDAwLDk4MDAwLDg5NjEwMDAsMTIyNjQwMDAsOTM2MjAwMCwzMzYzMDAwLDU5MTAwMDBdLCJvcHMiOjc5ODEwNjB9XSwidXBkYXRlZCI6IjIwMjMtMTAtMjdUMTc6MjE6NTguMTI2WiJ9)

It looks as if object wins out again.

so I'm going to leave this as an object.

fisker:
`Map#has` seems faster then `Map#get`

NullVoxPopuli:
that's curious. which makes for an interesting situation with this data structure, because using Map#has, makes the Map basically a Set, but Map performs way better than Set.
I wonder if that means browser vendors just haven't optimized Set as extensively yet?

chancancode:
Also if this intended to be public then it’s probably better to export a function to check rather than exposing whatever underlying data structure

NullVoxPopuli:
well, I use `@glimmer/syntax` at runtime for REPL stuff :sweat_smile:

NullVoxPopuli:
> Also if this intended to be public then it’s probably better to export a function to check rather than exposing whatever underlying data structure

@fisker how does prettier need to interact with this data?

fisker:
We bundle them to JSON... 😄

fisker:
And use it in as a Set https://github.com/prettier/prettier/blob/e682577c08e83c0ab272fc73ad1633883ad5d1a5/src/language-handlebars/utils.js#L16

chancancode:
Yeah then I think we just expose `getVoidTags` that returns an array and `isVoidTag` and keep the implementation private

chancancode (approved):
looks good to me! fwiw, I don't know how "public" anything really is, so at some point, probably when we talk about restructuring the packages to simplify things, we probably want to be more rigorous about delineating the public api surface

chancancode:
> well, I use @glimmer/syntax at runtime for REPL stuff 😅

I still doubt this is what is a hot path that makes it slow, but I can believe it if it shows up on benchmarks; anyway now that the implementation is private you are more free to tweak that if you find that necessary; if it ends up being _that_ important, then I suspect the answer is neither `Map` nor `Set`, but perhaps a switch statement in the function where the tags or ordered from the most to least common, and you would probably add fast paths for things like `div`/`p` (the most common non-void tags) that returns false early without having to go through the rest.

chancancode:
But, in general, the performance profiles of these micro things changes over time and we don't have time to keep up with that, so we just end up with a lot of micro optimizations that we don't know if are still necessary or actively causing problems

packages/@glimmer/syntax/lib/generation/printer.ts
```diff
@@ -45,8 +54,8 @@ export interface PrinterOptions {
  * Examples when false:
  *  - Link (component)
  */
-function isVoidTag(tag: string): boolean {
-  return !!(voidMap[tag.toLowerCase()] && tag[0]?.toLowerCase() === tag[0]);
+export function isVoidTag(tag: string): boolean {
+  return !!(voidMap.has(tag.toLowerCase()) && tag[0]?.toLowerCase() === tag[0]);
```
fisker, line 58: `!!` become useless.

NullVoxPopuli: thanks! I wasn't even looking at that -- more ancient code

NullVoxPopuli:
TIL `!!` in a string in bash is replaced with the previously ran command 
image

fisker:
Maybe this is a good chance to mention that several void tags actually missing.

https://github.com/prettier/prettier/pull/14110/files#diff-6faed7f5afc0a02a5e775622809f442bc831a1ac00bbb657cd08014962d5aebeR4

Someone may interest to in investigate.

If they are just missing by mistake, maybe we can switch to use [`html-void-elements`](https://www.npmjs.com/package/) package?

fisker:
And it's funny, three of them removed in HTML spec... https://github.com/wooorm/html-void-elements/pull/7/files

NullVoxPopuli:
oh I like that
image

@chancancode I can do this switch in a separate PR so the PRs stay focused

fisker:
[`html-tags`](https://www.npmjs.com/package/html-tags) is even popular.. But I suggest use `html-void-elements` since the owner is more active in related projects.

Issue #1469, opened 2023-10-27
Meta files should not publish to npm


fisker:
https://unpkg.com/browse/@glimmer/syntax@0.85.3/

I believe `.turbo/` and `rollup.config.mjs` are published unexpected?

NullVoxPopuli:
100% 

also, I really appreciate you opening all these issue! it's a massive help
thank you!!

PR #1470, opened 2023-10-27
Fix the package contents of each package, only publishing dist

NullVoxPopuli:
using `package.json#files`, we have better control over what is published.

Resolves https://github.com/glimmerjs/glimmer-vm/issues/1469

### DIFF
--- a/packages/@glimmer-workspace/eslint-plugin/package.json
+++ b/packages/@glimmer-workspace/eslint-plugin/package.json
@@ -9,8 +9,8 @@
     "eslint-import-resolver-typescript": "^3.5.5",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-n": "^15.7.0",
-    "eslint-plugin-regexp": "^1.15.0",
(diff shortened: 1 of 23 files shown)


packages/@glimmer-workspace/eslint-plugin/package.json
```diff
@@ -9,16 +9,16 @@
     "eslint-import-resolver-typescript": "^3.5.5",
     "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-n": "^15.7.0",
-    "eslint-plugin-regexp": "^1.15.0",
```
NullVoxPopuli, line 12: pnpm likes to alphabetize this stuff, and this change happened automatically

fisker:
Ready to publish a new version?

NullVoxPopuli:
I need to add a couple changes to the rollup build before publishing a new version.

while a bunch of fixes have landed, 0.85.x is still unusable, and I'd like to avoid the churn of slightly-less-broken releases until the build things are addressed.

Coming soon!

PR #1472, opened 2023-10-27
Fix the build. Split dev from prod. In dev, convert import.meta.env.DEV to DEBUG from `@glimmer/env`

NullVoxPopuli:
previously, were kind of compiling away our helpful-information-to-dev-mode users.
but we can't give them import.meta.env either, as our consumers don't have that available.
But they do have `DEBUG` from `@glimmer/env` via babel-plugin-debug-macros, so we'll use that.

I also added conditional exports, because:
- in prod mode (aka, someone using from CDN, or other NODE_ENV=production), we want all debug flags removed for performance
- in dev mode, we want to swap our import.meta.env for DEBUG from @glimmer/env so that our current main consumer (ember.js) can still to build-time swapping of values for _their_ dev vs prod environment.

After this PR, I'll do a patch release, and test setting up `@glimmer/vm` (and friends) in a jsbin, which should hopefully verify that all the prod stuff is configured correctly (this may also be an opportunity to update the js-framework-benchmark to see if anything's different over there

Other changes of note:
- we no longer ship a cjs build, ESM only is what is published
- if consumers need to bundle ESM to something, their existing bundlers can configure that.


package.json
```diff
@@ -20,7 +21,7 @@
     "build:control": "rollup -c rollup.config.mjs",
     "build:flags": "RETAIN_FLAGS=true ember build --env production --suppress-sizes",
     "lint": "npm-run-all lint:*",
-    "lint:files": "dotenv -- turbo test:lint",
+    "lint:files": "dotenv -- turbo lint",
```
NullVoxPopuli, line 24: lint now depends on publint, which also depends on build.
in another PR, I'll be adding an in-CI turbo server to help with build times

NullVoxPopuli:
image

NullVoxPopuli:
It works! https://jsbin.com/yuwofur/edit?html,output

PR #1481, opened 2023-11-01
MakePrimitiveRef generic so that UNDEFINED_REFERENCE can be Reference<undefined> rather than Reference<unknown>

NullVoxPopuli:
ran in to this when trying to update glimmer-vm in ember-source, here: https://github.com/emberjs/ember.js/pull/20561

in particular: 
```
packages/@ember/-internals/glimmer/lib/renderer.ts:396:47 - error TS2345: Argument of type 'Reference<unknown>' is not assignable to parameter of type 'Reference<OutletState | undefined>'.
  Type 'unknown' is not assignable to type 'OutletState | undefined'.

396     let dynamicScope = new DynamicScope(null, UNDEFINED_REFERENCE);
                                                  ~~~~~~~~~~~~~~~~~~~
```

this should be resolved after this PR

### DIFF
--- a/packages/@glimmer/reference/lib/reference.ts
+++ b/packages/@glimmer/reference/lib/reference.ts
@@ -56,8 +56,8 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef(value: unknown): Reference {
-  const ref = new ReferenceImpl(UNBOUND);
+export function createPrimitiveRef<T = unknown>(value: T): Reference<T> {
+  const ref = new ReferenceImpl<T>(UNBOUND);
 
   ref.tag = CONSTANT_TAG;
   ref.lastValue = value;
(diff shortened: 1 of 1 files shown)


packages/@glimmer/reference/lib/reference.ts
```diff
@@ -56,8 +56,8 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef(value: unknown): Reference {
-  const ref = new ReferenceImpl(UNBOUND);
+export function createPrimitiveRef<T = unknown>(value: T): Reference<T> {
```
chancancode, line 59: Don’t think you need `= unknown` here.

I wonder if you will catch any unexpected/unintended usages if you do `T extends string | number | boolean | null | undefined` (we may have a type alias for that already somewhere)

chancancode (approved):

packages/@glimmer/reference/lib/reference.ts
NullVoxPopuli: > Don’t think you need = unknown here.

I will try this out in a followup PR!
Since it's so hard to iterate between this PR and ember, I want try as isolated changes at a time as possible

chancancode: Sure. The `= unknown` is redundant regardless of whether the constraints work out or not. As far as I know it would always have inferred (to unknown in the worst case). For types/interface sometimes you need to say `= unknown` to be able to reference the type more easily, but I don’t think that’s needed in a function parameter position

NullVoxPopuli: > we may have a type alias for that already somewhere

looks like it's just duplicated everywhere -- probably good for better errors, tbh

NullVoxPopuli: https://github.com/glimmerjs/glimmer-vm/pull/1483

PR #1482, opened 2023-11-01
Setup the cache server

NullVoxPopuli:
Because we now build each package 3 times (prod (ESM), dev (ESM), and (CJS)), and because building is required before linting, the `lint` job takes a while. 

By using the cache server, we can hopefully speed up parts of CI.

For example, previously, in: https://github.com/glimmerjs/glimmer-vm/pull/1481 -- https://github.com/glimmerjs/glimmer-vm/actions/runs/6721577834/usage

- Total Runtime is 19m43s
- Human perceived time is 9m 42s

With this PRs changes, we see: https://github.com/glimmerjs/glimmer-vm/actions/runs/6726246999/usage
- Total runtime is about the same
- human percieved time on no-cache hits is slower, but as I was iterating on this, we have the potential to be 2x as fast, I think.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,8 +111,7 @@ jobs:
         with:
           node-version: 20.1.0
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - run: pnpm build
-      - run: pnpm test:types
+      - run: pnpm turbo test:types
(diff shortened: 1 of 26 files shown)


NullVoxPopuli:
it's not yet perfect, but its an improvement.

PR #1483, opened 2023-11-01
Use stricter type for createPrimitiveRef

NullVoxPopuli:
followup to https://github.com/glimmerjs/glimmer-vm/pull/1481/files


packages/@glimmer/reference/lib/reference.ts
```diff
@@ -56,7 +56,7 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef<T = unknown>(value: T): Reference<T> {
+export function createPrimitiveRef<T = string | number | boolean | null | undefined>(value: T): Reference<T> {
```
chancancode, line 59:
```suggestion
export function createPrimitiveRef<T extends string | number | boolean | null | undefined>(value: T): Reference<T> {
```

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -56,7 +56,7 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef<T>(value: T): Reference<T> {
+export function createPrimitiveRef<T extends string | number | boolean | null | undefined>(value: T): Reference<T> {
```
chancancode, line 59:
```suggestion
export function createPrimitiveRef<T extends string| symbol | number | boolean | null | undefined>(value: T): Reference<T> {
```

packages/@glimmer/reference/lib/reference.ts
```diff
@@ -56,7 +56,7 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef<T>(value: T): Reference<T> {
+export function createPrimitiveRef<T extends string| symbol | number | boolean | null | undefined>(value: T): Reference<T> {
```
chancancode, line 59: Technically, maybe `bigint` too, I am not sure we have a use case for it?

PR #1484, opened 2023-11-01
Unify reference factory function signatures

NullVoxPopuli:
Previously done in: https://github.com/glimmerjs/glimmer-vm/pull/1481/
But for createPrimitiveRef, this PR applies the same treatment to the other ref functions as well.

This should fix the last remaining types issue in the ember-source upgrade:
```
packages/@ember/-internals/glimmer/lib/utils/curly-component-state-bucket.ts:38:5 - error TS2322: Type 'Reference<unknown>' is not assignable to type 'Reference<Component<unknown>>'.
  Type 'unknown' is not assignable to type 'Component<unknown>'.

38     this.rootRef = createConstRef(component, 'this');
       ~~~~~~~~~~~~
```

### DIFF
--- a/packages/@glimmer/reference/lib/reference.ts
+++ b/packages/@glimmer/reference/lib/reference.ts
@@ -56,7 +56,7 @@ class ReferenceImpl<T = unknown> implements Reference<T> {
   }
 }
 
-export function createPrimitiveRef<T = unknown>(value: T): Reference<T> {
+export function createPrimitiveRef<T>(value: T): Reference<T> {
   const ref = new ReferenceImpl<T>(UNBOUND);
 
   ref.tag = CONSTANT_TAG;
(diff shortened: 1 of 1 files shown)


packages/@glimmer/reference/lib/reference.ts
chancancode, line 1: can you change the `<T = unknown>` to `<T>` in the same commit to keep it consistent? (it's fine if we never get to tightening the constraints). you probably won't need to wait for a CI run just for that

chancancode (approved):
See comment otherwise lgtm

packages/@glimmer/reference/lib/reference.ts
NullVoxPopuli: can do!

NullVoxPopuli: done

PR #1489, opened 2023-11-03
add tracked utilities for ember-inspector

patricklx:
so we do not need to do this hack:
https://github.com/emberjs/ember-inspector/blob/8c4cd42a50b614981f8b863dfe07e116ced04953/ember_debug/object-inspector.js#L41


patricklx:
@NullVoxPopuli can we have this?

NullVoxPopuli:
Maybe, i'm behind an getting the VM integrated in to ember, so that's blocking me actually reviewing all these prs.

At first glance though, does this need to be in production? Can we instead wrap all the behavior in meta.env?

patricklx:
@chancancode can you have a look at this? At least i would like the object/property info on the tag for inspector

NullVoxPopuli:
image

[artifact-1.pdf](https://github.com/user-attachments/files/15792795/artifact-1.pdf)

NullVoxPopuli:
Is there a way to make it faster?

patricklx:
> Is there a way to make it faster?

Maybe, nodejs creates a new hidden class whenever a new property is added to an object... Lets see if thats the cause

NullVoxPopuli:
image

[artifact-1.pdf](https://github.com/user-attachments/files/15805962/artifact-1.pdf)

packages/@glimmer/validator/test/tracked-utils-test.ts
NullVoxPopuli, line 1: thanks for the tests!! 🎉

packages/@glimmer/validator/index.ts
```diff
@@ -13,8 +13,9 @@ if (globalObj[GLIMMER_VALIDATOR_REGISTRATION] === true) {
 globalObj[GLIMMER_VALIDATOR_REGISTRATION] = true;
 
 export { debug } from './lib/debug';
-export { dirtyTagFor, tagFor, type TagMeta, tagMetaFor } from './lib/meta';
+export { dirtyTagFor, infoForTag, tagFor, type TagMeta, tagMetaFor } from './lib/meta';
 export { trackedData } from './lib/tracked-data';
+export * from './lib/tracked-utils';
```
NullVoxPopuli, line 18: since we export * here, should we do anything in the function implementations themselves to no-op in production?

NullVoxPopuli (approved):
👍 looks good to me.

I don't really know what to do about production debugging, because _it is_ handy to have a good debugging experience in production -- which we already have -- I wonder if that would be more of an opt-out in the future to have all the meta and stuff stripped in prod.

NullVoxPopuli:
I've added 5.11 and 6.0-alpha.1
And did a 6x CPU slowdown to try to account for random machine variance
https://ember-performance-testing-prod.pages.dev/report?benchmarks=%5B%22Render%20complex%20html%20(%40glimmer%2Fcomponent)%22%5D&clear=0&emberVersions=%5B%223.28%22%2C%224.0%22%2C%225.4%22%2C%225.5%22%2C%225.6%22%2C%225.7%22%2C%225.8%22%2C%225.9%22%2C%225.10%22%2C%225.11%22%2C%22ember-canary%22%2C%22ember-canary-custom%22%5D&timePerTest=3000

image

As you can see, there is still some variance, as there isn't really a lot that changed

https://github.com/emberjs/ember.js/compare/3dfb8a4...85a4f298f67ff70395cf7f9103682335162e0606



(but  some logic around EXTEND_PROTOTYPES.Array _did_ change).

3dfb8a4 is the actual v6 alpha.1 sha

NullVoxPopuli (approved):
Approving, as this is mostly moving code out of the inspector

PR #1492, opened 2023-11-12
improve arguments error detection for inspector 

patricklx:


packages/@glimmer/runtime/lib/environment.ts
```diff
@@ -101,9 +102,11 @@ export class EnvironmentImpl implements Environment {
   // Delegate methods and values
   public isInteractive: boolean;
 
+  isArgumentError: typeof isArgumentError;
```
chancancode, line 105: I know I made up the original name, but since we are now stashing it on a more general object, we probably want to name it something slightly more specific and less likely to collide/confuse with other meaning, like `isArgumentCaptureError`, `didErrorDuringCapture` or something similar.

packages/@glimmer/runtime/lib/vm/arguments.ts
```diff
@@ -500,45 +500,54 @@ export function reifyArgs(args: CapturedArguments) {
...
 }
 
+export interface ArgumentError {
+  error: any;
+}
+
+const ARGUMENT_ERROR = Symbol("ARGUMENT_ERROR");
+
+export function isArgumentError(arg: unknown): arg is ArgumentError {
+  return arg !== null && typeof arg === "object" && arg[ARGUMENT_ERROR];
```
chancancode, line 510:
```suggestion
  return arg !== null && typeof arg === "object" && (arg as { [ARGUMENT_ERROR]?: unknown })[ARGUMENT_ERROR]`;
```

...I think

packages/@glimmer/runtime/lib/vm/arguments.ts
```diff
@@ -500,45 +500,54 @@ export function reifyArgs(args: CapturedArguments) {
...
+}
+
+export class ArgumentErrorImpl {
```
chancancode, line 513: Shouldn't need to be exported, yeah?

chancancode (approved):
Other than fixing the lint/type stuff this broadly looks good to me, I don't have a strong opinion on whether we merge it now or after the error branch. I'll leave that up to @NullVoxPopuli

patricklx:
@NullVoxPopuli can we have another look here?

PR #1462, opened 2023-10-24
Error Recovery

wycats:
This pull request implements error recovery in the VM.

More information to come.

PR #1493, opened 2023-11-13
Revert "fix capture render tree fails when errors in args"

NullVoxPopuli:
Reverts glimmerjs/glimmer-vm#1460

This is being reverted to help progress https://github.com/emberjs/ember.js/pull/20561 and address some design stuff around errors in args.

Sorry, @patricklx We'll do a take 3 soon!
(sorry I'm slow at the the ember.js PR! it's reeeeaaaalllly close, I promise!!)

(Also, we probably want to build this work on top of https://github.com/glimmerjs/glimmer-vm/pull/1462 )

### DIFF
--- a/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
+++ b/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
@@ -25,7 +25,7 @@ import {
   JitRenderDelegate,
   RenderTest,
   suite,
-  test, tracked,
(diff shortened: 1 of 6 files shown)


chancancode (approved):
Yep let's work on this in parallel and not block the current merge. I think we can land the other API we discussed without the error-recovery branch, too, and the basic approach would be fine, but yes, I think with Yehuda's branch you would have ways to extract the errors from the new facilities too.

PR #1494, opened 2023-11-14
Compile away local debug for publish

NullVoxPopuli:
Until our primary consumer of the VM natively supports import.meta.env and can incrementally opt in to features via that macro system, we must compile away certain things as they were before in 0.84.x and prior.
Instead of compiling away all DEV to false (which would hide a bunch of info from ember), i made a new flag 

Here is evidence of the local debug stuff being compiled away:
image
(top: compiled, buttom: source)

### DIFF
--- a/packages/@glimmer/local-debug-flags/index.ts
+++ b/packages/@glimmer/local-debug-flags/index.ts
@@ -1,4 +1,8 @@
+
+
 export const LOCAL_DEBUG: true | false =
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
   import.meta.env.VM_LOCAL_DEV &&
   (() => {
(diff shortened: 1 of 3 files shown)


packages/@glimmer/local-debug-flags/index.ts
```diff
@@ -1,5 +1,9 @@
+
+
 export const LOCAL_DEBUG: true | false =
-  import.meta.env.DEV &&
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
NullVoxPopuli, line 5: I didn't know how to best extend the meta.env type here, so I ignored it for now since it's compiled away during publish, and only used in this one file.

chancancode (approved):

PR #1498, opened 2023-11-21
Update guides

NullVoxPopuli:
Extraction from https://github.com/glimmerjs/glimmer-vm/pull/1462


chancancode (changes requested):
These changes wouldn't make sense outside of the PR that introduced the concepts, so I don't think it would make sense to merge it separately –e.g. if we revert we would need to remember to do both. I think we can just leave it with the PR, I am not too worried about 5 extra md files in a clearly separated folder, I was mostly concerned about the formatting changes there

NullVoxPopuli:
makes sense!

PR #1499, opened 2023-11-21
Error Recovery

NullVoxPopuli:
Work from: https://github.com/glimmerjs/glimmer-vm/pull/1462/files

Squashed all 51 commits into one commit and then rebased. 
I was initially rebasing all commits individually, but there were just too many commits.
In any case, CI isn't even green, so... I figured I'd start trying to figure that out.

PR #1500, opened 2023-11-21
Extract formatting / lint changes from the Error Recovery PR, and add prettier to ci / local tooling

NullVoxPopuli:
Instead of: https://github.com/glimmerjs/glimmer-vm/pull/1499

I figured it'd be better to extract the formatting changes from https://github.com/glimmerjs/glimmer-vm/pull/1462/

tl;dr:
- setup/update eslint / prettier on `master`
- squash `feature/handle-error` to single commit without merge
- cherry-pick lint/formatting files to `feature/handle-error`
- format / fix
- `git rebase origin/master`

The way this'll work is we have a one time cost of "run prettier / eslint --fix once" on `master`, and get things in a predictable spot -- then on the error-recovery branch, we'll rebase / squash down to one commit, copy the lint configs from this branch (as the error recovery branch doesn't have passing lints atm, and some stuff is TODO'd / disabled), then we can run lint:fix over there -- then finally rebase on top of `master` -- the goal is to reduce the diff from both branches by getting everything formatting / linting the same.


turbo.json
NullVoxPopuli, line 1: this file was invalid, it's been fixed (had duplicate keys)

package.json
```diff
@@ -21,9 +21,10 @@
     "build:control": "rollup -c rollup.config.mjs",
     "build:flags": "RETAIN_FLAGS=true ember build --env production --suppress-sizes",
     "lint": "npm-run-all lint:*",
+    "lint:format": "prettier -c .",
```
NullVoxPopuli, line 24: formatting wasn't being checked on C.I.... :upside_down_face:

chancancode (approved):
👍 on doing this separately in a single pass, obviously I can’t/didn’t review every file, but as long as this is all automated and without code changes then let’s do it!

wycats:
This is awesome! Thanks so much @NullVoxPopuli.

For future reference, the current formatting rules assume that you're running the equivalent of `eslint --fix` and then `prettier -w` to get the right formatting.

The `eslint` part does a lot of heavy lifting around imports and exports. Some of that heavy lifting is pure formatting (sorting, grouping, etc.) but `eslint` also enforces rules that mandate the use of `import type` when the import itself is unused as a value. This is essential to ensure that the code actually runs as intended at runtime and is also enforced (without fix support) via the [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) tsconfig setting.

In vscode, the expectation is that saving a file will automatically run both eslint fixes and prettier. The settings checked in to `.vscode/settings.json`, assuming the relevant extensions are installed, should make this happen automatically.

PR #1501, opened 2023-11-22
Pre-Error Recovery Diff PR

NullVoxPopuli:
Supersedes: https://github.com/glimmerjs/glimmer-vm/pull/1499

Goal of this PR is to figure out furuther extracting PRs we need before we can have a one-shot at rebasing the Error Recovery PR to reduce the diff.


.eslintrc.cjs
chancancode, line 1: Same question as the other PR: what’s up here?

bin/opcodes.mts
chancancode, line 1: How come this wasn’t caught by the existing eslint rules?

.eslintrc.cjs
NullVoxPopuli: from yehuda:

> TL;DR they're all moved into the plugin

So I can delete this now

bin/opcodes/utils.mts
```diff
@@ -1,7 +1,7 @@
-import { existsSync, readFileSync, writeFileSync } from 'node:fs';
-import { dirname, relative, resolve } from 'node:path';
 import chalk from 'chalk';
 import { spawnSync } from 'node:child_process';
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
+import { dirname, relative, resolve } from 'node:path';
```
chancancode, line 4: @NullVoxPopuli this

bin/update-package-json.mts
```diff
@@ -3,7 +3,7 @@ import { resolve } from 'node:path';
 
 import chalk from 'chalk';
 
-import { type Package, packages } from './packages.mjs';
+import { packages, type Package } from './packages.mjs';
```
chancancode, line 6: @NullVoxPopuli this

NullVoxPopuli: https://github.com/glimmerjs/glimmer-vm/pull/1507

bin/opcodes/utils.mts
NullVoxPopuli: https://github.com/glimmerjs/glimmer-vm/pull/1507

NullVoxPopuli: this is just a mistake, and there is nothing I can do about it other than manual rebasing -- went over the lints with yehuda, and things look correct.

bin/opcodes.mts
NullVoxPopuli: there isn't a lint rule to remove unused imports

chancancode: huh, should we add one/any reason we can’t? doesn’t type script complain about unused stuff in vscode visually (dimming it, at least, but I thought it squiggly it too), I also thought it’s part of all the eslint defaults

NullVoxPopuli:
@chancancode I don't think we should rollback the rename -- there isn't a 1 to 1 mapping between all the concepts.

packages/@glimmer/opcode-compiler/lib/syntax/content.ts
```diff
@@ -0,0 +1,327 @@
...
+
+      () => {
+        op(Op.PushRemoteElement);
+        InvokeStaticBlock(op, block);
+        op(Op.PopRemoteElement);
+      }
+    );
+  });
+
+  statements.add(SexpOpcodes.HandleError, (op, [, handler, block, _inverse]) => {
```
chancancode, line 168: This seems not done

PR #1502, opened 2023-11-22
Lint and config changes extraction part two

NullVoxPopuli:
Followup to https://github.com/glimmerjs/glimmer-vm/pull/1500
Further extraction from https://github.com/glimmerjs/glimmer-vm/pull/1501/
Which supersedes: https://github.com/glimmerjs/glimmer-vm/pull/1499/
Which are both a condensed version of: https://github.com/glimmerjs/glimmer-vm/pull/1462

The goal of #1501 and #1499 are/were to get CI green for the work done in #1462 while reducing the diff.

This is the set of changes on top of #1462 that get the build and lints green


chancancode (approved):
looks good to me! can you squash?

PR #1507, opened 2023-11-22
Adding the prettier organize imports plugin

NullVoxPopuli:
Introduced in: https://github.com/glimmerjs/glimmer-vm/pull/1462

I'm not a fan of this, because there is currently a conflict with eslint/prettier where they can't be ran without creating git diffs.

If this is going to be viable, I need to find the eslint rule that is changing import order.

This was one thing not in the original lint extractions, because the behavior seemed unstable.
- eslint -> caused git diff
- prettier -> caused git diff

(unless both ran together)


NullVoxPopuli:
oh turns out we don't want this, because it was intentionally disabled.

PR #1511, opened 2023-12-05
[perf | trim] Simplify IS_CURRIED usage

lifeart:
It seems we don't really need Set for curried objects


NullVoxPopuli:
why was this closed?

lifeart:
@NullVoxPopuli after double checking, it seems `set` is faster than property check. So, that's why i'm closing it here :)

NullVoxPopuli:
ah ok, cool!

thanks for the context!

PR #1512, opened 2023-12-05
[perf | size] simplify array iterations

lifeart:


packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
```diff
@@ -5,10 +5,10 @@ import { enumerate } from '@glimmer/util';
...
   return createComputeRef(() => {
-    let parts = new Array<string>();
+    const parts = new Array<string>(partsRefs.length).fill('');
 
     for (const [i, ref] of enumerate(partsRefs)) {
-      let value = valueForRef(ref) as Maybe<Dict>;
+      const value = valueForRef(ref) as Maybe<Dict>;
 
       if (value !== null && value !== undefined) {
         parts[i] = castToString(value);
```
lifeart, line 14: using `Array.fill` on line 8 we avoiding holes in `parts` array

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
```diff
@@ -24,7 +24,9 @@ export function createConcatRef(partsRefs: Reference[]) {
 }
 
 function castToString(value: Dict) {
-  if (typeof value.toString !== 'function') {
+  if (typeof value === 'string') {
```
lifeart, line 27: if value is already string, likely we don't need to `toString` it

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
```diff
@@ -89,19 +89,19 @@ APPEND_OPCODES.add(Op.FlushElement, (vm) => {
 });
 
 APPEND_OPCODES.add(Op.CloseElement, (vm) => {
-  let modifiers = vm.elements().closeElement();
+  const modifiers = vm.elements().closeElement();
+  if (modifiers === null) {
```
lifeart, line 93: most elements don't have modifiers, seems `(modifiers === null)` is faster than `(modifiers)`

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: The type said this takes a `Dict`, is the type wrong?

lifeart: it's just ordering, we return most relevant case first

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
lifeart: @chancancode if value will be `null` or `undefined` - hole is created for this index in array

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
chancancode: Is this a general belief/known to be significant, or this spot specifically showed up in macro benchmarks?

packages/@glimmer/runtime/lib/scope.ts
```diff
@@ -43,21 +43,13 @@ export function isScopeReference(s: ScopeSlot): s is Reference {
 
 export class PartialScopeImpl implements PartialScope {
   static root(self: Reference<unknown>, size = 0, owner: Owner): PartialScope {
-    let refs: Reference<unknown>[] = new Array(size + 1);
-
-    for (let i = 0; i <= size; i++) {
-      refs[i] = UNDEFINED_REFERENCE;
-    }
+    let refs: Reference<unknown>[] = new Array(size + 1).fill(UNDEFINED_REFERENCE, 0, size);
```
chancancode, line 46: Same, I am inclined to think this is not a problem unless there are new sources that indicates otherwise

packages/@glimmer/runtime/lib/scope.ts
```diff
@@ -43,21 +43,13 @@ export function isScopeReference(s: ScopeSlot): s is Reference {
...
     return new PartialScopeImpl(refs, owner, null, null, null).init({ self });
   }
 
   static sized(size = 0, owner: Owner): Scope {
-    let refs: Reference<unknown>[] = new Array(size + 1);
-
-    for (let i = 0; i <= size; i++) {
-      refs[i] = UNDEFINED_REFERENCE;
-    }
+    let refs: Reference<unknown>[] = new Array(size + 1).fill(UNDEFINED_REFERENCE, 0, size);
```
chancancode, line 52: Same, I am inclined to think this is not a problem unless there are new sources that indicates otherwise

packages/@glimmer/runtime/lib/scope.ts
lifeart: @chancancode it's 70% faster https://jsben.ch/bDWmT

packages/@glimmer/runtime/lib/scope.ts
lifeart: @chancancode it's 70% faster https://jsben.ch/bDWmT

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: I see, yes that is definitely a problem!

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
lifeart: inside `if` condition if `modifiers is null` it will be casted to bool first, and using explicit `===` with null we avoiding it

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: The problem is in skipping the assignment conditionally though, not necessary with the `[i] = ` style of filling an array

lifeart: @chancancode should it be `array.push` in this case?

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
lifeart: as I see this code used inside `createConcatRef`, and it's  seems we likely have strings here..

packages/@glimmer/runtime/lib/scope.ts
chancancode: Ok sure, yeah it's fine in this case when we are filling it with the same value (though I don't think we should be using that site for benchmarks, and that we don't trust microbenchmarks in general).

What I was trying to say is that you don't have to do this _in general_, when you are going to initialize the array with non-uniform data anyway:

```js
const array = new Array(size + 1);

for (let i = 0; i<size+1; i++) {
  // this should be just fine
  array[i] = Math.random();
}
```

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: The existing for loop should be fine/equivalent `.push` as long as we don't skip it conditionally, I would think

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: Yeah it seems like the type is wrong then, probably should have been `unknown` if we had no idea what's in it

packages/@glimmer/runtime/lib/scope.ts
lifeart: @chancancode agree! but in exact case data will be filled with only one value: `UNDEFINED_REFERENCE`

chancancode: The new code seems buggy though. The previous code was iterating up to `<= size` (as opposed to the normal `< size`) to account for the `+ 1`. So I think you just want to fill the whole array normally as opposed to deliberately under-fill it (which seemed very suspect).

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
chancancode: This is far from the only place we do it though, and it seems like in the past we have reached the conclusion that it doesn't end up mattering in the macro. Anyway, I don't mind the change, but maybe make it `if (modifiers !== null)` so you don't have to generate a diff for the rest of the function?

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: Ah, but yeah, I see, if we switch back to not pre-allocating the length, then we can just `.push` as many times as as needed and then do the `.join`, that seems good to me. That can also get rid of the `enumerate()` thing here I assume

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: Can you change the type too?

lifeart: @chancancode sure, will do it!

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
lifeart: done

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
lifeart: done

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
```diff
@@ -1,17 +1,15 @@
-import type { Dict, Maybe } from '@glimmer/interfaces';
 import type { Reference } from '@glimmer/reference';
 import { createComputeRef, valueForRef } from '@glimmer/reference';
-import { enumerate } from '@glimmer/util';
 
 export function createConcatRef(partsRefs: Reference[]) {
   return createComputeRef(() => {
-    let parts = new Array<string>();
+    const parts = [];
```
chancancode, line 6: The previous annotation was for TypeScript reason (so that it won't let you `push` arbitrary value into it later). I think we still need that right?

lifeart: good catch! Fixed

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: The correct type would be `unknown`

packages/@glimmer/runtime/lib/scope.ts
lifeart: @chancancode agree! Fixed in both cases

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
lifeart: @chancancode in this case we need to add extra check for `object` 
image

packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
lifeart: done

packages/@glimmer/runtime/lib/compiled/expressions/concat.ts
chancancode: ahh, I see, because we already checked for `null` and `undefined` outside the call. yeah, ok in that case `string | object` is probably good enough

chancancode (approved):

PR #1513, opened 2023-12-05
[perf] faster equals

lifeart:


packages/@glimmer/runtime/lib/dom/normalize.ts
```diff
@@ -16,12 +16,12 @@ export function normalizeStringValue(value: unknown): string {
 }
 
 export function normalizeTrustedValue(value: unknown): TrustingInsertion {
-  if (isEmpty(value)) {
-    return '';
-  }
   if (isString(value)) {
```
lifeart, line 19: trusted value in apps likely string and less likely empty

packages/@glimmer/runtime/lib/dom/normalize.ts
```diff
@@ -40,7 +40,7 @@ export function shouldCoerce(
 }
 
 export function isEmpty(value: unknown): boolean {
-  return value === null || value === undefined || typeof (value as Dict).toString !== 'function';
+  return value === undefined || value === null || typeof (value as Dict).toString !== 'function';
```
lifeart, line 43: empty values likely undefined, and less likely null

boris-petrov: Isn't it better to just do `value == null`?

NullVoxPopuli:
how much faster is this?

packages/@glimmer/runtime/lib/dom/normalize.ts
lifeart: I'm not trying to change logic of this function, just ordering code paths to get more "matches" first (in my understanding in any web app `undefined` could be found more often than `null`

lifeart:
@NullVoxPopuli I think we won't get performance benefit for js-framework benchmark, and in theory we could close it.
But in my understanding, if we could do less comparisons and eager return from function, we save computing time.
And according to my exp, working with real life apps, especially without typescript `undefined` could be found more often than `null`.

lifeart:
└────────────────────────────────────────┴──────────────────────────────┘
    ! ALERT     Regression found exceeding the set regression threshold of 25 ms
    Benchmark Results Summary    
duration phase estimated regression +34ms [5ms to 63ms] OR +0.59% [0.09% to 1.07%]
render1000Items1End phase no difference [-7ms to 2ms]
clearItems1End phase no difference [0ms to 0ms]
render1000Items2End phase no difference [-2ms to 4ms]
clearItems2End phase no difference [0ms to 1ms]
render10000Items1End phase no difference [-3ms to 44ms]
clearItems3End phase no difference [-1ms to 2ms]
render1000Items3End phase estimated regression +6ms [1ms to 11ms] OR +1.47% [0.33% to 2.74%]
append1000Items1End phase no difference [-4ms to 3ms]
updateEvery10thItem1End phase no difference [0ms to 1ms]
selectFirstRow1End phase no difference [0ms to 1ms]
selectSecondRow1End phase no difference [-1ms to 0ms]
removeFirstRow1End phase no difference [-1ms to 2ms]
removeSecondRow1End phase no difference [-2ms to 0ms]
swapRows1End phase estimated regression +1ms [0ms to 3ms] OR +2.34% [0.47% to 4.65%]
clearItems4End phase no difference [0ms to 1ms]

PR #1518, opened 2023-12-05
[perf] avoid double memory usage in append opcodes

lifeart:


NullVoxPopuli:
This one def wants to wait on https://github.com/glimmerjs/glimmer-vm/pull/1501

NullVoxPopuli:
@lifeart did you have numbers on this one? I think something this focused is fine to land before error recovery, as the rebase situation wouldn't be too bad over there

lifeart:
@NullVoxPopuli no numbers here, I think we could count it as cleanup from `IE11` support, because `Array.fill` not working in `11` - https://caniuse.com/mdn-javascript_builtins_array_fill
And list of functions is created to simulate this.

Most of usages is static things (initialized once), we unable to catch 'difference'.

Only improvement kinda here
image (looks like it's a mistake in code and `.slice();` is some left copy-pasta from somewhere.

NullVoxPopuli (approved):

PR #1522, opened 2023-12-07
[memory] replace forEach with arrayFrom

lifeart:
Spillover PR from https://github.com/glimmerjs/glimmer-vm/pull/1515#discussion_r1418465619


lifeart:
@NullVoxPopuli I can't say because JS framework benchmark seems support only chrome.

But, using `Array.from` is definitely less hacky comparing to existing  implementation (?micro optimisation?) and should win over years.

I think `forEach` here is used from [IE11](https://caniuse.com/?search=array%20from) support times.
image

NullVoxPopuli:
makes sense to me! thanks!

PR #1523, opened 2023-12-08
Restore a working linking setup

chancancode:
https://github.com/glimmerjs/glimmer-vm/assets/55829/bfed4b98-9c50-427f-963a-6adf1b6d8047

The purpose of this is to restore a working setup where we can link all glimmer-vm to ember.js without publishing.

Previously this was just a script that `yarn link` on both sides.

It seems like we tried to use `yalc` at some point.

Neither of those other solutions fully works with the constraints:

* supports pnpm
* supports pnpm workspace
* supports `publishConfig`
* resolve internal dependencies (like `@glimmer/encoder`) when linked

This script makes it work with all of the above.

Usage:

1. pnpm install, develop, make changes, etc
2. pnpm build
3. pnpm link:all
4. pnpm link:glimmer-vm from ember.js
5. validate, iterate, re-run pnpm build/pnpm link:all as needed
6. pnpm unlink from ember.js
7. pnpm unlink:all here

See also https://github.com/emberjs/ember.js/pull/20590


chancancode:
In hindsight, it may be nice to integrate or do some parts of this as a rollup plugin in the existing build pipeline, so that the `linkable-dist` folder can be kept up-to-date automatically, but that can be future work (it's also a little tricky to get right – right now the script uses `pnpm pack` to get all the right behavior automatically, if we want to do this in rollup, it would have to replicate the `.npmignore`, scripts, etc)

bin/link-all.mts
```diff
@@ -0,0 +1,116 @@
...
+
+    await untar({
+      file: tarball,
+      strip: 1,
+      cwd: pkgDest,
+    });
+
+    await unlink(tarball);
+
+    // https://github.com/pnpm/pnpm/issues/881
```
NullVoxPopuli, line 54: thanks for leaving this link here!

NullVoxPopuli (approved):

PR #1527, opened 2023-12-12
Squashed feature handle error 2

chancancode:
Draft for now, I am not (yet) confident that these are the correct fix


NullVoxPopuli:
`pnpm dedupe` should fix the type error

PR #1531, opened 2023-12-21
[perf]: improve list retaining logic

lifeart:
I found that:

* `shouldRemove` function is not used in list-item opcode
*  it's `reset` function called in only one place and could be replaced with simple assignment

~In addition, it seems we do non-optimal thing here: `updateRef` is called if node is retained or moved, but it not necessary require tag update, adding `valueForRef` here to update tags that actually differs.~ And scope of `updateRef` usage reduced to `updateReferences` opcode function.

This PR a-bit conflicting with https://github.com/glimmerjs/glimmer-vm/pull/1529/files, but could be easily rebased.

### DIFF
--- a/packages/@glimmer/runtime/lib/vm/update.ts
+++ b/packages/@glimmer/runtime/lib/vm/update.ts
@@ -202,14 +202,6 @@ export class ListItemOpcode extends TryOpcode {
     updateRef(this.value, item.value);
     updateRef(this.memo, item.memo);
   }
-
-  shouldRemove(): boolean {
-    return !this.retained;
-  }
-
-  reset() {
-    this.retained = false;
-  }
 }
 
 export class ListBlockOpcode extends BlockOpcode {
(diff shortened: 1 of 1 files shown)


lifeart:
```
    Benchmark Results Summary    

duration phase no difference [-85ms to 63ms]
render1000Items1End phase no difference [-11ms to 3ms]
clearItems1End phase no difference [-1ms to 0ms]
render1000Items2End phase estimated improvement -7ms [-13ms to -1ms] OR -1.61% [-2.99% to -0.31%]
clearItems2End phase no difference [-1ms to 1ms]
render10000Items1End phase no difference [-58ms to 37ms]
clearItems3End phase no difference [-5ms to 5ms]
render1000Items3End phase no difference [-5ms to 12ms]
append1000Items1End phase no difference [-9ms to 5ms]
updateEvery10thItem1End phase no difference [-1ms to 0ms]
selectFirstRow1End phase no difference [0ms to 3ms]
selectSecondRow1End phase no difference [-2ms to 0ms]
removeFirstRow1End phase no difference [-1ms to 4ms]
removeSecondRow1End phase no difference [-2ms to 1ms]
swapRows1End phase no difference [-2ms to 2ms]
clearItems4End phase no difference [-1ms to 2ms]
```
[artifact-1.pdf](https://github.com/glimmerjs/glimmer-vm/files/13759357/artifact-1.pdf)

PR #1532, opened 2023-12-22
Attempt to get perf testing work

lifeart:
image

`pnpm run benchmark:setup`

Comment: this is just setup with build, checkout and auto artifact upload, comment posting (if configured), it not provide actual bench code. We just making sure it's working. Proper performance marks and benchmark logic should be settled independently.

Note, to get GH message working for PR's made from forks, we need to explore this topic: https://github.com/marketplace/actions/add-pr-comment#proxy-for-fork-based-prs


benchmark/benchmarks/krausest/lib/index.js
NullVoxPopuli, line 1: excellent, so in a future PR we could make different render calls or click around and measure the results afterwards.

like, some things we probably want to measure:
- appending 1 row
- appending 1000 rows
- clearing all rows
- switching two rows' positions
- updating data within one row
  - ideally this doesn't cause every row to re-compute, even though I think today it does... :see_no_evil: (depends on where the data is located)
- probably other stuff as well

benchmark/benchmarks/krausest/lib/index.js
```diff
@@ -20,6 +20,31 @@ export default async function render(element, isInteractive) {
...
-
+  enforcePaintEvent();
+  performance.mark('glimmer-render-1000-rows-start');
+  enforcePaintEvent();
   await benchmark.render('Application', args, element, isInteractive);
+  performance.mark('glimmer-render-1000-rows-finished');
+  enforcePaintEvent();
+}
+
+function enforcePaintEvent() {
```
NullVoxPopuli, line 31: do we know why this exists?

benchmark/benchmarks/krausest/vite.config.mts
```diff
@@ -2,15 +2,44 @@ import fs from 'node:fs';
...
+    enforce: 'pre',
     name: '@glimmer/benchmark',
+    resolveId(id) {
+      if (id === '@glimmer/env') {
+        return '\0@glimmer/env';
+      } else if (id === '@glimmer/local-debug-flags') {
+        return '\0@glimmer/local-debug-flags';
+      }
+    },
     load(id) {
```
NullVoxPopuli, line 37: is this needed? the build / package.json#exports should have already handled this :sweat_smile:

NullVoxPopuli: oh I see it was pre-existing from here: https://github.com/glimmerjs/glimmer-vm/pull/1532/files#diff-860eda2adf31f087997e1c92eab487cb90113b3f303653d36adabd42213d5c42L26 

We can probably evaluate removing / cleaning it up in a separate PR then.

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+        (add more cases, and still be able to compare with control)
+      * we should be able to re-run bench in CI from current branch with updated perf source
+
+*/
+
+const experimentBranchName =
+  process.env['EXPERIMENT_BRANCH_NAME'] || (await $`git rev-parse --abbrev-ref HEAD`).stdout.trim();
+const controlBranchName = process.env['CONTROL_BRANCH_NAME'] || 'main';
+const markers =
+  process.env['MARKERS'] ||
```
NullVoxPopuli, line 22: is there any way to track all markers? or do we have to list them out? (I'm a tracerbench newb)

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+const EXPERIMENT_BENCH_DIR = join(EXPERIMENT_DIR, 'benchmark', 'benchmarks', 'krausest');
+
+const pwdRaw = await $`pwd`;
+const pwd = pwdRaw.toString().trim();
+
+// we use benchmark from current commit, very useful if we need to tweak it
+const benchmarkFolder = 'benchmark';
+
+// remove node_modules from benchmark folder, maybe we could figure out better option to distribute bench source
+await $`rm -rf ${join(pwd, benchmarkFolder, 'node_modules')}`;
```
NullVoxPopuli, line 43: non-blocking, I think execa has a `$` now, idk if you need zx

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+await $`rm -rf ${CONTROL_DIR}`;
+await $`rm -rf ${EXPERIMENT_DIR}`;
+await $`mkdir ${CONTROL_DIR}`;
+await $`mkdir ${EXPERIMENT_DIR}`;
+
+const isMacOs = os.platform() === 'darwin';
+
+const BENCHMARK_FOLDER = join(pwd, benchmarkFolder);
+
+const rawUpstreamUrl = await $`git ls-remote --get-url upstream`;
```
NullVoxPopuli, line 55: where is upstream set?

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+
+const CONTROL_PORT = 4020;
+const EXPERIMENT_PORT = 4021;
+const CONTROL_URL = `http://localhost:${CONTROL_PORT}`;
+const EXPERIMENT_URL = `http://localhost:${EXPERIMENT_PORT}`;
+
+// we can't do it in parallel on CI,
+
+// setup experiment
+await within(async () => {
```
NullVoxPopuli, line 78: within is nice!

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+  await $`rm -rf ./benchmark`;
+  await $`cp -r ${BENCHMARK_FOLDER} ./benchmark`;
+
+  console.info('installing experiment source');
+  await $`pnpm install --no-frozen-lockfile`.quiet();
+  console.info('building experiment source, may take a while');
+  await $`pnpm build`.quiet();
+
+  if (isMacOs) {
+    await $`find ./packages -name 'package.json' -exec sed -i '' 's|"main": "index.ts",|"main": "./dist/prod/index.js","module": "./dist/prod/index.js",|g' {} \\;`;
```
NullVoxPopuli, line 91: note for myself and a later PR, I bet we can do this from part of the script that @chancancode made that allows the linking, where a whole new monorepo is created, but just with the published assets and resolved publishConfigs

bin/setup-bench.mjs
```diff
@@ -0,0 +1,155 @@
...
+  await $`git checkout ${experimentBranchName}`;
+  await $`rm -rf ./benchmark`;
+  await $`cp -r ${BENCHMARK_FOLDER} ./benchmark`;
+
+  console.info('installing experiment source');
+  await $`pnpm install --no-frozen-lockfile`.quiet();
+  console.info('building experiment source, may take a while');
+  await $`pnpm build`.quiet();
+
+  if (isMacOs) {
```
NullVoxPopuli, line 90: why is this only needed on MacOS?

NullVoxPopuli (commented):
Just got some questions so I can understand some things.

Thanks for doing this, this is a huge help!!! <3

benchmark/benchmarks/krausest/lib/index.js
lifeart: Yep, all right. We may need to explore/cleanup existing logic (from prev bachn, located https://github.com/glimmerjs/glimmer-vm/tree/main/packages/%40glimmer-workspace/benchmark-env)

bin/setup-bench.mjs
lifeart: @NullVoxPopuli we don't really need all markers (at least, in current step), it's confusing way to measure things, because we must know what we trying to measure.

bin/setup-bench.mjs
lifeart: @NullVoxPopuli I was spend 4 h dealing with `cwd` issues with `zx` (in CI). At the moment I don't have mental health to backport it to execa :) , but not against if smb could unify it.

bin/setup-bench.mjs
lifeart: @NullVoxPopuli it's for developers who work with fork of `glimmerjs/glimmer-vm` (to be able to bench my personal branch locally against `main` in origin)

bin/setup-bench.mjs
lifeart: yep, or we could expose special "build-like-for-bublishing" command (and use it inside folders) to avoid this magic.

bin/setup-bench.mjs
lifeart: In short 
image

bin/setup-bench.mjs
NullVoxPopuli: after I finished reading this file, I'm 100% ok with zx -- the utility functions it provides look good

bin/setup-bench.mjs
NullVoxPopuli: but I mean, how does CI use it? I didn't see a `git remote add upstream ...` anywhere? maybe it's not needed?
I see this if block: https://github.com/glimmerjs/glimmer-vm/pull/1532/files/b23d20697db4bb8d68a69373a2cc41abd7512343#diff-1dbea00b97d62abb4c27f03038299813b6c09433bb293cfe31040c8714cdf4d0R60-R63

but I don't quite understand why you'd check if `upstreamUrl` === `upstream` 
to try myself, I tried on a repo that has no `upstream` remote, and saw this output:
```
❯ git ls-remote --get-url origin
git@github.com:NullVoxPopuli/limber.git

limber-3 on  main [$] 
❯ git ls-remote --get-url upstream
upstream
```

I understand now... but... wtf, git? lol

anywho, thanks! looks good

bin/setup-bench.mjs
NullVoxPopuli: yeah, I think that's what I'll make (already implemented for the linking) and share it with this code

bin/setup-bench.mjs
NullVoxPopuli: curse MacOS and their non gnu shell commands!

NullVoxPopuli (approved):

PR #1533, opened 2023-12-23
Improve benchmark - adding more VM cases

lifeart:
image

image

1. All bench logic is now typescripted
1. Styles added
1. Removed unused code

# Added following cases:

* create 1_000 rows
* create 10_000 rows
* append 1_000 rows
* clear rows
* update every 10 row
* swap rows
* select row
* remove row

[artifact-1.pdf](https://github.com/glimmerjs/glimmer-vm/files/13759060/artifact-1.pdf)


NullVoxPopuli (approved):
Thanks for doing this!

A lot of this code looks veeeery similar to all the other js-framework-benchmark implementations, so that's a win!

PR #1547, opened 2024-01-19
Remove support for Partials

NullVoxPopuli:
wip


packages/@glimmer/runtime/lib/compiled/opcodes/debugger.ts
```diff
@@ -52,16 +52,13 @@ class ScopeInspector {
     let { scope, locals } = this;
```
chancancode, line 52: It seems like we still generally treat the `{{debugger}}` statement as an "runtime eval", which is the same kind of thing as partial.

In order to support it, we have to keep around a lot more strings labels (the names of the variables) than we otherwise need to, just so when we "eval" code (in the case of `{{debugger}}`, calling the `get("...")` function in the console where `"..."` can be any local variables, named arguments, etc.

I think now that "implicit this" and partials are no longer supported, I think we should be able to accomplish the same by doing a bit more analysis at pre-compile time. We should be able to see all the in-scope local variables at that point, and we can internally rewrite:

```hbs
{{#let this.foo as |foo|}}
  {{#let foo.bar as |bar|}} 
    {{!-- magically `get("foo")` and `get("bar")` has to work in the console --}}
    {{debugger}}
  {{/let}}
{{/let}}
```

into 

```hbs
{{#let this.foo as |foo|}}
  {{#let foo.bar as |bar|}} 
    {{debugger foo=foo bar=bar}}
  {{/let}}
{{/let}}
```

Basically we pass every known local variables (and there is no longer such thing as "unknown local variables" anymore) as named arguments, and that should give sufficient information for the `{{debugger}}` implementation to make `get()` work without any pervasive book-keeping everywhere.

Issue #1490, opened 2023-11-09
provide parts nodes for ElementNode and block params


patricklx:
**An ElementNode** can have dots like `<Foo.Bar />`.
But only `Foo` references an upper scope param or variable in JS.
ElementNode should have a `parts` property with an array of nodes for each name part. `Foo` and `Bar` in the example above.

**Regarding block params.**
when block params are available, they are represented as strings only. 
They should also be represented as AST nodes.

this is needed in eslint parser:
https://github.com/ember-cli/eslint-plugin-ember/blob/3ec4b269f8b8cc47fe8fca2110f4d671a80d0fab/lib/parsers/gjs-gts-parser.js#L264
and
https://github.com/ember-cli/eslint-plugin-ember/blob/3ec4b269f8b8cc47fe8fca2110f4d671a80d0fab/lib/parsers/gjs-gts-parser.js#L290

- [x] #1552 
- [x] #1553

PR #1552, opened 2024-01-25
add block param nodes

patricklx:
part of #1490 

made it backwards compatible

### DIFF
--- a/packages/@glimmer/syntax/lib/v1/nodes-v1.ts
+++ b/packages/@glimmer/syntax/lib/v1/nodes-v1.ts
@@ -48,6 +48,7 @@ export interface Program extends CommonProgram {
 
 export interface Block extends CommonProgram {
   type: 'Block';
+  blockParamNodes: BlockParam[];
 }
 
 export type EntityEncodingState = 'transformed' | 'raw';
@@ -180,6 +181,7 @@ export interface ElementNode extends BaseNode {
   selfClosing: boolean;
   attributes: AttrNode[];
   blockParams: string[];
+  blockParamNodes: BlockParam[];
   modifiers: ElementModifierStatement[];
   comments: MustacheCommentStatement[];
   children: Statement[];
@@ -326,6 +328,16 @@ export interface HashPair extends BaseNode {
   value: Expression;
 }
 
+/**
+ * a param inside the pipes of elements or mustache blocks,
+ * <Foo as |bar|>... bar is a BlockParam.
+ * {{#Foo as |bar|}}... bar is a BlockParam.
+ */
+export interface BlockParam extends BaseNode {
+  type: 'BlockParam';
+  value: string;
+}
+
 export interface StripFlags {
   open: boolean;
   close: boolean;
@@ -361,6 +373,7 @@ export type Nodes = SharedNodes & {
   PathExpression: PathExpression;
   Hash: Hash;
   HashPair: HashPair;
+  BlockParam: BlockParam;
 };
 
 export type NodeType = keyof Nodes;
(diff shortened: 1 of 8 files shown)


bin/run-tests.mjs
NullVoxPopuli, line 1: why did this file change?

NullVoxPopuli (approved):

bin/run-tests.mjs
patricklx: Oh , this was to run them on windows. But isn't really necessary. Can just run vite

packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -314,6 +314,30 @@ data-barf="herpy"
...
+  }
+});
+
+test('mustache block params', () => {
+  let ast = parse(`{{#Foo as |ab cd efg|}}{{/Foo}}`);
+
+  let [Foo] = ast.body;
+  if (assertNodeType(Foo, 'BlockStatement')) {
+    let [ab, cd, efg] = guardArray({ blockParamNodes: Foo.program.blockParamNodes }, { min: 3 });
+    locEqual(ab, 1, 11, 1, 13);
```
NullVoxPopuli, line 335: can the value / name be asserted as well?

bin/run-tests.mjs
NullVoxPopuli: can you back out these changes and PR separately if they're needed? <3

NullVoxPopuli (changes requested):
revert bin/run-tests.mjs <3

packages/@glimmer/syntax/test/loc-node-test.ts
patricklx: i think this should be done separately. i will add it  in parser-node-test

packages/@glimmer/syntax/lib/v1/nodes-v1.ts
```diff
@@ -293,6 +295,16 @@ export interface HashPair extends BaseNode {
   value: Expression;
 }
 
+/**
+ * a param inside the pipes of elements or mustache blocks,
```
NullVoxPopuli, line 299: :raised_hands: perfect! thank you!

packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -314,6 +314,30 @@ data-barf="herpy"
   }
 });
 
+test('element block params', () => {
+  let ast = parse(`<Foo as |ab cd efg|></Foo>`);
+
+  let [Foo] = ast.body;
+  if (assertNodeType(Foo, 'ElementNode')) {
+    let [ab, cd, efg] = guardArray({ blockParamNodes: Foo.blockParamNodes }, { min: 3 });
+    locEqual(ab, 1, 8, 1, 12);
```
NullVoxPopuli, line 323: since the signature for `locEqual` is
```js
function locEqual(
  node: AST.Node | null | undefined,
  startLine: number,
  startColumn: number,
  endLine: number,
  endColumn: number,
  message = JSON.stringify(node)
) {
```
should `endColumn` be 2 characters more than `startColumn`? right now it sees like.. 4?

patricklx: oh! thanks for noticing. since those are initially parsed as attributes by hbs parser the location contains the pipe as well... need to fix it. not sure why it also contains the whitespace, might be a bug in hbs parser
```js
't=<Foo as |ab cd efg|></Foo>'
t.slice(8,12)
'|ab '
t.slice(12,15)
'cd '
t.slice(15,19)
'efg|'
```

NullVoxPopuli (approved):
ranges look good, thank you!

PR #1556, opened 2024-02-01
Restore test harness on CI

chancancode:
The test harness haven't been accurately reporting failures, this restores it.


packages/@glimmer/validator/lib/validators.ts
NullVoxPopuli, line 1: 🎉

PR #1560, opened 2024-02-05
add in element to debug render tree

patricklx:
My learnings from hacking on this in inspector
https://github.com/emberjs/ember-inspector/pull/2549

Adding modifiers is also easy, will add it in another pr

### DIFF
--- a/packages/@glimmer/interfaces/lib/runtime/debug-render-tree.d.ts
+++ b/packages/@glimmer/interfaces/lib/runtime/debug-render-tree.d.ts
@@ -3,7 +3,7 @@ import type { SimpleElement, SimpleNode } from '@simple-dom/interface';
 import type { Bounds } from '../dom/bounds.js';
 import type { Arguments, CapturedArguments } from './arguments.js';
 
-export type RenderNodeType = 'outlet' | 'engine' | 'route-template' | 'component';
+export type RenderNodeType = 'outlet' | 'engine' | 'route-template' | 'component' | 'keyword';
 
 export interface RenderNode {
   type: RenderNodeType;
--- a/packages/@glimmer/runtime/lib/vm/element-builder.ts
+++ b/packages/@glimmer/runtime/lib/vm/element-builder.ts
@@ -1,6 +1,7 @@
 import type {
   AttrNamespace,
   Bounds,
+  CapturedArguments,
   Cursor,
   CursorStackSymbol,
   ElementBuilder,
@@ -12,6 +13,7 @@ import type {
   Maybe,
   ModifierInstance,
   Nullable,
+  Reference,
   SimpleComment,
   SimpleDocumentFragment,
   SimpleElement,
@@ -20,6 +22,7 @@ import type {
   UpdatableBlock,
 } from '@glimmer/interfaces';
 import { destroy, registerDestructor } from '@glimmer/destroyable';
+import { createConstRef } from '@glimmer/reference';
 import { assert, expect, Stack } from '@glimmer/util';
 
 import type { DynamicAttribute } from './attributes/dynamic';
@@ -100,7 +103,6 @@ export class NewElementBuilder implements ElementBuilder {
 
   constructor(env: Environment, parentNode: SimpleElement, nextSibling: Nullable<SimpleNode>) {
     this.pushElement(parentNode, nextSibling);
-
     this.env = env;
     this.dom = env.getAppendOperations();
     this.updateOperations = env.getDOM();
@@ -214,15 +216,31 @@ export class NewElementBuilder implements ElementBuilder {
     element: SimpleElement,
     guid: string,
     insertBefore: Maybe<SimpleNode>
-  ): Nullable<RemoteLiveBlock> {
-    return this.__pushRemoteElement(element, guid, insertBefore);
+  ): RemoteLiveBlock {
+    const block = this.__pushRemoteElement(element, guid, insertBefore);
+    if (this.env.debugRenderTree) {
+      const namedArgs: Record<string, Reference> = {};
+      if (insertBefore !== undefined) {
+        namedArgs['insertBefore'] = createConstRef(insertBefore, false);
+      }
+      this.env.debugRenderTree.create(block, {
+        type: 'keyword',
+        name: 'in-element',
+        args: {
+          named: namedArgs,
+          positional: [createConstRef(element, false)],
+        } as CapturedArguments,
+        instance: null,
+      });
+    }
+    return block;
   }
 
   __pushRemoteElement(
     element: SimpleElement,
     _guid: string,
     insertBefore: Maybe<SimpleNode>
-  ): Nullable<RemoteLiveBlock> {
+  ): RemoteLiveBlock {
     this.pushElement(element, insertBefore);
 
     if (insertBefore === undefined) {
@@ -236,12 +254,20 @@ export class NewElementBuilder implements ElementBuilder {
     return this.pushLiveBlock(block, true);
   }
 
-  popRemoteElement() {
-    this.popBlock();
+  popRemoteElement(): void {
+    const block = this.popBlock();
     this.popElement();
+    const parentElement = this.element;
+    if (this.env.debugRenderTree) {
+      this.env.debugRenderTree?.didRender(block, {
+        parentElement: () => parentElement,
+        firstNode: () => block.firstNode(),
+        lastNode: () => block.lastNode(),
+      });
+    }
   }
 
-  protected pushElement(element: SimpleElement, nextSibling: Maybe<SimpleNode> = null) {
+  protected pushElement(element: SimpleElement, nextSibling: Maybe<SimpleNode> = null): void {
     this[CURSOR_STACK].push(new CursorImpl(element, nextSibling));
   }
 
@@ -268,7 +294,7 @@ export class NewElementBuilder implements ElementBuilder {
     return element;
   }
 
-  willCloseElement() {
+  willCloseElement(): void {
     this.block().closeElement();
   }
 
--- a/packages/@glimmer/runtime/lib/vm/element-builder.ts
+++ b/packages/@glimmer/runtime/lib/vm/element-builder.ts
@@ -232,6 +232,9 @@ export class NewElementBuilder implements ElementBuilder {
         } as CapturedArguments,
         instance: null,
       });
+      registerDestructor(block, () => {
+        this.env.debugRenderTree?.willDestroy(block);
+      });
     }
     return block;
   }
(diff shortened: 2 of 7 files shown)


NullVoxPopuli (approved):
Looks good to me. 

Let's give @chancancode or @wycats a chance to review tho (couple days?)

packages/@glimmer/interfaces/lib/runtime/debug-render-tree.d.ts
```diff
@@ -3,7 +3,12 @@ import type { SimpleElement, SimpleNode } from '@simple-dom/interface';
 import type { Bounds } from '../dom/bounds.js';
 import type { Arguments, CapturedArguments } from './arguments.js';
 
-export type RenderNodeType = 'outlet' | 'engine' | 'route-template' | 'component';
+export type RenderNodeType =
+  | 'outlet'
+  | 'engine'
+  | 'route-template'
+  | 'component'
+  | 'remote-element';
```
chancancode, line 11: I think it would make more sense to add a more generic type here, either `keyword` or `block` (as in `{{#this-is-a-block}}`). This would, in theory, allow it to be used in the future for other things like `{{#if}}`, `{{#each}}`, etc. 

Not saying we should add those now – with the current architecture of the debug render tree it's not obvious that we can afford to do it, both in terms of the added (unstrippable) code size and runtime cost. I am ok with added these things that are high value (improving the UX in the inspector that is otherwise very annoying to infer from within the inspector code) and happens infrequently, like `{{#in-element}}`, but, I would otherwise be quite uneasy about doing it for other keywords (and probably helpers, which presumably happens a lot more than modifiers).

We inherited the current UX and architecture of that tab in the inspector from wayyyyy back. DebugRenderTree was a massive improvement compared to the code that existed before in terms of rationalizing the data structure and processing, but IMO the architecture itself (where the code lives, how to make it more opt-in, the amount of work it takes after each re-render, and the serializing/sending/deserializing of the tree) is itself overdue for a ground-up rethink, especially in light of Embroider/code-stripping. So we are a bit bottlenecked in terms of what kinds of improvements we can be making that are still viable within the limits of the current architecture.

That being said, I think there is no reason to limit ourselves here and lock ourselves in here. Hence the suggestion. I am imagining that when the inspector code sees the `keyword`/`block` type node, it can just always render something like `{{#${node.name} ${node.args.positional} ${node.args.named}}}`. If we pull on this thread more we'll eventually have to figure out what to do with block params (if we can even do anything about it), but we can worry about that another time.

packages/@glimmer/runtime/lib/vm/element-builder.ts
```diff
@@ -83,6 +84,7 @@ export class NewElementBuilder implements ElementBuilder {
   [CURSOR_STACK] = new Stack<Cursor>();
   private modifierStack = new Stack<Nullable<ModifierInstance[]>>();
   private blockStack = new Stack<LiveBlock>();
+  private remoteStateStack: Object[];
```
chancancode, line 87: Shouldn't be necessary (and even if it were, `Object` would definitely be the wrong type)

packages/@glimmer/runtime/lib/vm/element-builder.ts
```diff
@@ -215,6 +217,29 @@ export class NewElementBuilder implements ElementBuilder {
...
+        } as any,
+        instance: {
+          args: {
+            destination: ref,
+          },
+          constructor: {
+            name: 'InElement',
+          },
+        },
+      });
```
chancancode, line 241:
```suggestion
      this.env.debugRenderTree.create(state, {
        type: 'keyword',
        name: 'in-element',
        args: createCapturedArgs(
          // not sure why the order is backwards, but it's named then positional
          { insertBefore: createConstRef(insertBefore, false) },
          [createConstRef(element, false)],
        ),
        instance: null,
      });
```

packages/@glimmer/runtime/lib/vm/element-builder.ts
```diff
@@ -237,6 +262,15 @@ export class NewElementBuilder implements ElementBuilder {
...
   popRemoteElement() {
+    if (this.env.debugRenderTree) {
+      const state = this.remoteStateStack.pop();
+      const element = this.element;
+      this.env.debugRenderTree?.didRender(state!, {
+        parentElement: () => (element as any).parentElement,
+        firstNode: () => element,
+        lastNode: () => element,
+      });
+    }
```
chancancode, line 273: `this.popBlock()` returns the bounds so you would just call it first, store it in a local variable, and then pass that to `didRender`, rather than having to make your own

chancancode: As for "state", it just have to be a stable object that uniquely identify this node, and so it actually works to just use the "block" object – the one you get back from calling `pushLiveBlock` and `popBlock` (regarding my comment above, `Block implements Bounds`), which would be the same object, because that is already stored in a stack, which is why I said it would be unnecessary for you to maintain yet another parallel stack just to track the same.

However, it looks like something is currently wrong with the types so you would have to fix it before you can do that.

First of all, you should have access to the block returned by `pushLiveBlock` (so you can pass it to `debugRenderTree.create` from within `pushRemoteElement`). However, that is happening from with `__pushRemoteElement`, which currently has a return type of `Nullable<RemoteLiveBlock>`. I think that is wrong – specifically, the `Nullable<...>` is not accurate, as far as I can tell, there is no code path that would return `null`, and if there is, it would be a bug to unconditionally call `popBlock` here in `popRemoteElement`.

So you'll have to fix that return type, and then do the same in the rehydration builder, then, you should have everything you need to implement the feature without having to add your own parallel stack.

chancancode: And – while you are in this file, we really shouldn't be inferring the return types of these methods in this file (ie. a bunch of methods does not have `: ...` here). It seems like most of them are just `: void` and we should have typed that explicitly.

Ideally we should fix that in a separate PR, either just in this file, or across the whole codebase and enforce it with a lint, that would be great. It will probably give @NullVoxPopuli some rebase nightmares though, so you may want to coordinate with them on Discord about how to approach it; I think it is ultimately a rather mechanical fix that should be possible to do, but @NullVoxPopuli can figure that out with you.

chancancode (changes requested):
see all of the above

packages/@glimmer/runtime/lib/vm/element-builder.ts
chancancode: *see the other comment regarding `state`

packages/@glimmer/interfaces/lib/runtime/debug-render-tree.d.ts
patricklx: used keyword

packages/@glimmer/runtime/lib/vm/element-builder.ts
patricklx: block seems to have the bounds set to the children of the remote element. so I reverted to pass custom bounds

chancancode: That is the correct usage of bounds though. It is what each node has rendered. A component’s bounds is its content. And you need that if you want to draw the bounding box if the inspector

packages/@glimmer/runtime/lib/vm/element-builder.ts
chancancode: Hm, let's see how we can resolve that, what was the cycle?

packages/@glimmer/runtime/lib/vm/element-builder.ts
chancancode: Still relevant (maybe re-read the comment about using the bounds object to uniquely identify the node?)

packages/@glimmer/runtime/lib/vm/element-builder.ts
patricklx: I'm not entirely sure

I think
* arguments
* stack
* low-level
* opcodes
* element-builder

But i also saw this:

* arguments
* compiled/opcodes/-debug-strip
* ../vm/arguments

patricklx: The error happens in glimmer/runtime/lib/compiled/opcodes/vm.ts:38:1
APPEND_OPCODES is not defined

packages/@glimmer/runtime/lib/vm/element-builder.ts
patricklx: removed

packages/@glimmer/runtime/lib/vm/element-builder.ts
```diff
@@ -214,15 +216,31 @@ export class NewElementBuilder implements ElementBuilder {
     element: SimpleElement,
     guid: string,
     insertBefore: Maybe<SimpleNode>
-  ): Nullable<RemoteLiveBlock> {
-    return this.__pushRemoteElement(element, guid, insertBefore);
+  ): RemoteLiveBlock {
+    const block = this.__pushRemoteElement(element, guid, insertBefore);
+    if (this.env.debugRenderTree) {
+      const namedArgs: Record<string, Reference> = {};
+      if (insertBefore) {
```
chancancode, line 223: I'll see if I can figure out how we can use the proper utilities so we can get rid of the type cast below; but by the way, we shouldn't guard it like this – it maybe fine to omit `undefined` (or we can just let inspector make the choice), but for `in-element`, `null` and `undefined are specifically different – `undefined` clears the element and `null` inserts it at the end of the children (ie append), so we would want to preserve that

packages/@glimmer/runtime/lib/vm/element-builder.ts
patricklx: True, it will not work with insertAfter, otherwise this would work. 
Would it be okay to still have custom bounds where parentElement is the parent of the in-element? I would like to have that link for the inspector.
 first&last would use the block bounds.
From the code in glimmer i think parent element is checked to verify if a node was removed or moved to another block.
In the inspector it was only used to check if a node is still in the dom.  It uses node.isConnected now.

chancancode: @patricklx I don't really understand this, but no, something sounds off. The bounds represents a range in the DOM, denoted by `(element, firstChild, lastChild)` where first/last child belongs to the element. You should expect to be able to find a sub-slice in `element.childNodes` that matches `(firstChild, lastChild)`. The inspector should be using this to highlight (bounding rect overlay). So something seems fishy about what you wanted, can you link to the code?

patricklx: yes, first and last child is the same .
And inspector only uses those:
https://github.com/emberjs/ember-inspector/blob/56575c08bcd3228d830ace596ebb778af2dd4824/ember_debug/libs/render-tree.js#L326

For another use case i would like to have parentElement. Or another way to get it.
Basically with modifiers i am putting them under an html element in inspector. Also all components that are under that element are also put under it. But i cannot know if the in-element was also under that element because it's disconnected

NullVoxPopuli (approved):

PR #1566, opened 2024-02-21
Allow to render TrustedHTML objects similar to SafeString

crypto:


NullVoxPopuli:
Is this no longer worth pursuing?

NullVoxPopuli:
Ah i see, reopened here: https://github.com/glimmerjs/glimmer-vm/pull/1567

PR #1567, opened 2024-02-21
Allow to render TrustedHTML objects similar to SafeString

crypto:
# Summary
Currently [TrustedHTML](https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML) objects converted to strings during rendering before hitting DOM API that means all HTML insertions handled by default policy. That prevents as from migrating to Trusted Types. This change allows to render TrustedHTML objects similar to SafeString.

# Testing done

```
ok 2382 [integration] jit :: TrustedHTML > renders TrustedHTML similar to SafeString
ok 2383 [integration] jit :: TrustedHTML > renders TrustedHTML in attribute context as string
1..2686
# pass 2685
# skip 1
# todo 0
# fail 0
```

### DIFF
--- a/packages/@glimmer-workspace/integration-tests/test/trusted-html-test.ts
+++ b/packages/@glimmer-workspace/integration-tests/test/trusted-html-test.ts
@@ -1,0 +1,49 @@
+import type { TrustedTypePolicy, TrustedTypesWindow } from 'trusted-types/lib';
+
+import { jitSuite, RenderTest, test } from '..';
+
+let policy: TrustedTypePolicy | undefined;
+if (typeof window !== 'undefined') {
+  let trustedTypes = (window as unknown as TrustedTypesWindow).trustedTypes;
+  if (trustedTypes?.createPolicy) {
+    policy = trustedTypes.createPolicy('test', {
+      createHTML: (s: string) => s,
+      createScript: (s: string) => s,
+      createScriptURL: (s: string) => s,
+    });
+  }
+}
+
+export class TrustedHTMLTests extends RenderTest {
+  static suiteName = 'TrustedHTML';
+
+  @test
+  'renders TrustedHTML similar to SafeString'() {
+    if (!policy) return;
+
+    let html = '<b>test\'"&quot;</b>';
+    this.registerHelper('trustedHTML', () => {
+      return policy?.createHTML(html);
+    });
+
+    this.render('<div>{{trustedHTML}}</div>');
+    this.assertHTML('<div><b>test\'""</b></div');
+    this.assertStableRerender();
+  }
+
+  @test
+  'renders TrustedHTML in attribute context as string'() {
+    if (!policy) return;
+
+    let html = '<b>test\'"&quot;</b>';
+    this.registerHelper('trustedHTML', () => {
+      return policy?.createHTML(html);
+    });
+
+    this.render('<a title="{{trustedHTML}}">{{trustedHTML}}</a>');
+    this.assertHTML('<a title="<b>test\'&quot;&amp;quot;</b>"><b>test\'""</b></a>');
+    this.assertStableRerender();
+  }
+}
+
+jitSuite(TrustedHTMLTests);
--- a/packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
+++ b/packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
@@ -74,6 +74,11 @@ export function StdAppend(
         op(Op.AppendSafeHTML);
       });
 
+      when(ContentType.TrustedHTML, () => {
+        op(Op.AssertSame);
+        op(Op.AppendHTML);
+      });
+
       when(ContentType.Fragment, () => {
         op(Op.AssertSame);
         op(Op.AppendDocumentFragment);
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/content.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/content.ts
@@ -11,7 +11,7 @@ import { isObject } from '@glimmer/util';
 import { ContentType, CurriedType, Op } from '@glimmer/vm';
 
 import { isCurriedType } from '../../curried-value';
-import { isEmpty, isFragment, isNode, isSafeString, shouldCoerce } from '../../dom/normalize';
+import { isEmpty, isFragment, isNode, isSafeString, isTrustedHTML, shouldCoerce } from '../../dom/normalize';
 import { APPEND_OPCODES } from '../../opcodes';
 import DynamicTextContent from '../../vm/content/text';
 import { CheckReference } from './-debug-strip';
@@ -32,6 +32,8 @@ function toContentType(value: unknown) {
     return ContentType.Helper;
   } else if (isSafeString(value)) {
     return ContentType.SafeString;
+  } else if (isTrustedHTML(value)) {
+    return ContentType.TrustedHTML;
   } else if (isFragment(value)) {
     return ContentType.Fragment;
   } else if (isNode(value)) {
@@ -87,7 +89,7 @@ APPEND_OPCODES.add(Op.AppendHTML, (vm) => {
   let reference = check(vm.stack.pop(), CheckReference);
 
   let rawValue = valueForRef(reference);
-  let value = isEmpty(rawValue) ? '' : String(rawValue);
+  let value = isEmpty(rawValue) ? '' : isTrustedHTML(rawValue) ? rawValue as string : String(rawValue);
 
   vm.elements().appendDynamicHTML(value);
 });
--- a/packages/@glimmer/runtime/lib/dom/normalize.ts
+++ b/packages/@glimmer/runtime/lib/dom/normalize.ts
@@ -1,4 +1,5 @@
 import type { Dict, SimpleDocumentFragment, SimpleNode } from '@glimmer/interfaces';
+import type { TrustedTypesWindow } from 'trusted-types/lib';
 
 export interface SafeString {
   toHTML(): string;
@@ -41,6 +42,18 @@ export function shouldCoerce(
 
 export function isEmpty(value: unknown): boolean {
   return value === null || value === undefined || typeof (value as Dict).toString !== 'function';
+}
+
+let isHTML: ((value: unknown) => boolean) | undefined;
+if (typeof window !== 'undefined') {
+  let trustedTypes = (window as unknown as TrustedTypesWindow).trustedTypes;
+  if (trustedTypes?.isHTML) {
+    isHTML = trustedTypes?.isHTML.bind(trustedTypes);
+  }
+}
+
+export function isTrustedHTML(value: unknown): boolean {
+  return isHTML ? isHTML(value) : false;
 }
 
 export function isSafeString(value: unknown): value is SafeString {
--- a/packages/@glimmer-workspace/integration-tests/package.json
+++ b/packages/@glimmer-workspace/integration-tests/package.json
@@ -29,6 +29,7 @@
     "@simple-dom/document": "^1.4.0",
     "@simple-dom/serializer": "^1.4.0",
     "@simple-dom/void-map": "^1.4.0",
+    "@types/trusted-types": "^2.0.7",
     "js-reporters": "^2.1.0",
     "qunit": "^2.19.4",
     "simple-html-tokenizer": "^0.5.11"
--- a/packages/@glimmer/runtime/package.json
+++ b/packages/@glimmer/runtime/package.json
@@ -43,7 +43,8 @@
     "@glimmer/util": "workspace:^",
     "@glimmer/validator": "workspace:^",
     "@glimmer/vm": "workspace:^",
-    "@glimmer/wire-format": "workspace:^"
+    "@glimmer/wire-format": "workspace:^",
+    "@types/trusted-types": "^2.0.7"
   },
   "devDependencies": {
     "@glimmer-workspace/build-support": "workspace:^",
--- a/packages/@glimmer/vm/lib/content.ts
+++ b/packages/@glimmer/vm/lib/content.ts
@@ -7,4 +7,5 @@ export const ContentType = {
   Fragment: 5,
   Node: 6,
   Other: 8,
+  TrustedHTML: 9,
 } as const;


NullVoxPopuli:
Thanks for submitting this, this is neat!

crypto:
Nice, "All checks have passed"

NullVoxPopuli (approved):

NullVoxPopuli:
Of note, 
- `TrustedHTML` is (at this time) not supported by FireFox and Safari
  - Compatibility: https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML#browser_compatibility
  - Polyfill: https://github.com/w3c/trusted-types

packages/@glimmer/runtime/lib/dom/normalize.ts
```diff
@@ -43,6 +44,18 @@ export function isEmpty(value: unknown): boolean {
   return value === null || value === undefined || typeof (value as Dict).toString !== 'function';
 }
 
+let isHTML: ((value: unknown) => boolean) | undefined;
+if (typeof window !== 'undefined') {
+  let trustedTypes = (window as unknown as TrustedTypesWindow).trustedTypes;
```
NullVoxPopuli, line 49: this is a good way to progressively allow the feature for browsers that have it and not break in browsers that don't

packages/@glimmer-workspace/integration-tests/test/trusted-html-test.ts
```diff
@@ -0,0 +1,49 @@
...
+      return policy?.createHTML(html);
+    });
+
+    this.render('<div>{{trustedHTML}}</div>');
+    this.assertHTML('<div><b>test\'""</b></div>');
+    this.assertStableRerender();
+  }
+
+  @test
+  'renders TrustedHTML in attribute context as string'() {
```
NullVoxPopuli, line 35: this looks like it's (expectedly) still escaped (which we want!, because we don't want `trustedHTML` adding attributes -- should the test name be updated to explain this expectation?

crypto: Theoretically this not ideal, trustedHTML potentially, can contain HTML entities that means developers will not notice that attributes double encoded. But I keep this similar to SafeString.
> because we don't want trustedHTML adding attributes
Yes, but I think ideally both SafeString and TrustedHTML should be HTML decoded in attribute value context and set as value raw (not escaping from attribute value context). 
> should the test name be updated to explain this expectation
Yes, I'll be more verbose in test.

NullVoxPopuli: In the discord, in #dev-glimmer-vm, @chancancode suggested that this scenario should actually _error_, because TrustedHTML isn't meant for attributes, so the fact it can end up here is likely is a mistake at best, and we should help steer folks towards correct usage

crypto: I agree with that statement, but I don't want to make that radical decision by myself in this PR. I know that it would create a ton of unnecessary work to fix every instance of sanitized values used in attribute context.

packages/@glimmer/runtime/lib/compiled/opcodes/content.ts
```diff
@@ -87,7 +96,11 @@ APPEND_OPCODES.add(Op.AppendHTML, (vm) => {
   let reference = check(vm.stack.pop(), CheckReference);
 
   let rawValue = valueForRef(reference);
-  let value = isEmpty(rawValue) ? '' : String(rawValue);
+  let value = isEmpty(rawValue)
+    ? ''
+    : isTrustedHTML(rawValue)
+      ? (rawValue as string)
```
chancancode, line 102: This cast seems suspect

chancancode: I would think, if we made it here, we should have already checked and unwrapped the value?

packages/@glimmer/runtime/lib/dom/normalize.ts
```diff
@@ -43,6 +44,18 @@ export function isEmpty(value: unknown): boolean {
...
+  if (trustedTypes?.isHTML) {
+    isHTML = trustedTypes?.isHTML.bind(trustedTypes);
+  }
+}
+
+export function isTrustedHTML(value: unknown): boolean {
```
chancancode, line 55: `function isTrustedHTML(value: unknown): value is TrustedHTML` probably?

packages/@glimmer/runtime/lib/compiled/opcodes/content.ts
crypto: We can modify whole chain of functions to accept `string | TrustedHTML` type, but I thought we should keep it when decide to fully migrate glimmer to Trusted Types.

packages/@glimmer/runtime/lib/dom/normalize.ts
crypto: I really wanted to do that, and it is related to previous comment, we will need to modify whole chain of functions. I can create different PR to specifically that.

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
```diff
@@ -74,6 +74,11 @@ export function StdAppend(
         op(Op.AppendSafeHTML);
       });
 
+      when(ContentType.TrustedHTML, () => {
+        op(Op.AssertSame);
+        op(Op.AppendHTML);
```
chancancode, line 79: Yeah, I see, (mis)using `AppendHTML` here is what caused you to need that unsafe `as string` cast in the other spot. It could probably be DRY-ed up a bit but the easiest thing to do would be just copy the `AppendSafeHTML` opcode into `AppendTrustedHTML` and use that here. 

The more ideal thing is to insert a step in between `Op.AssertSame` and `Op.AppendHTML` that pops the value of the stack, unwraps it, and puts the string value back, that way, `Op.AppendSafeHTML` doesn't need to exist and everything can just use `Op.AppendHTML`

packages/@glimmer/runtime/lib/dom/normalize.ts
crypto: Maybe keep it until more browsers support Trusted Types. Firefox planned to do that.

NullVoxPopuli: >  Firefox planned to do that.

Do you have a reference? I'd like to update https://github.com/glimmerjs/glimmer-vm/pull/1567#issuecomment-1957989082 with the status of implementation and all that

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
crypto: I've reused AppendHTML because we don't make modifications to TrustedHTML unlike SafeString. It reflects behavior similar to 
```
foo.innerHTML = trustedHTML
```
we use same DOM methods for 'string' and TrustedHTML, if not casting to string we even wouldn't need this PR.

> The more ideal thing is to insert a step in between Op.AssertSame and Op.AppendHTML that pops the value of the stack, unwraps it, and puts the string value back, that way, Op.AppendSafeHTML doesn't need to exist and everything can just use Op.AppendHTML

that scares me a bit, I don't know engine good enough to make these changes.

packages/@glimmer/runtime/lib/dom/normalize.ts
crypto: https://github.com/mozilla/standards-positions/issues/20#issuecomment-1853427823

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/stdlib.ts
crypto: > So your intention is to actually pass the value all the way through to the DOM functions so it works in whatever stricter environment that would have errored if you do innerHTML(string)

exactly, simply don't break our html we want to render

> SSR/SimpleDOM

It should work with polifill, I also don't think this is concern of Glimmer. Glimmer doesn't create TrustedHTML objects and doesn't try to render them, it simply doesn't modify them.

packages/@glimmer/runtime/lib/dom/normalize.ts
NullVoxPopuli: thank you!

packages/@glimmer/runtime/lib/compiled/opcodes/content.ts
crypto: I've added  TrustedHTML for appendHTML methods, let's see if we have any issues

chancancode: Yeah, that's what I am trying to say, if you follow the types it'll lead you to the places where this doesn't just work. We can certainly consider changing/adding the requirement to simple-dom, we rarely do, but if this is a platform feature I think we can make a case for it; ultimately if you need it to make it all the way to the DOM (to satisfy CSP requirements, for example), then I think that's what this would entail. To just make it not error, we can unwrap it, but sounds like that may not be enough for you.

crypto: https://github.com/ember-fastboot/simple-dom/blob/master/packages/%40simple-dom/interface/index.d.ts#L99
html should be unknown, not string. Since insertAdjacentHTML can accept any value and stringify it.

chancancode: simple-dom (as the README pointed out) isn't trying to model the entire DOM API, it's trying to model a small subset of the functionality that we actually need, as little as possible to make it work for our SSR use cases, it deliberately does not try to support all the features

chancancode: For example, it also doesn't have `.innerHTML = `, because just `insertAdjacentHTML` was enough to cover all the use cases we have (and in this case, it's a blessing because otherwise we would have to implement the TrustedHTML support in multiple places)

crypto: that passed validation without modification of 
https://github.com/glimmerjs/glimmer-vm/blob/1cb493b128a9c6704482a1775a478aac20259935/packages/%40glimmer/runtime/lib/dom/operations.ts#L69
Do you want me to do other changes?

chancancode: Yes, we definitely still have to plumb this through to simple-dom. I am actually not sure why TypeScript is not catching the TypeError, but essentially the problem is you changed the interface without changing the implementations, so for example, in places like these, it's all going to be problematic:

https://github.com/glimmerjs/glimmer-vm/blob/main/packages/%40glimmer/runtime/lib/dom/operations.ts#L69-L74

https://github.com/glimmerjs/glimmer-vm/blob/1cb493b128a9c6704482a1775a478aac20259935/packages/%40glimmer/node/lib/node-dom-helper.ts#L23-L25

The code is a bit odd, because the primary implementation of `insertHTMLBefore` is on a class that doesn't claim to implement the interface you modified, so no type error there. But this extends and implement it, so it should be caught there, but it wasn't: https://github.com/glimmerjs/glimmer-vm/blob/main/packages/%40glimmer/runtime/lib/dom/api.ts#L18

Anyway missing type errors aside, there are some real, a bit tedious but solvable problems, there are also a few more things to think through around attributes (and then the grunt work to actually plumb it through, get simple-dom released, etc), so it may be a while before we can land this.

crypto: I've changed it back to not change interface of insertHTMLBefore.

crypto: If you think it will take a while, I suggest to taking small steps and in this PR just stringify TrusteHTML, but render it as HTML.

crypto: > more things to think through around attributes
Let me explain why I'm against error on rendering TrustedHTML on attribute context.

I've tested LinkedIn where we use SafeString in attribute context. Almost everywhere, it's related to our i18n system. Every i18n string is like small HTML template where all params HTML encoded, not every i18n string contains HTML tags, but potentially every string can contain HTML entities, like `&nbsp;`. For this reason i18n helper uses htmlSafe for every i18n string. And obviously same htmlSafe strings used in attribute context. If wea are going to replace SafeString with TrustedHTML it will not change anything. Throwing errors would completely prevent migration to TrustedHTML.

I would recommend to keep same behavior as SafeString, and in best case don't stringify TrusteHTML object.

Another point is DOM API, some attributes require TrustedHTML to work, for example iframe.srcdoc
```
trustedTypes.getAttributeType('iframe', 'srcdoc') // TrustedHTML
```
https://developer.mozilla.org/en-US/docs/Web/API/TrustedTypePolicyFactory/getAttributeType

NullVoxPopuli: Thanks for walking through those use cases. That's helpful, and I'm fine with not throwing

chancancode:
I discussed this with @wycats and here is a summary of the things we discussed.

In general:
* It's great that the platform feature is gaining traction and it probably supersedes the Ember side RFCs
* Indeed eventually this can be something that the browsers handle for us and we won't have to worry about it, some day we can deprecate `SafeString`
* That being said, it's probably still early/premature to RFC the `SafeString` deprecation or even recommend using trusted types to users generally, as it is only implemented by one browser and still going through changes
* But it makes sense to unblock (not get in the way of) those who are adopting the feature early

Research:
* The prevailing consensus is that as `expression` is not a thing anymore, CSS is no longer considered an _injection_ (XSS) vector, but it is still very much an _exfiltration_ vector
* The current proposal decided that for v1, [it will only cover XSS](https://github.com/w3c/trusted-types/issues/104#issuecomment-514329496), and does not currently do anything for exfiltration and other vectors
* It is [planned](https://github.com/w3c/trusted-types/issues/241) that the feature can be expanded to cover other surfaces down the road with additional CSP opt-ins

For content position:
* It makes sense to allow `{{{this.someTrustedHTML}}}`
* To make that work, it will require plumbing it through to `@simple-dom/interfaces`:
  * Introduce a branded/opaque types for `TrustedHTML`, and any other trusted types we ended up using
  * `insertAdjacentHTML` needs to accept `TrustedHTML`
  * Introduce a `SimpleTrustedTypes` – a subset interface of `TrustedTypePolicyFactory` with the corresponding `is*` methods
  * Implementations need to be updated to match (specifically, SSR/rehydration) to account for those
* One additional challenge is that we currently we assume we can just pass around the `SimpleDocument` and have all the DOM APIs we need, with this change that is no longer the case, we need to somehow be able to accept and pass around `SimpleTrustedTypes` as well
  * We need to figure out how that is supposed to be configured (both here and in Ember)
  * In particular, that is relevant for fastboot/SSR, if we don't do this correctly all your `{{{this.someTrustedHTML}}}` will ended up being escaped

For attributes:
* Because the trusted types stringifies into their underlying values, for the most part, this already "work"
* ...except in the few cases where the v1 feature currently cares, i.e. `getAttributeType(...) !== null`, or `getPropertyType(...) !== null`
* For these to really work, similarly we need to plumb the logic to `@simple-dom/interfaces`:
  * We need to add `getAttributeType()` and `getPropertyType()` to `SimpleTrustedTypes`
  * `setAttribute`/`setAttributeNS` needs to accept the trusted types
  * Implementations need to be updated to match (specifically, SSR/rehydration) to account for those
* We have similar logic [here](https://github.com/glimmerjs/glimmer-vm/blob/main/packages/%40glimmer/runtime/lib/vm/attributes/dynamic.ts), we can consider deferring to `SimpleTrustedTypes.getAttributeType()`, however
  * There may be cases where we don't currently require a `SafeString` but `getAttributeType()` returns non-null, these are arguably cases that are currently missed but it may be too much of a breaking change to just start requiring them
  * Conversely, we currently cover a wider surface than the V1 API: for example, we have a warning for `styles={{...}}` that can be silenced with a `SafeString`
    * We are _not_ going to allow `TrustedHTML` to bypass the warning – the V1 API does not intend to provide any protection against the exfiltration vectors, if we allow `TrustedHTML` to bypass the style warning, we are conferring additional meaning to the type that the spec does not intend
    * If the spec expands to cover the exfiltration vectors, it may (and I would argue likely) introduce/require new trusted types (e.g. `TrustedCSS`) for these positions, so it would actually be incorrect to encourage developers to use `TrustedHTML` for this purpose today
    * **So, `TrustedHTML` is not going to be a complete replacement of `SafeString` today**. There are cases where you currently need a `SafeString` today that you can't simply replace with a `TrustedHTML` (or any of the other trusted types)

So overall, we think this is certainly a worthwhile feature to pursue, it's a bit of a project to get it implemented correctly, but a fairly bounded one.

If you think it would help enough to land a version where we simply stringify, and if you think that helps enough with your use case (keep in mind that it will still trigger CSP warnings because it won't make it all the way to the DOM APIs), we can certainly explore that. However, I'm not sure that would be much less work than doing everything correctly – the PR as it stands is still problematic, because we can't just rely on accessing the trusted types on the global object in SSR/fastboot, and if we are going to fix that we may as well do everything else?

Does LinkedIn not need this to work in SSR/fastboot?

chancancode:
> Internally we will be using object similar to SafeString or Trusted Types polyfill, I don't see issues with SSR. We will make sure that it works.

I don't think you can. FastBoot/SSR runs in a [sandboxed environment](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/packages/fastboot/src/sandbox.js#L15-L39), so as it stand, what would happen is that there will be no `window.TrustedType` in SSR, and `isHTML()` will always return false, and so if you try to render `{{this.someTrustedHTML}}` there it will get escaped contrary to what you want.

NullVoxPopuli:
in fastboot's sandboxed environment, you could provide an implementation for `TrustedHTML`, yeah?

is this feasible? or not worth it? :shrug:

```js
./config/fastboot.js
module.exports = function () {
  return {
    buildSandboxGlobals(defaultGlobals) {
      return {
        ...defaultGlobals,
        fetch,
        AbortController,
  ReadableStream:
          typeof ReadableStream !== 'undefined'
            ? ReadableStream
            : require('node:stream/web').ReadableStream,
        WritableStream:
          typeof WritableStream !== 'undefined'
            ? WritableStream
            : require('node:stream/web').WritableStream,
        TransformStream:
          typeof TransformStream !== 'undefined'
            ? TransformStream
            : require('node:stream/web').TransformStream,
        Headers: typeof Headers !== 'undefined' ? Headers : undefined,
        // custom stuff here
        TrustedHTML: YourImplementationHere       
      }
    }
  }
};
```

crypto:
I just planned to use this code to generate TrustedHTML objects:

```
let createTrustedHTML: (s: string) => TrustedHTML;
let isTrustedHTML: (s: unknown) => boolean;
const trustedTypes = window.trustedTypes;
if (trustedTypes?.createPolicy && trustedTypes?.isHTML) {
  const policy = trustedTypes.createPolicy('jSecure', {
    createHTML: (s: string) => s
  });
  createTrustedHTML = (s: string) => policy.createHTML(s);
  isTrustedHTML = (value: unknown) => trustedTypes.isHTML(value);
} else {
  // if Trusted Types are not available, create object similar to Ember.js SafeString
  class HTML extends String {
    toHTML(): string {
      return this.toString();
    }
  }
  createTrustedHTML = (s: string) => new HTML(s) as unknown as TrustedHTML;

  isTrustedHTML = (value: unknown): boolean => {
    return (
      value !== null &&
      typeof value === 'object' &&
      'toHTML' in value &&
      typeof (value as HTML).toHTML === 'function'
    );
  };
}
```

chancancode:
It think it should probably get passed here: https://github.com/ember-fastboot/ember-cli-fastboot/blob/242683f4137ca3dbcd53fdc0d8cc65f4055e3b57/packages/fastboot/src/ember-app.js#L362-L374

Along side the `document: new SimpleDOM.Document()` would be `trustedTypes: { ... }`

chancancode:
> is this feasible? or not worth it? 🤷

It's not that different from the plan I laid out, the main difference is that we would only define a strict subset in `Simple.TrustedType`, so you wouldn't have be implement the parts that we don't care about, and it won't force you to assign that to the global, because your polyfill may be incomplete, may not work with other libraries, etc.

Which is the same reason why you don't just set `document` to `new Simple.Document()` in that hook, because `Simple.Document` is missing a lot of things that are on the real `Document`, and if you make it globally available, it could break other code/library that tries to feature-detect `document` and do stuff with it.

crypto:
Asking some experts about CSS exfiltration, I believe style attribute is not that bad. And htmlSafe doesn't provide security benefits there.
https://twitter.com/shafigullin/status/1761166166687617133
https://portswigger.net/research/blind-css-exfiltration

chancancode:
That would be a pretty strong claim with a pretty high bar to clear, and there are additional considerations for client side JS apps.

Off the top of my head, a scenario would be that you have `style={{...}}` on a conditionally rendered element, then you can use a background image url to determine the fact that the element was rendered. At minimum, that could convey the fact that a e-mail/message was opened/read, or inferring what kind of access a user has, etc.

Besides exfiltration, it is also a phishing vector, e.g. re-positioning a comment/message box directly over a password field, hiding important warning text, that kind of stuff.

We can certainly discuss whether we originally meant to protect users against that kind of threat, or was it just meant for XSS because expressions were a thing back then. My recollection was that it was for both, and it was added in response to (or at least around the same time) industry experts calling out this kind of vectors in Angular/other frameworks. But in any case, that is not a small change that we are just going to make in a PR without an RFC.

crypto:
I'm checking Discord conversation
https://discord.com/channels/480462759797063690/480502165434138624/1209976708082696222
> but also, CSP would also have blocked innerHTML = "<script>...</script>

For this specific reason CSP have additional directives:
[script-src-elem](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src-elem)
[script-src-attr](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src-attr)

wycats:
Some big-picture thoughts:

## The Status Quo

Today, Ember attempts to address certain potential security risks via the `htmlSafe` mechanism.

### XSS Mitigations

The recommended approach for XSS is:

1. If at all possible, insert user content into the page using a template, which constructs HTML by creating DOM nodes rather than through string concatenation. This avoids XSS hazards caused by malicious user content using HTML characters (like `>` or `"`) to escape the expected HTML context. 
2. If it's necessary to create an HTML string that contains user content, concatenate your own literals (e.g. `<span>`) with **escaped** user content (`escapeExpression(userContent)`). Then, mark the entire string as `htmlSafe`, which will allow it to be inserted into content positions in a template.

This approach maps onto security best practices that predate trusted types. It's not a confusion-free approach, but it's straight-forward once you understand what's going on, and it's pretty straight-forward for npm packages to use the approach to create helpers that produce safe HTML.

### Exfiltration and Clickjacking

Ember uses the `htmlSafe` mechanism for another purpose: to mark content that will be used as part of the `style` attribute as trusted.

Concatenating malicious user content into the `style` attribute can result in clickjacking attacks (like [this one](https://www.scmagazine.com/news/style-sheet-vulnerability-allowed-attacker-to-hijack-linkedin-pages) that affected LinkedIn in 2015).

This mitigation is less effective than the mitigation against XSS because:

- It does not offer a pattern for escaping user content before marking a string as trusted, which leads many developers to mark strings as trusted to "make the error go away".
- It is a [defense-in-depth](https://www.wikiwand.com/en/Defense_in_depth_(computing)) strategy rather than a complete mitigation.

We may want to reconsider this mitigation, but it would be inappropriate (in my opinion) to weaken a Defense in Depth strategy simply because we're strengthening an unrelated mitigation.

## Trusted Types

Trusted Types offer a way to address XSS (but [not exfiltration or clickjacking][1]).

Before Trusted Types, it was possible to eliminate sources of XSS vulnerabilities through CSP opt-ins. These opt-ins work in a blunt-force way: they disable all of the DOM "sinks" that can result in the execution of dynamic code.

Trusted Types build on this foundation by allowing applications to mark specific dynamic strings as trusted, allowing them to be passed to DOM sinks that execute code even when stricter CSP settings are enabled.

(The API is carefully crafted to require the application author (the person who is responsible for the HTML code that bootstrapped the app) to have a direct hand in marking a particular string as trusted.)

Ultimately, this should result in wider adoption of stricter CSP settings in situations where a small amount of dynamic content (e.g. localization bundled fetched as JSON) currently prevent adoption.

## How Ember Should Adopt Trusted Types

Ember's XSS mitigations behave differently from Trusted Types: they work even in the absence of strict CSP settings. In a sense, Ember's templating syntax emulates strict CSP settings, but only when constructing and inserting DOM through Ember templates.

First, a quick comparison of the features.

| feature | With CSP | Without CSP | Style Mitigation |
| --- | --- | --- | --- |
| Trusted Types | high-fidelity | inactive | out of scope |
| `htmlSafe` | imperfect | still active | limited, defense in depth |

In my opinion, we want the following changes to Glimmer:

1. When a piece of Glimmer syntax corresponds to a Trusted Types sink (e.g. `{{{` corresponds to `insertAdjacentHTML`), Glimmer should accept the appropriate Trusted Type in addition to a SafeString. This means that the attributes specified as requiring `TrustedScriptURL` should accept `TrustedScriptURL` (but not `TrustedHTML`!) in addition to `SafeString`.
2. When a string is marked `htmlSafe`, Glimmer should **not** turn it into a Trusted Type. This is both because Trusted Types are more granular than `htmlSafe` *and* because the current usage of `htmlSafe` is not well-defined enough to satisfy the security assumptions of a security team that wants to enable strict CSP rules but allow Trusted Types.
3. When Glimmer syntax accepts a SafeString in situations that are out of scope for Trusted Types, we should **not** accept `TrustedHTML` or other Trusted Types. This is because marking a string as Trusted HTML explicitly does **not** imply that it should be trusted in other contexts.


| Glimmer Syntax | Accepts `htmlSafe`? | Accepts Trusted Type? |
| --- | --- | --- |
| Corresponds to a Trusted Type sink | Yes, with CSP Disabled[^2] | Yes |
| Does not Correspond to a Trusted Type Sink | Yes | No |

[1]: https://www.w3.org/TR/trusted-types/#non-goals
[2]: When CSP is enabled, any Glimmer syntax that corresponds to a Trusted Types sink will produce an error unless the string is marked as a Trusted Type. This is what we want.

crypto:
@chancancode is anything else required for this PR?

chancancode:
…yes?

chancancode:
I believe the previous comments/reviews still reflect the status quo, has anything changed, or am I missing/misremembering something? Do you want to summarize what you see as the current status and how it aligns with the direction we seem to all wait, and maybe we can find the disconnect that way?

crypto:
> Do you want to summarize what you see as the current status
yes, are we going to render TrustedHTML as HTML or not?

PR #1568, opened 2024-02-23
Cleanup v1 AST

chancancode:
See the commit message for each individual commits


packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
```diff
@@ -463,10 +463,6 @@ export function preprocess(
 
   let program = new TokenizerEventHandlers(source, entityParser, mode).acceptTemplate(ast);
 
-  if (options.strictMode) {
-    program.blockParams = options.locals ?? [];
-  }
```
chancancode, line 468: Do we know if this something AST consumers actually rely on? (Specifically, the "block params" on a root level `Template` node has the `locals` name that were passed here) It seems like it would be incomplete/unreliable at best, and it wasn't used by any tests, so presumably we don't rely on it ourselves (we don't even read it)

packages/@glimmer/syntax/lib/v1/visitor-keys.ts
```diff
@@ -35,11 +34,6 @@ const visitorKeys = {
   Hash: ['pairs'],
   HashPair: ['value'],
   BlockParam: [],
-
-  // v2 new nodes
-  NamedBlock: ['attributes', 'modifiers', 'children', 'comments'],
-  SimpleElement: ['attributes', 'modifiers', 'children', 'comments'],
-  Component: ['head', 'attributes', 'modifiers', 'children', 'comments'],
```
chancancode, line 42: I "ensured these stays in sync with typing" as instructed by the comment on line 3

I don't think these are meant to be here, v2 nodes won't appear in the v1 AST and defining visitors for them doesn't do anything, and in any case, these aren't even in sync with what is currently in the "v2" AST

packages/@glimmer/syntax/lib/generation/printer.ts
```diff
@@ -320,7 +320,7 @@ export default class Printer {
       return;
     }
 
-    this.buffer += mustache.escaped ? '{{' : '{{{';
+    this.buffer += mustache.trusting ? '{{{' : '{{';
```
NullVoxPopuli, line 323: good change

packages/@glimmer/syntax/lib/parser/tokenizer-event-handlers.ts
```diff
@@ -461,21 +441,24 @@ export function preprocess(
...
+  );
 
-  if (options.strictMode) {
-    program.blockParams = options.locals ?? [];
+  if (options.strictMode && options.locals?.length) {
+    template = b.template({ ...template, locals: options.locals });
   }
 
-  if (options && options.plugins && options.plugins.ast) {
+  if (options.plugins && options.plugins.ast) {
```
NullVoxPopuli, line 453:
```suggestion
  if (options.plugins?.ast) {
```
:shrug:

packages/@glimmer/syntax/lib/traversal/walker.ts
```diff
@@ -33,40 +33,21 @@ export default class Walker {
...
       case 'Block':
       case 'Template':
-        return visitors.Program(this, node as unknown as ASTv1.Program, callback);
+        walkBody(this, node.body, callback);
+        return;
       case 'ElementNode':
-        return visitors.ElementNode(this, node, callback);
+        walkBody(this, node.children, callback);
+        return;
       case 'BlockStatement':
```
NullVoxPopuli, line 41: TIL:
```hbs
{{#if a}}

  hi
{{else if b}}
  hello
{{else}}
  hola
{{/if}}
```

everything that's not the _first_ block is in "inverse"
image

kinda good :see_no_evil:

packages/@glimmer/syntax/test/parser-node-test.ts
```diff
@@ -285,47 +343,36 @@ test('Element modifiers', () => {
   );
 });
 
-test('Element paths', (assert) => {
```
NullVoxPopuli, line 288: I didn't see a replacement test for this scenario -- why was the test removed? (or did I miss an equiv test?)

packages/@glimmer/syntax/test/loc-node-test.ts
```diff
@@ -134,29 +138,6 @@ test('html elements', () => {
   }
 });
 
-test('html elements with paths', () => {
-  let ast = parse(`
-    <Foo as |bar|>
-      <bar.x.y class='bar'/>
```
NullVoxPopuli, line 140: I didn't see a replacement test for this scenario -- why was the test removed? (or did I miss an equiv test?)

chancancode: Because I only pushed the part one of two (the block statement part) not the element one so far

patricklx:
I'm liking the changes. Much better to do it at tokenizer/parser level

gwak (commented):
The deprecation warnings introduced here seems to be triggered In Ember projects when templates are compiled. I get thousands of lines like the following when building:
```sh
DEPRECATION: The original property on literal nodes is deprecated, use value instead
DEPRECATION: The original property on literal nodes is deprecated, use value instead
DEPRECATION: The original property on literal nodes is deprecated, use value instead
``` 

Which seems to be triggered by Boolean literals like `{{true}}`

NullVoxPopuli:
Can you provide the output of `pnpm why @glimmer/syntax`?

Today, when you have that log spew, it usually mean an addon is out of date, or a custom internal transform hasn't yet been updated to support both the deprecated api and the new api.

gwak:
> pnpm why @glimmer/syntax

Hi @NullVoxPopuli , thanks for answering, here's the output of the command attached (it's quite big)
[glimmer-syntax.txt](https://github.com/user-attachments/files/18846864/glimmer-syntax.txt)

NullVoxPopuli:
You have a wide range of `@glimmer/syntax` versions.
To solve this deprecation, you'll want to:
1. find where it's come from (can override console.xyz and place a debugger to find out)
2. upgrade dependencies that support the newer `@glimmer/syntax`
3. if the deprecation comes from an internal transform, adapt the transform to account for multiple `@glimmer/syntax` versions.
4. An option -- force newer versions via overrides

gwak:
> 1. find where it's come from (can override console.xyz and place a debugger to find out)

Ok, it seems to be coming from `ember-ref-bucket` addon, see here:
https://github.com/lifeart/ember-ref-bucket/blob/dc3aa0d227dc37f56c3e70f32048accfcdb1ee06/lib/ref-transform.js#L32

I'll try to propose a PR soon. As always, thanks for your help @NullVoxPopuli !

gwak:
@NullVoxPopuli PR submitted here https://github.com/lifeart/ember-ref-bucket/pull/72

PR #1573, opened 2024-03-08
Refactor on modifier implementation

chancancode:


packages/@glimmer/runtime/lib/modifiers/on.ts
```diff
@@ -239,7 +298,7 @@ function addEventListener(
   @method on
   @public
 */
-class OnModifierManager implements InternalModifierManager<OnModifierState | null, object> {
+class OnModifierManager implements InternalModifierManager<OnModifierState, object> {
```
NullVoxPopuli, line 301: it's so nice now!!

PR #1557, opened 2024-02-01
[ENHANCEMENT+CLEANUP] Better compile time errors for strict mode

chancancode:
Removing support for implicit this fallback {{foo}} -> {{this.foo}}

When a strict mode template references an undefined variable, it really should produce a compile-time error, arguably that is the point of the feature.

However, it seems like previously a lot of these errors ended up propagating and is actually handled by the runtime compiler.

The original/primary goal here started out as cleaning up all the code for implicit this fallback, but as I delete code it eventually revaled all the places where strict mode was inappropiately using the same infrastructure ("unresolved free variables", which isn't really a sensible thing).

Also:

* Removes `{{#with}}`
* Removes "deprecated call" @foo={{bar}}
* Removes unused opcodes
* Syncs keyword list with Ember

PR #1574, opened 2024-03-09
[CLEANUP] Drop support for `{{#with}}` keyword

chancancode:
This was dropped on Ember side a while back, just never cleaned up

Ironically this probably meant when Ember dropped the deprecation/ error, nothing was actually stopping `{{#with}}` from anymore, so perhaps for a while we have actually _re-introduced_ support for the syntax for some time?

Extracted from #1557


chancancode:
> Ironically this probably meant when Ember dropped the deprecation/ error, nothing was actually stopping {{#with}} from anymore, so perhaps for a while we have actually re-introduced support for the syntax for some time?

Yeah,I think we actually did, just tried in on 5.4 and `{{#with}}` certainly works there.

Specifically, in here: https://github.com/emberjs/ember.js/pull/19632

This is partially why, among other reasons, when removing a deprecated feature, we typically want to swap the `deprecate` call for an `assert` call for at least 1-2 LTS versions.

Issue #1564, opened 2024-02-16
Declare license in package.json of published packages


jelhan:
The license is only declared in `package.json` at workspace root but not in the `package.json` of the published packages. Would be great if we could add it their as well as many tools rely on that information (e.g. `pnpm licenses ls`).

jelhan:
Likely a duplicate of #1463.

NullVoxPopuli:
A PR to improve would be most welcome!

jelhan:
Thanks a lot @andreyfel!

PR #1579, opened 2024-03-20
Add missing licenses

### LINKED ISSUE #1463: License metadata for v0.85 packages is missing
We use [licensee](https://www.npmjs.com/package/licensee) to check license compatibility of all our dependencies in Discourse. The recent `0.85` version bump of glimmer packages is causing [licensee checks to fail](https://github.com/discourse/discourse/pull/24085) on `@glimmer/syntax` and `@glimmer/interfaces` because it is unable to find valid version Metadata.

This is also evident on NPM, where [v0.84.3](https://www.npmjs.com/package/@glimmer/syntax/v/0.84.3) correctly shows "License: MIT", while [v0.85.3](https://www.npmjs.com/package/@glimmer/syntax/v/0.85.3) shows "License: none".

I'm not sure what has changed here. It doesn't look like [the 0.84.3 package.json](https://github.com/glimmerjs/glimmer-vm/blame/v0.84.3/packages/%40glimmer/syntax/package.json) files had a `license` key either... so something (related to the release process?) must have been magically inserting the license metadata in the past. 🤔

andreyfel:
to all package.json files which do not have `"private": true`

Fixes #1463 and #1564.

### DIFF
--- a/packages/@glimmer/compiler/package.json
+++ b/packages/@glimmer/compiler/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@glimmer/compiler",
   "version": "0.89.0",
+  "license": "MIT",
   "repository": "https://github.com/glimmerjs/glimmer-vm/tree/main/packages/@glimmer/compiler",
   "type": "module",
   "main": "index.ts",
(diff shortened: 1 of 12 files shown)


NullVoxPopuli:
thank you!!

PR #1589, opened 2024-04-30
support modifier delegate manager in debug render tree

patricklx:
im currently trying out ember beta which has the included modifier support for inspector, but i'm not getting an instance for e.g. did-insert modifier.
 think CustomModifierManager.getDebugInstance should return the delegate

### DIFF
--- a/packages/@glimmer/manager/lib/public/modifier.ts
+++ b/packages/@glimmer/manager/lib/public/modifier.ts
@@ -125,8 +125,8 @@ export class CustomModifierManager<O extends Owner, ModifierInstance>
     }
   }
 
-  getDebugInstance({ modifier }: CustomModifierState<ModifierInstance>) {
-    return modifier;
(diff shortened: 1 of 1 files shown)


patricklx:
@chancancode @NullVoxPopuli

NullVoxPopuli (approved):

NullVoxPopuli:
From @chancancode :  

> [rather than `|| delegate`] we need to add a hook that custom modifier managers can implement

patricklx:
Would the hook need to be defined e.g. here: https://github.com/ember-modifier/ember-modifier/blob/ca6e654dd822f33e0ac82598ec33f61d835ccce1/ember-modifier/src/-private/class-based/modifier-manager.ts#L56
?

Wouldn't that mean that everyone would need to update their implementation?
Or update their dependency on ember/modifier?

PR #1591, opened 2024-06-25
fix arrow modifiers debug name

patricklx:
currently arrow functions or functions without name as displayed with their whole implementation

### DIFF
--- a/packages/@glimmer/manager/lib/public/modifier.ts
+++ b/packages/@glimmer/manager/lib/public/modifier.ts
@@ -119,9 +119,9 @@ export class CustomModifierManager<O extends Owner, ModifierInstance>
 
   getDebugName(definition: object) {
     if (typeof definition === 'function') {
-      return definition.name || definition.toString();
+      return definition.name || '<unknown>';
     } else {
-      return '<unknown>';
+      return definition.toString() || '<unknown>';
     }
   }
 


patricklx:
> what will this look like?

`{{my-modifier fn}}` is how it can look like.

Currently... look at you limber site :) it's a good example

NullVoxPopuli:
I see! 
image

looks like anything that fixes this is good for me

patricklx:
@NullVoxPopuli actually, current release of inspector cannot handle latest ember... need nightly:
Screenshot 2024-06-27 at 10 39 38

packages/@glimmer/runtime/lib/compiled/opcodes/component.ts
```diff
@@ -504,7 +504,7 @@ export class ComponentElementOperations implements ElementOperations {
       }
 
       let { element, constructing } = vm.elements();
-      let name = manager.getDebugName(definition.state);
+      let name = manager.getDebugName(state, definition.state);
```
NullVoxPopuli, line 507: Why definition.state here, if the argument is 'definition' (for modifiers)?

Is this an inconsistency between modifiers and components managers?

patricklx: From what i understand: Because state is from the 'main' manager, and definition.state is from the delegate. And state has the delegate needed to delegate to function call

NullVoxPopuli (approved):

wycats:
This looks good to me. Let's do it!

packages/@glimmer-workspace/integration-tests/lib/modifiers.ts
```diff
@@ -44,7 +44,7 @@ export class TestModifierManager
     return tag;
   }
 
-  getDebugName({ Klass }: TestModifierDefinitionState) {
+  getDebugName(state: TestModifierDefinitionState, { Klass }: TestModifierDefinitionState) {
```
chancancode, line 47: This change seems fishy, why is it passing the same argument twice?

packages/@glimmer/interfaces/lib/managers/modifier.d.ts
```diff
@@ -19,4 +19,6 @@ export interface ModifierManager<ModifierStateBucket> {
   installModifier(instance: ModifierStateBucket, element: Element, args: Arguments): void;
   updateModifier(instance: ModifierStateBucket, args: Arguments): void;
   destroyModifier(instance: ModifierStateBucket, args: Arguments): void;
+  getDebugName?(instance: object): string;
```
chancancode, line 22: Not sure if instance is just misnamed here, but if that is named correctly then it doesn't seem good, we shouldn't need to have created the instance just to get the name, we want to be able to use this statically (on the class, for error messages, etc)

packages/@glimmer/interfaces/lib/managers/modifier.d.ts
```diff
@@ -19,4 +19,6 @@ export interface ModifierManager<ModifierStateBucket> {
...
+  getDebugInstance?(instance: object): any;
```
chancancode, line 23: `: object` is fishy, probably should be `ModifierStateBucket` or something

packages/@glimmer-workspace/integration-tests/lib/modifiers.ts
patricklx: the type was the same, but not the object. anyway it was confusing, i changed it

packages/@glimmer/interfaces/lib/managers/modifier.d.ts
patricklx: now its ModifierStateBucket

packages/@glimmer/interfaces/lib/managers/modifier.d.ts
patricklx: right

packages/@glimmer/manager/lib/public/modifier.ts
```diff
@@ -32,13 +32,13 @@ export function modifierCapabilities<Version extends keyof ModifierCapabilitiesV
   });
 }
 
-export interface CustomModifierState<ModifierInstance> {
+export interface CustomModifierState<ModifierStateBucket> {
```
patricklx, line 35: renamed the type here to be the same as in modifier.d.ts file

packages/@glimmer/interfaces/lib/managers/internal/modifier.d.ts
```diff
@@ -22,8 +22,8 @@ export interface InternalModifierManager<
   // the modifier's update hooks need to be called (if at all).
   getTag(modifier: TModifierInstanceState): UpdatableTag | null;
 
-  getDebugName(Modifier: TModifierDefinitionState): string;
-  getDebugInstance(Modifier: TModifierInstanceState): unknown;
+  getDebugName(modifier: TModifierInstanceState): string;
```
chancancode, line 25: This chance (and everywhere else) is backwards – as I tried to explain previously we want `getDebugName` to be static (takes the definition instead of instance) so we can get the name without constructing it. See the component manager for parallel

patricklx: but the name could depend on the state returned from createModifier

patricklx: done. i think

packages/@glimmer-workspace/integration-tests/lib/modifiers.ts
```diff
@@ -44,8 +44,8 @@ export class TestModifierManager
     return tag;
   }
 
-  getDebugName({ Klass }: TestModifierDefinitionState) {
```
chancancode, line 47: `TestModifierDefinitionState` was correct

packages/@glimmer/interfaces/lib/managers/internal/modifier.d.ts
chancancode: doesn't look like it, did you forget to push or something?

packages/@glimmer/interfaces/lib/managers/modifier.d.ts
```diff
@@ -19,4 +19,6 @@ export interface ModifierManager<ModifierStateBucket> {
   installModifier(instance: ModifierStateBucket, element: Element, args: Arguments): void;
   updateModifier(instance: ModifierStateBucket, args: Arguments): void;
   destroyModifier(instance: ModifierStateBucket, args: Arguments): void;
+  getDebugName?(definition?: object): string;
```
chancancode, line 22: For this file, it's probably more appropriate to make this `factory: unknown` to match `createModifier`, since they are talking about the same thing. Not sure if it's missing a generic for the definition or if it's trying to be deliberately vague about it, but either way that problem predated your change so I'd just focus on matching those two here

packages/@glimmer/manager/lib/public/modifier.ts
```diff
@@ -32,13 +32,13 @@ export function modifierCapabilities<Version extends keyof ModifierCapabilitiesV
...
   tag: UpdatableTag;
   element: SimpleElement;
-  modifier: ModifierInstance;
-  delegate: ModifierManager<ModifierInstance>;
+  modifier: ModifierStateBucket;
+  delegate: ModifierManager<ModifierStateBucket>;
   args: Arguments;
-  debugName?: string;
+  definition?: object;
```
chancancode, line 41: I don't think we'd need to store it since it gets passed in?

packages/@glimmer/manager/lib/public/modifier.ts
```diff
@@ -99,41 +99,44 @@ export class CustomModifierManager<O extends Owner, ModifierInstance>
...
 
     return state;
   }
 
-  getDebugName(definition: object) {
-    if (typeof definition === 'function') {
-      return definition.name || definition.toString();
-    } else {
-      return '<unknown>';
+  getDebugName({ delegate, definition }: CustomModifierState<ModifierStateBucket>) {
```
chancancode, line 121: Something about this isn't right – this signature looks like it's still expecting an instance?

NullVoxPopuli:
Seems good 
image

packages/@glimmer/interfaces/lib/managers/internal/modifier.d.ts
NullVoxPopuli: all that changed is capitalization

packages/@glimmer/manager/lib/public/modifier.ts
NullVoxPopuli, line 41: is this still needed?

packages/@glimmer/manager/lib/public/modifier.ts
NullVoxPopuli: iirc, all the getDebugNames are passed the result from create

NullVoxPopuli (changes requested):
tiny cleanup thing

packages/@glimmer/manager/lib/public/modifier.ts
patricklx: it now receives the definition only

packages/@glimmer/manager/lib/public/modifier.ts
```diff
@@ -99,41 +98,44 @@ export class CustomModifierManager<O extends Owner, ModifierInstance>
...
     return state;
   }
 
   getDebugName(definition: object) {
-    if (typeof definition === 'function') {
-      return definition.name || definition.toString();
-    } else {
-      return '<unknown>';
+    const delegate = this.factory.prototype;
+    if (typeof delegate?.getDebugName === 'function') {
```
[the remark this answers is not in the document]
patricklx: yeah, I tried that, the problem is in https://github.com/glimmerjs/glimmer-vm/blob/10eae7429b702b1e7f5434b91802d5767ff7ad9a/packages/%40glimmer/runtime/lib/environment.ts#L68
there is no direct way to access the owner, maybe more changes would be requried?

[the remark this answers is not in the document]
chancancode: oof, sorry this has been *rough*. It's the code equivalent of random bureaucracy. Notionally we have the information, just not consistently threaded/easily accessible.

Since we narrowed this problem down to these private APIs (I think, this class is private, right?), I don't really care that much about the API surface.

I think the API surface of what we call the "delegate" here is important (that's what people actually implement right?), and I think we got that down pretty good.

As long as we don't affect the public API, we can do anything (that is type safe, etc) for this wrapper class. Maybe we keep `getDebugName(owner, definition)` and `getDebugNameFromInstance(instance)` in this internal class and call it a day for now. For reference: this is the kind of place we may want to use the former, which we _do_ have the owner: https://github.com/glimmerjs/glimmer-vm/blob/34888e9b0de56475c1c9fab88c4007816dd1bf44/packages/%40glimmer/runtime/lib/compiled/opcodes/dom.ts#L147-L153

For what its worth I think a lot of this up is overdue for a cleanup/rethink pass, especially for the manages stuff. The pure form of the idea is good, but we made some decisions along the way that incidentally complicated things a fair bit.

NullVoxPopuli (approved):

PR #1592, opened 2024-07-05
Partial revert of 64eb186f

davidtaylorhq:
This commit introduced a slight performance regression. Trying to revert all the functional changes to see what the impact is on the benchmark results


NullVoxPopuli:
what was the outcome?

davidtaylorhq:
CI wouldn't run here automatically (cos I'm not a contributer). Testing over in https://github.com/davidtaylorhq/glimmer-vm/pull/1 (it's still running)

Locally, this change gives me a 2-4% improvement in the benchmarks. So, not a complete explanation for the perf regression, but it's a step in the right direction.

PR #1596, opened 2024-08-15
Explore performance

NullVoxPopuli:
The goal of this PR is to try to see if we can undo the perf losses reported by
- https://github.com/glimmerjs/glimmer-vm/issues/1590

Issue #1590, opened 2024-06-13
Rendering performance regression between 0.84.3 and 0.85.1?


davidtaylorhq:
In Discourse, and [in Emberperf](https://github.com/eviltrout/ember-performance/pull/132), we saw a fairly significant rendering-performance hit as part of the Ember 5.5 -> 5.6 bump:

<img width="400" src="https://github.com/glimmerjs/glimmer-vm/assets/6270921/67d1d2bb-cc46-4379-9000-92817f0cca1a"/>

Ember 5.6 included a bump of glimmer-vm from 0.84.3 to 0.85.1 (https://github.com/emberjs/ember.js/pull/20561)

Unfortunately 0.84.3 -> 0.85.1 include a lot of structural changes in glimmer-vm, much of which was done without glimmer-vm's own performance testing in working order.

I was able to boot the glimmer-vm benchmark app on a handful of old commits, and run [tachometer](https://github.com/google/tachometer) on them to compare the 'render' `performance.measure` metric.

| commit | Avg time |
|--|--|
| 144133 (v0.84.3) | 29.19ms - 29.31ms |
| 99fee7 (first merge of pnpm) | 29.08ms - 29.21ms (no change) |
| 664a74 ('restore performance tests' - after a bunch of refactoring) [^1] | 39.14ms - 39.29ms |
| 41f152 (v0.86.0)  | 51.85ms - 52.01ms |

[^1]: with 56ddfa cherry picked on top to make the benchmark app work

These numbers are clearly going in the wrong direction.

I would love to be able to bisect into specific commits to identify what caused the regressions. Unfortunately, on all the intermediate commits I've tried, I've been unable to get the benchmark app to boot because of various import/dependency/package-json errors. It seems the '`perf.yml`' GitHub CI job was disabled for much of this time, so I assume this was a known problem on these commits, and not a problem with my local setup.

So... I don't really know where that leaves us. Does anyone have any pointers for what else we can do to isolate the source of the regression(s)?

davidtaylorhq:
I took a stab at reverting some of the extra assertions & iterator changes introduced in 64eb186f (see https://github.com/davidtaylorhq/glimmer-vm/pull/1). It does improve things by a couple of percentage points, but it's nothing compared to the 30-40% regression shown above 😢

davidtaylorhq:
It looks like a further rendering-speed regression has been released as part of Ember 5.10 😭 

[...]

PR #1597, opened 2024-08-15
Explore Performance 2

NullVoxPopuli:
Related:
  https://github.com/glimmerjs/glimmer-vm/pull/1596

The goal of this PR is to try to see if we can undo the perf losses reported by
- https://github.com/glimmerjs/glimmer-vm/issues/1590


NullVoxPopuli:
> duration phase estimated improvement -110ms 

woah, that's nice -- Suppose I should compare with the other versions

NullVoxPopuli:
The improvement revealed by tracerbench translates all they way through to ember.
image

image


image

PR #1598, opened 2024-08-16
Explore performance 3

NullVoxPopuli:
Related:
  - https://github.com/glimmerjs/glimmer-vm/pull/1596
  - https://github.com/glimmerjs/glimmer-vm/pull/1597

The goal of this PR is to try to see if we can undo the perf losses reported by
- https://github.com/glimmerjs/glimmer-vm/issues/1590

### DIFF
--- a/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
+++ b/packages/@glimmer/runtime/lib/compiled/opcodes/dom.ts
@@ -1,24 +1,19 @@
 import type {
   CapturedPositionalArguments,
   Environment,
+  Maybe,
   ModifierDefinition,
   ModifierDefinitionState,
   ModifierInstance,
   Nullable,
   Owner,
+  SimpleElement,
+  SimpleNode,
   UpdatingOpcode,
   UpdatingVM,
 } from '@glimmer/interfaces';
 import type { Reference } from '@glimmer/reference';
 import type { Revision, Tag } from '@glimmer/validator';
-import {
-  check,
-  CheckElement,
-  CheckMaybe,
-  CheckNode,
-  CheckOption,
-  CheckString,
-} from '@glimmer/debug';
 import { associateDestroyableChild, destroy, registerDestructor } from '@glimmer/destroyable';
 import { getInternalModifierManager } from '@glimmer/manager';
 import { createComputeRef, isConstRef, valueForRef } from '@glimmer/reference';
@@ -33,7 +28,6 @@ import { isCurriedType, resolveCurriedValue } from '../../curried-value';
 import { APPEND_OPCODES } from '../../opcodes';
 import { CONSTANTS } from '../../symbols';
 import { createCapturedArgs } from '../../vm/arguments';
-import { CheckArguments, CheckOperations, CheckReference } from './-debug-strip';
 import { Assert } from './vm';
 
 APPEND_OPCODES.add(Op.Text, (vm, { op1: text }) => {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Some preliminary improvement here (development mode): 
image


(at the cost of type safety)

NullVoxPopuli:
In production mode, it looks like we get back to 5.6 performance

image

chancancode:
> (at the cost of type safety)

It seems like these were intended to be stripped/strippable in prod/turned into no-op, any idea why that isn’t happening? (or maybe the stripping was just never implemented?)

NullVoxPopuli:
> It seems like these were intended to be stripped/strippable in prod/turned into no-op, any idea why that isn’t happening? (or maybe the stripping was just never implemented?)

I think the idea was that it would be dependent on the user's terser config -- and that is not reliable.

There is a prod build in the glimmer packages, but it seems ember projects don't currently use it (this is something we could try to fix, but I think it'd be better to try to solve with a merged monorepo (glimmer-vm in to ember-source)), so we had less layers of build configs.
I also haven't been able to tell if the checks are removed in glimmer's prod build, as everything is minified / obfuscated. 

At the same time, this change alone is not enough to regain lost perf, so there is more exploration to be had.

chancancode:
> > It seems like these were intended to be stripped/strippable in prod/turned into no-op, any idea why that isn’t happening? (or maybe the stripping was just never implemented?)
> 
> I think the idea was that it would be dependent on the user's terser config -- and that is not reliable.

I'm not sure that is a real issue for a few reasons.

For one thing, Ember itself relies on the same kinds of transformations for its own assertions, etc, and if all of that is generally not working correctly, then we have bigger problems to worry about.

Also, if the transformation is active and working properly, it should ultimately result in something along the lines of:

```
import { someDebugStuff } from "@glimmer/debug/hopefully-stripped-in-prod";

// ...

if (false) {
  someDebugStuff(maybeExpensiveArgument());
}
```

_That_ is the part that we would hope terser to remove – delete the branch altogether, then realize the import is unused and remove it (or maybe our transformations have already inlined the debug code and removed the import).

Even if _that_ is not working properly due to terser config, the worst that could happen is bloating the bundle size a bit, which is not ideal but shouldn't have the kind of effect we are seeing. At runtime, the engines won't ever run the actual expensive checking code, evaluate its arguments, etc. So if that's all there is to it, I would expect the worse that could happen is a slightly inflated boot time but gets amortized away in the repeated runtime benchmarks.

So I would think that to whatever extent that debug code gets into the final Ember build at all, and, worse _running_ at runtime, is indicative of a build configuration issue on our (or Ember's end), and #1599 seems to support that. That appears to account for most of the regression and lines up with the timeline (build refactors here, build refactors on Ember side, etc)

NullVoxPopuli:
> https://github.com/glimmerjs/glimmer-vm/pull/1602

it's perhaps not just terser, or we need more inlining or something.

Or maybe it's possible that the way the checks are written -- they just aren't strippable.

They don't have any production swaps like assert does so I don't know how terser would remove them

chancancode:
Yea it seems pretty clear to me the intent is to replace those `check(expr, xyz)` into just `expr` (then hopefully something else will realize xyz is not otherwise used and drop the import), but something we own has to be doing that rewrite. 

Either there is a misconfiguration and that rewrite stopped happening, or @wycats wrote those checks intending to later write the transform but never did. If it’s the latter, assuming we are already doing these sort of rewrites for some other cases, it shouldn’t be too bad to add this case too.

The checks are doing a useful thing (both for typescript during development, and to catch otherwise difficult to catch bugs in dev mode/when running tests), just have to be dropped somehow in production

PR #1628, opened 2024-10-10
Update puppeteer, node (for perf testing), and the setup script which launches tracerbench

NullVoxPopuli:


patricklx:
@NullVoxPopuli maybe try upgrading node version in perf.yml to `20.18.0`

.github/workflows/perf.yml
```diff
@@ -39,7 +39,12 @@ jobs:
       - uses: wyvox/action-setup-pnpm@v3
         if: steps.did-change.outputs.changed == 'true'
         with:
-          node-version: '20.1.0'
+          node-version: '20.18.0'
+
+      - uses: browser-actions/setup-chrome@v1
+        with:
+          chrome-version: 128
```
patricklx, line 46: I think puppeteer manages it's own chrome version.
You might need to downgrade puppeteer instead.
It had an update 4 days ago. Maybe its related

patricklx:
:( maybe try specific pupeteer version 20.0.0

patricklx:
Mmm, out of ideas..., looking at the logs it's still using latest chrome?

patricklx:
Mmm, out of ideas..., looking at the logs it's still using latest chrome?

.github/workflows/perf.yml
patricklx: I now saw that there is also puppeteer-chromium-resolver. According to docs it should be able to detect local chrome. But i think chrome must be installed before doing pnpm install.

patricklx:
It has options for chrome version selection https://www.npmjs.com/package/puppeteer-chromium-resolver#option-from-root-packagejson-with-pcr-object

patricklx:
I found a workaround. I noticed that the error is around GC multi thread handling. not sure why it happens
So, I found the nodejs flag to make it single thread:
in bin/setup-bench.mjs, line 177 replace command with
`node --single-threaded-gc ./node_modules/tracerbench/bin/run`

bin/setup-bench.mjs
```diff
@@ -174,7 +174,7 @@ await new Promise((resolve) => {
 
 try {
   const output =
-    await $`./node_modules/.bin/tracerbench compare --regressionThreshold 25 --sampleTimeout 60 --fidelity ${fidelity} --markers ${markers} --controlURL ${CONTROL_URL} --experimentURL ${EXPERIMENT_URL} --report --headless --cpuThrottleRate ${throttleRate}`;
+    await $`node --single-threaded-gc ./node_modules/.bin/tracerbench compare --regressionThreshold 25 --sampleTimeout 60 --fidelity ${fidelity} --markers ${markers} --controlURL ${CONTROL_URL} --experimentURL ${EXPERIMENT_URL} --report --headless --cpuThrottleRate ${throttleRate}`;
```
patricklx, line 177: ./node_modules/tracerbench/bin/run

packages/@glimmer/runtime/index.ts
```diff
@@ -1,4 +1,5 @@
 /**
+ * big change
```
patricklx, line 2: Don't forget to remove your big changes :)

PR #1631, opened 2024-10-15
add debug alias for let

patricklx:
This helps especially on cases for e.g.
When one has a let var which collides with a html element 

```hbs 
{{#let x as |b|}}
  {{b}}
  <b\>
{{/let}}
```


NullVoxPopuli:
Any ideas an the perf check failing? (It's failing on all branches)

We can't really merge anything until that's fixed :(

NullVoxPopuli (approved):
Seems good 
Screenshot_20241019-075938

patricklx:
@wycats i saw you are working on this kind of stuff. Maybe this us already fixed?

patricklx:
Btw, this is not derived debug alias, but the actual debug symbol what the user typed

Issue #1629, opened 2024-10-14
Fix major issue affecting ember-source 5.6+ with false-negatives w/ backtracking re-render assertions


NullVoxPopuli:

evoactivity:
Reproduction repo https://github.com/evoactivity/backtracking-repro/blob/main/app/components/thing/index.js

krisselden:
@NullVoxPopuli this is as designed, you are reading and writing to a tracked property in one render pass.  This makes it so that there are potentially 2 versions of the same tracked property.  There isn't any way for it to know you are not using the prior version you read.

There are also simple refactors to avoid it.   Most of this is because you want a tracked property for local changes based on derived state. You can create a new instance with a tracked property and cache it to the source tracked. Like tracked url to derived class State tracked isLoading = true, cached on url ==.   This lets you set isLoading to false when it finishes and if the url changes, you cache invalidates and you get a new State for the new url.

krisselden:
This issue comes up because the old assertion was tied to a chain of state tied all the way into the rendering.  As things move to tracked, rendering is basically cached on the source versions of tracked and doesn't book keep all the way to rendering. so the more systems that refactor to rely on autotracking the more you hit cases where people are reading during render and even if they aren't using that read value in any derived state, autotracking can't assume that.

NullVoxPopuli:
I think the reproductions I have are _not_ reproductions, because they fail on 0.84.3 as well.

boris-petrov:
I'm not sure I understand correctly what you're talking about here but could [this](https://github.com/emberjs/ember.js/issues/20657) also be what you're looking for as a reproduction?

wycats:
> the more systems that refactor to rely on autotracking the more you hit cases where people are reading during render and even if they aren't using that read value in any derived state, autotracking can't assume that

@krisselden A lot of the time, we're talking about patterns where people are reading values as part of a larger process that doesn't actually *depend* on the value (e.g. lazy initialization, memoization).

## An Aside About Autotracking Safety (and `untracked`)

In these cases, the value is guaranteed to be read exactly once per render transaction, and always for the purpose of producing the same answer for all calls during that render transaction. Many common use-cases for this kind of pattern are captured in [tracked-toolbox](https://github.com/tracked-tools/tracked-toolbox), although the code in `tracked-toolbox` is subtle enough that it's often hard to tell what's going on by reading the code.

[...]

PR #1632, opened 2024-10-15
Backtracking re-render during lazy initialization

NullVoxPopuli:
Issue:
- https://github.com/glimmerjs/glimmer-vm/issues/1629


NullVoxPopuli:
This isn't a reproduction as errors manifest on older VM versions

PR #1633, opened 2024-10-15
Update dependencies and pnpm version

wycats:
This commit updates a number of dependencies used during package development and as part of the build process.

In addition to updating dependencies, this commit also updates pnpm to v9.12, which adds the `catalog:` protocol[^1] (it was added in v9.5).

The `catalog:` protocol makes it possible to define package versions in `pnpm-workspace.yaml` and use those versions throughout the workspace.

This commit moved to `catalog:` for:

- eslint
- typescript
- vite

These dependencies are used extensively throughout the workspace, and `catalog:` is an improvement over the implicit `*` dependency we were using for TypeScript.

We could make more use of `catalog:` in the future, but this commit focuses on a few dependencies that were updated in multiple places in the workspace as part of the dependency updates in this commit.

[^1]: https://pnpm.io/catalogs


pnpm-workspace.yaml
```diff
@@ -9,3 +9,7 @@ packages:
   - 'packages/@glimmer-workspace/*'
   - 'packages/@glimmer-workspace/*/test'
   - 'packages/@types/*'
+catalog:
+  eslint: ^8.57.1
+  typescript: ^5.6.3
```
NullVoxPopuli, line 14: I think this has to be on ~5.5 -- ts-eslint doesn't even support 5.6 yet

kategengler:
We have merged this repo with https://github.com/emberjs/ember.js.

PR #1634, opened 2024-10-16
Plumb debug symbols when using lexical scope

wycats:
This change introduces a new `emit.debugSymbols` option that can be used to enable debug symbols when using lexical scope.

### DIFF
--- a/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
+++ b/packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
@@ -18,12 +18,15 @@ import type { EmberishCurlyComponent } from '..';
 import {
   BaseEnv,
   createTemplate,
+  defComponent,
+  defineSimpleModifier,
   GlimmerishComponent,
   JitRenderDelegate,
   RenderTest,
   suite,
   test,
   tracked,
+  trackedObj,
 } from '..';
 
 interface CapturedBounds {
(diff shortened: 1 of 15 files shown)


NullVoxPopuli (changes requested):
Some lint fixes are needed, also this PR demonstrates a performance regression

image

NullVoxPopuli:
Now the bench says things are faster... 

image

packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
```diff
@@ -74,6 +77,88 @@ class DebugRenderTreeTest extends RenderTest {
...
+        instance: null,
+        template: 'root.hbs',
+        bounds: this.elementBounds(this.delegate.getInitialElement()),
+        children: [
+          {
+            type: 'component',
+            name: 'HelloWorld',
+            args: { positional: [], named: { arg: 'first' } },
+            instance: null,
+            template: '(unknown template module)',
```
NullVoxPopuli, line 105: would be good to drop this, I think.
it's not helpful to have `(unknown template module)` for every strict-mode component -- they .. have no template?, yea? they aren't separate modules anymore.

NullVoxPopuli (approved):

packages/@glimmer-workspace/integration-tests/test/debug-render-tree-test.ts
wycats: I plan to do some follow-up PRs that finish the job of making the render tree reflect what the user typed. For now, I tried to leave things alone if they weren't directly on point with the specific changes here.

Specifically regarding `template`, I think we should consider whether we can replace it with module information. There's more to say about that, and I'd want to check with @ef4 about (a) whether it's realistic, and (b) what the right information to include would be.

In the short-term, though, the `template` field is increasingly likely to say `(unknown template module)` (and that's reliably true in the GJS format). We therefore never want to display it to users as-is and I agree it should be removed in a follow-up PR.

PR #1640, opened 2024-10-29
Run perf tests if a PR has the 'perf' label

NullVoxPopuli:
It's pretty easy to run the perf tests locally, and we can have a result in < 6 minutes on lower-end hardware.

On CI, it takes ~ 1 hour.


NullVoxPopuli:
image

PR #1643, opened 2024-10-30
Streamline append VM implementation and remove enums

wycats:
This PR begins the process of incrementally implementing many of the improvements in the error recovery branch.

I'm going to start with the easy stuff and work my way up to more complicated parts of the implementation, and in that spirit, this PR has two changes:

1. Streamline the AppendVM by using private fields and an optional `debug` field pattern. This simplifies the implementation and also makes it possible to consolidate `InternalAppendVM` into the normal `VM`. The benefit of the "opcode-only" subset of the VM was waning anyway, and most of the remaining distinction had to do with private implementation details that are no longer present in the `VM` interface because they're now implemented using private fields.
2. Replace `const enum` and `enum` with normal constants and literal types. There was a good reason for us to use `const enum` when we originally made the decision. But the dynamics have changed. Between (a) the fact that we can't reliably that `const enum` will be compiled away anymore, and (b) the fact that TS now has literal types, using normal constants and literal types is a better choice.


packages/@glimmer/compiler/index.ts
```diff
@@ -7,7 +7,8 @@ export {
   s,
   unicode,
 } from './lib/builder/builder';
-export { Builder, type BuilderStatement } from './lib/builder/builder-interface';
+export { type BuilderStatement } from './lib/builder/builder-interface';
+export * from './lib/builder/constants';
```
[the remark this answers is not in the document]
NullVoxPopuli: there is a lot in here: https://github.com/glimmerjs/glimmer-vm/pull/1643/files#diff-f881a95cdbb07c9ee63e2f6fca320f7d00323b27434dbee753f4cffb38d30a85R6

does it all need to be public API?

packages/@glimmer/compiler/lib/builder/constants.ts
NullVoxPopuli, line 1: is this file generated?

NullVoxPopuli (approved):
I'm surprised how much getting rid of enums helps with general readability

NullVoxPopuli:
Ope -- this is a regression 
image

NullVoxPopuli (changes requested):
We have a significant perf regression for some reason.

packages/@glimmer/compiler/index.ts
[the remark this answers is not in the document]
wycats: It was all public API but sucked into the `Builder` enum. And the codebase already does a lot of `export *`.

packages/@glimmer/compiler/lib/builder/constants.ts
wycats: Negative.

NullVoxPopuli: it's not too bad

wycats:
Superseded by:

- #1645 
- #1646 
- #1647 
- #1648

PR #1645, opened 2024-10-31
Move from const enums to consts

wycats:
This PR ended up being much more extensive than originally intended because it ended up including quite a bit of cleanup to the benchmark setup and cleaned up mistakes in the `package.json`s that were causing subtle issues in the benchmark environment.

[EDIT] @NullVoxPopuli pushed me to extract a lot of the incidental changes into their own PRs. I did quite a bit of that and it did streamline this PR. Thanks @NullVoxPopuli!


packages/@glimmer-workspace/integration-tests/lib/setup-harness.ts
NullVoxPopuli, line 1: Probably should have been a separate PR

bin/setup-bench.mjs
NullVoxPopuli, line 1: Probably should have been a separate PR

packages/@glimmer/compiler/index.ts
```diff
@@ -7,8 +7,9 @@ export {
   s,
   unicode,
 } from './lib/builder/builder';
-export { Builder, type BuilderStatement } from './lib/builder/builder-interface';
+export { type BuilderStatement } from './lib/builder/builder-interface';
 export { defaultId, precompile, precompileJSON, type PrecompileOptions } from './lib/compiler';
+export * from '@glimmer/constants';
```
NullVoxPopuli, line 12: How many things are in here?
Why *?

(This is an expansion of public api for glimmer compiler, so i just want to make sure this makes sense, rather than mostly a convenience (because idk))

packages/@glimmer/compiler/lib/passes/2-encoding/index.ts
NullVoxPopuli, line 1: These logging changes probably should have been a separate PR

packages/@glimmer/compiler/package.json
NullVoxPopuli, line 1: Probably should have been a separate PR?

packages/@glimmer/compiler/test/compiler-test.ts
```diff
@@ -50,8 +51,8 @@ function test(desc: string, template: string, ...expectedStatements: BuilderStat
   });
 }
 
-const Append = Builder.Append;
-const Concat = Builder.Concat;
+const Append = BUILDER_APPEND;
```
NullVoxPopuli, line 54: Why alias these?

It's how it was, and unaliasing them should probably be a separate PR

packages/@glimmer/debug-util/package.json
NullVoxPopuli, line 1: Probably should have been a separate PR

packages/@glimmer/local-debug-flags/index.ts
```diff
@@ -1,23 +1,171 @@
...
-      return false;
+/* eslint-disable no-console */
+
+declare global {
+  interface ImportMetaEnv {
+    VM_LOCAL_DEV?: boolean;
+  }
+}
+
+// All of these flags are expected to become constant `false` in production builds.
```
NullVoxPopuli, line 9: And those that exist through to ember (via debug mode, which is always used) should be strippable through ember's build process.
And those that make it through that (hopefully just DEBUG at that point) should be strippable by userland terser

packages/@glimmer/local-debug-flags/index.ts
NullVoxPopuli, line 1: Probably should have been a separate PR 🙈

packages/@glimmer/runtime/lib/opcodes.ts
```diff
@@ -190,4 +88,113 @@ export class AppendOpcodes {
   }
 }
 
+if (import.meta.env.VM_LOCAL_DEV) {
```
NullVoxPopuli, line 91: Nice!

NullVoxPopuli (approved):
Lots of changes! But overall looks good, and all the types of changes here are straight forward.

Good improvement, yay!

Left a question tho
(Doesn't affect approval)

packages/@glimmer/compiler/index.ts
wycats: This is a good point. There are a few things that were re-exported before when they were a single enum *type* that I carried forward and shouldn't have.

packages/@glimmer/compiler/test/compiler-test.ts
wycats: These are entirely internal aliases in this test file. It makes sense to unalias them, absolutely.

packages/@glimmer/local-debug-flags/index.ts
wycats: I addressed all of the other extraction requests. This one is sensible, but it would be a bunch of fine-grained work to pull it out and I'd rather just move ahead with this now that I pulled #1646, #1648 and 57d9118 into their own granular changes.

wycats:
image

Local results are "no change". I would have expected at least a small change, but the main motivation for this change was to move away from enums and const enums in favor of normal constants and literal types, which eliminates a dependency on increasingly non-standard compilation infrastructure.

So no news is good news.

NullVoxPopuli (approved):

NullVoxPopuli:
No perf difference on either of our local tests -- CI thinks improvement, so we'll merge.

PR #1649, opened 2024-11-03
Continue on with error-recovery infra refresh

wycats:
This PR focuses on bringing in the improvements to local logging and debug infrastructure.

The code is a little bit elaborate, but it does a good job of making the debug trace logs for VM operations clear and comprehensible.

Here are some examples of improved output:

image

image

Notable:

- The overall output is significantly clearer, with a decent amount of effort placed into turning internal representations into representations that would be useful to people working on the VM.
- Objects are represented *inline* as footnotes, and then logged as normal objects on a following line
- Format specifiers (`%d`, `%o`, etc.) are used for inline values with a built-in inline representation
- The trace logs now report *changes* to the VM state, rather than the entire VM state each time. 
- Array changes are represented in terms of `...x items` followed by a representation of push and pop operations.

In addition, it's possible to enable "subtle" mode, which shows *all* values rather than just changes.

image

Unchanged values are generally greyed out, so turning on subtle mode can be useful to get the full VM state for a change in normal trace logging mode, while still retaining the ability to identify changes.

Internally, there are two parts of this change:

1. An internal refactor to the debug state used to emit the opcode logs so that they're more consistent and more consistently snapshottable.
2. An internal, trace-only API for creating loggable fragments with support for subtle variations. These fragments can be combined naturally using template strings and combinators (e.g. `join` to join a list of fragments with a separator). When combined, the new fragment has natural subtle behavior (e.g. a list made entirely of subtle values `join`ed with `, ` is subtle).


wycats:
There are still a number of places that could benefit from further improvements to their trace log representation, but this PR creates a strong foundation to build on and is, itself a significant improvement to the status quo.

I ran the benchmarks locally and this PR didn't affect performance. This is to be expected since virtually all of the changes happened inside of `LOCAL_DEBUG` checks.

NullVoxPopuli:
perf test shows no difference 🎉

packages/@glimmer/debug/lib/dism/opcode.ts
```diff
@@ -0,0 +1,306 @@
...
+  if (nextSibling) {
+    return frag`${as.type('cursor')} { ${as.kw('before')} ${Fragment.special(nextSibling)} }`;
+  } else {
+    return frag`${as.type('cursor')} { ${as.kw('append to')} ${Fragment.special(element)} }`;
+  }
+}
+
+function describeBlock(
+  block: AppendingBlock,
+  type: 'block:simple' | 'block:remote' | 'block:resettable'
```
NullVoxPopuli, line 251: simple blocks were [renamed to appending blocks](https://github.com/glimmerjs/glimmer-vm/pull/1663/files#diff-d81a95325f353a0a60d137f71a5a57f6f4ef3da2a4dad75069c65744035ade47R99), yeah?

what determines simple vs remote vs resettable?

packages/@glimmer/debug/lib/render/annotations.ts
NullVoxPopuli, line 1: yay oklch!

NullVoxPopuli (approved):
Do we need / want to inline `@glimmer/debug-util` so that it's stripped during the build?

wycats:
> Do we need / want to inline `@glimmer/debug-util` so that it's stripped during the build?

Yep, [we do that](https://github.com/glimmerjs/glimmer-vm/blob/35817f76252fcc37703a2ce4a7dd0db152c4c886/packages/%40glimmer-workspace/build/lib/config.js#L115-L125)

packages/@glimmer/debug/lib/dism/opcode.ts
wycats: - Appending: A block that will never be cleared. However, since it's possible to append a block to an Appending Block, the boundaries of an Appending Block can change. See `FirstNode` / `LastNode` for details on how this works.
- Resettable: A block that can be reset (i.e. cleared). When a block is reset, its contents are cleared and it produces a new cursor to render a block into. This is used for constructs like `{{#if}}` / `{{else}}`, which clear and rerender the block whenever the boolean value changes.
- Remote: A block that represents a remote element. When the block is destroyed, the nodes rendered into the remote element are removed.

PR #1650, opened 2024-11-06
allow uppercase elements in strict mode

patricklx:
this would allow to have web components with upper case style.

right below the code it actually says `In strict mode, values are always elements unless they are in scope`
so lets make it that way

### DIFF
--- a/packages/@glimmer/syntax/lib/v2/normalize.ts
+++ b/packages/@glimmer/syntax/lib/v2/normalize.ts
@@ -762,13 +762,6 @@ class ElementNormalizer {
     let inScope = variable[0] === '@' || variable === 'this' || this.ctx.hasBinding(variable);
 
     if (this.ctx.strict && !inScope) {
-      if (uppercase) {
-        throw generateSyntaxError(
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (approved):

NullVoxPopuli:
Have to revert, apologies. This accidentally blocks `<Style>` from becoming a keyword -- something we are entertaining the idea of here: https://github.com/emberjs/rfcs/pull/1037

patricklx:
Just a web component.
```gjs
<template>
  Hello, Glimmer!
  <MyWebComponent />
</template>
```
https://share.glimdown.com/jnNZc45SRtWdT93u7GXt5Q

This also affects https://github.com/ember-native/ember-native
Where all nativescript components are dom elements.
By default nativescript writes them with upper case. I wanted to have it work the same .
And it strict mode everything that is not in scope could just be rendered as html element.
Why should `randomElementName` work, but not `RandomElementName`? In strict mode. gjs.

I'm not sure how this would block the Style keyword.
It would need code changes in glimmer anyway? A element keyword looks like a new concept. Could it not be something importable? (Well, maybe this part is for the other discussion)
If it is uppercase then the usage of it is currently blocked :D, if it doesn't have an import.

PR #1657, opened 2024-11-13
Failing test for inline-if component return

NullVoxPopuli:
https://github.com/glimmerjs/glimmer-vm/issues/1642


NullVoxPopuli:
oh no 
image

PR #1663, opened 2024-11-14
Streamline and modernize naming

wycats:
This PR pulls out the naming and documentation changes in #1649 so they can be landed separately.

I just finished resolving all of the merge conflicts, so I expect tests, types and lints to fail at the moment. 

It may still be appropriate to break this PR up further. Once tests are passing, we can evaluate that question.

[NOTE] This PR builds on #1664, and will be a bit smaller once that lands.


mise.toml
```diff
@@ -0,0 +1,3 @@
+[tools]
```
NullVoxPopuli, line 1: isn't the point of these tools (mise, proto, etc) to not need tool-specific configs in the repo?

the packageManager field in the root package.json should be sufficient, yeah?
and then an engines entry?

wycats: > isn't the point of these tools (mise, proto, etc) to not need tool-specific configs in the repo?

Whether or not this is true, it's an accident that this got pushed.

packages/@glimmer/interfaces/lib/dom/attributes.d.ts
```diff
@@ -12,23 +12,30 @@ import type {
...
-export interface SimpleLiveBlock extends LiveBlock {
-  parentElement(): SimpleElement;
-  firstNode(): SimpleNode;
-  lastNode(): SimpleNode;
-}
-
-export type RemoteLiveBlock = SimpleLiveBlock;
-
-export interface UpdatableBlock extends SimpleLiveBlock {
+/**
```
NullVoxPopuli, line 23: I like these much better than their previous names

packages/@glimmer/interfaces/lib/program.d.ts
```diff
@@ -44,36 +41,47 @@ export interface CompileTimeHeap extends OpcodeHeap {
...
+   * The compiled program itself: the constants and heap
+   */
+  readonly program: Program;
+  /**
+   * The offsets to stdlib functions
+   */
   readonly stdlib: STDLib;
-
+  /**
+   * A framework-specified resolver for resolving free variables in classic templates.
```
NullVoxPopuli, line 65: I super appreciate all these inline comments!

NullVoxPopuli (approved):
lgtm!

PR #1664, opened 2024-11-14
Move curried types to constants

wycats:
Curried types were still not using constants after the previous cleanups in e18b5b6176175ea509cc42b87c94eef4c11a6e50 and cdf810ee9b3df089c7614c936ec16a1fb18ae424, and this commit finishes the job.


NullVoxPopuli (approved):

NullVoxPopuli:
1st local run show no difference across all tests (except for clearItems4End shows a tiny improvement - and render1000Items2end shows minor regression -- both improvement and regression I think would be in the margin of error, given the range.

2nd local run shows no difference for all measurements (I made more sure to not use my laptop during this time).

so.. I think we're safe!

PR #1665, opened 2024-11-19
Feature/pass args to debugger

wycats:
This commit eliminates debugger infrastructure that derived from the old partial/eval infrastructure.

Instead, it passes the in-scope values as a constant to the `debugger` wire format opcode and uses it directly in the scope inspector.

This will be followed up with support for accessing strict-mode lexical variables.


NullVoxPopuli:
local perf test shows no difference compared to main 🎉

PR #1666, opened 2024-11-20
Tolerant parsing: continue parsing when a Glimmer syntax is invalid

BlueCutOfficial:
This PR is far from ready, it simply aims at starting a discussion. 

## Context

When using `ember-template-lint`, any syntax error found throws a `GlimmerSyntaxError` and makes the result of the lint process invalid. On a project I've been working on with @mansona,  we would have needed some sort of "tolerant mode", a way to be notified about errors without blocking the linter.

The current state of this PR is just a start, and aims at gathering feedbacks about the direction it should take: 
- It takes the `PartialStatement` as an example and doesn't throw when a Handlebars partial is found, and the resulting AST template simply doesn't contain the node.
- The next step would be to build some kind of error node that would have a representation in the resulting AST template.
- Using a new class field for the `Parser` seemed cleaner and less intrusive to me than trying to implement it as parse function's argument.

### DIFF
--- a/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
+++ b/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
@@ -105,6 +105,15 @@ export abstract class HandlebarsNodeVisitors extends Parser {
     return node;
   }
 
+  triggerError(message: string, loc: SourceSpan): ASTv1.ErrorNode {
+    if(this.continueOnError) {
+      let error = b.error({ loc,  message });
+      appendChild(this.currentElement(), error);
+      return error;
+    }
+    throw generateSyntaxError(message, loc);
+  }
+
   BlockStatement(block: HBS.BlockStatement): ASTv1.BlockStatement | void {
     if (this.tokenizer.state === 'comment') {
       this.appendToCommentData(this.sourceForNode(block));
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
> feedbacks about the direction it should take:
> 1. It takes the PartialStatement as an example and doesn't throw when a Handlebars partial is >   found, and the resulting AST template simply doesn't contain the node.
> 1. The next step would be to build some kind of error node that would have a representation in >   the resulting AST template.
> 1. Using a new class field for the Parser seemed cleaner and less intrusive to me than trying to implement it as parse function's argument.

I think we should only focus on #2 here. As a parser, we don't really want to have varying APIs as we churn through version numbers (as we normally would in post-1.0 packages). This is slightly constrained by the VM still being pre 1.0, but also, fault tolerance is something all our parsing tools want:
- Glint
- Eslint
- Prettier

As when we can correctly parse and recover from errors we can provide more helpful information to the user to aid them in fixing their syntax errors.

For example, here is what tree-sitter does (error nodes):

https://github.com/user-attachments/assets/bfee79ae-d479-4c74-9cff-ceca15ac8d38

Even within an error node, it tries to recover immediately, which also keeps syntax highlighting nice, but this also has the benefit of allowing in-progress intellisense, such as this more realistic example: 
image

BlueCutOfficial:
> I think we should only focus on `#2` here.

I started to work on the Error node thing. I hope I can post one more commit tomorrow, and you'll  tell me if I am doing this right or wrong. Thanks for the first answer and recordings 👍

BlueCutOfficial:
Here is the result template I have so far for the "continue parsing after an error" test: 
Screenshot 2024-11-21 at 10 14 40

Please let me know your thought about the code and what you think is missing to get this PR ready (I am a bit in discovery mode here, I don't have personal opinions to defend).

NullVoxPopuli:
yea, for partials, that's that a very nice output!

BlueCutOfficial:
If I have this piece of code: `let t = <div {{mymodifier}}></div>`

<details>

<summary>Parsing t results to the following structure ⬇️ </summary>

```javascript
{
      "attributes": [],
      "blockParams": [],
      "children": [],
      "comments": [],
      "modifiers": [
        {
          "hash": {
            "pairs": [],
            "type": "Hash"
          },
          "params": [],
          "path": {
            "head": {
              "name": "mymodifier",
              "original": "mymodifier",
              "type": "VarHead"
            },
            "original": "mymodifier",
            "tail": [],
            "type": "PathExpression"
          },
          "type": "ElementModifierStatement"
        }
      ],
      "params": [],
      "path": {
        "head": {
          "name": "div",
          "original": "div",
          "type": "VarHead"
        },
        "original": "div",
        "tail": [],
        "type": "PathExpression"
      },
      "selfClosing": false,
      "tag": "div",
      "type": "ElementNode"
    }
``` 

</details>

Now if I have something invalid: `let t = <div {{...attributes}}></div>`.

Where do you think the error node should go in the structure? Would it belong to 1. the list of modifiers, or 2. a list `"errors"` to be added to `ElementNode` fields.

reteps:
Hi there -- I'm thinking this PR would be a first step for prettier to support some handlebars variants. Ideally, I would love if when Glimmer fails to parse the AST of a node, it can expose enough information that the the rest of the document can be parsed and formatted. For example if `<div {{mymodifier}}></div>` fails, information like an `ErrorNode="<div {{mymodifier}}></div>"`.

There are currently a couple prettier issues with handlebars templates, so making the parser more tolerant would support those cases. The ones I am aware of are: 

https://github.com/prettier/prettier/issues/14834
https://github.com/prettier/prettier/issues/14090
https://github.com/prettier/prettier/issues/11834

NullVoxPopuli:
aye, we have some of that work happening over here: https://github.com/glimmerjs/glimmer-vm/pull/1690

which has lead to vastly improved error reporting :tada:

PR #1673, opened 2025-01-09
Lexical "this" capture

ef4:
Opening for discussion on whether this is a good direction to solve https://github.com/emberjs/babel-plugin-ember-template-compilation/issues/61

You can see the other half of the changes in:

https://github.com/emberjs/babel-plugin-ember-template-compilation/pull/67

which patches this PR's changes into ember-template-compiler.js.

The idea here is that `@glimmer/syntax` continues to just trust the outer layer (babel-plugin-ember-template-compilation) to manage all aspects of JS scope, including whether or not `this` should be considered to in local scope. If it's in scope, it compiles to a normal variable reference.


ef4:
Was discussed and approved during last week's spec meeting.

PR #1677, opened 2025-01-13
Refresh tsconfig and eslint setup

wycats:
This is a significant overhaul of the configuration setup for the repository, driven largely by the work required to update the repository to flat config and ESLint 9.

---

1. Migrate to ESLint 9 + new config format
2. Migrate to a single tsconfig file for the entire repo

In general, this results in more consistency across the repo, including in one-off packages and tests, but it still gives us flexibility to create special rules if necessary.

As part of this process, a new `repo-metadata` package was created that allows other infrastructure packages to get workspace metadata. This package has a static JSON file with the information about each package in the workspace. This was important in the current PR because it allows our eslint config to be sensitive to information in individual packages
without slowing down the boot time of eslint.

Fully migrating to ESLint 9 required:

- Making the whole repo clean with `eslint .`
- Making `eslint .` in individual packages work (and perhaps move the custom configuration to configs in each package
- Moving the guts of the code into `@glimmer-workspace/eslint-plugin`. This should be **much** more straightforward with ESLint 9.
- Porting all of the ESLint rules from the current setup to the new config, or deciding to ignore them.
- Updating our JSON linting

## Increased Benchmark Isolation and Predictability

Rather than using a workspace package for benchmarks and trying to get `vite build` to emulate the expected behavior, we `pack` the dependencies and refer to them directly in the `package.json`.

This is a little bit more involved than it might seem like it should be because the benchmark has to run on the `main` branch, so it has to be very general and can't use any infrastructure in the workspace itself that could affect the builds.

On the bright side, that constraint pushes us towards treating the benchmark environment as a normal consumer of the normal Glimmer packages, which is good.

The only tiny exception to that rule is that the build script builds the `@glimmer-workspace/benchmark-env` package, even though it's not a public package. This means that the code in `@glimmer-workspace/benchmark-env` needs to make sure it can be built.

The current setup allows us to use the `main` branch's `@glimmer-workspace/benchmark-env` package with the experiment's `krausest` package, so it seems reasonable to build the benchmark env and use it as a traditional package rather than trying to get two
workspaces to somehow mesh together (which is what got us into trouble in the first place).

> [!NOTE]
>
> Getting this working right now requires a little bit of hackery when setting up the current `control` (i.e. `main`) branch, because `main` isn't set up to build the `@glimmer-workspace/benchmark-env`.
>
> Once this PR is merged, we should be able to remove those hacks and rely on vanilla package builds for the entire setup going forward.

## Meta Updater and Repo Scripts

This PR introduces a [meta-updater](https://github.com/pnpm/meta-updater) setup that keeps packages aligned with repository conventions based upon whether they're published, need a build, are test packages, etc.

It further standardized per-package scripts via `turbo`, and created top-level `repo:*` scripts to differentiate repo workflows from standard scripts in each package. This PR also begins to document repo workflows (in `guides/workspace`).

## Inline Notes

This PR adds a `@bandaid` highlighting rule to capture situations where code exists as a temporary workaround. The rule and allows you to document the reason for the workaround so it can be identified and cleaned up later.

<kbd> image </kbs>

I also made more use of an existing `@fixme` annotation.

<kbd> image </kbd>

> I think we should improve our workflow around `fixme`s so that they're linked to issues and perhaps can expire based on certain conditions, but we don't need to block this PR on those improvements.
>
> The `@fixme`s added in this PR are almost entirely attached to `eslint-disable` annotations that didn't exist before because of how inconsistently our previous lint setup applied to specific files. They should be cleaned up as soon as possible, but they represent an improvement to the status quo (not a regression).

## Testing Entry Points

While developing this PR, the tests began to fail because Vite was failing to interpret the
`import.meta.glob`s in `index.html` file as entry points. This resulted in multiple copies of the validator code getting included in the tests (via chunks and via the `.ts` module itself).

To make matters worse, the validator code included in the chunk was missing the code that detects duplicates, so the only failure symptom was an error reporting that the `COMPUTE` symbol wasn't present on an instance of `Tag`.

Being explicit about entry points fixed the problem and made the tests pass. I don't know why Vite doesn't interpret `import.meta.glob` properly in this situation, but I think we can live with duplicating the test glob in `vite.config.mjs` for now.

## Other Improvements

- Document vscode workflow recommendations
- Explain recommended extensions for vscode
- Remove many unused dependencies


.vscode/extensions.json
```diff
@@ -1,3 +1,12 @@
 {
-  "recommendations": ["foxundermoon.shell-format"]
+  "recommendations": [
```
wycats, line 2: These recommendations are documented in guides/workspace/vscode.md

.vscode/settings.json
```diff
@@ -3,110 +3,37 @@
...
-        "fontWeight": "bold",
-        "color": "#ff0000"
-      }
+    "editor.rulers": [
+      100
+    ],
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "editor.formatOnSave": false
+  },
+  "[json][jsonc][markdown][yaml]": {
```
wycats, line 19: Same deal with JSON. Note that JSON and JSONC are linted via `eslint` (via `eslint-plugin-jsonc`) in
this repository, so `fixAll` is meaningful.

.vscode/settings.json
```diff
@@ -3,110 +3,37 @@
...
-  ],
-  "eslint.validate": ["javascript", "typescript", "json", "jsonc"],
-  "files.exclude": {
-    "**/.DS_Store": true,
-    "**/.git": true,
-    "**/node_modules": true,
-    "**/dist": true,
-    "tracerbench-results": true
-  },
-  "files.insertFinalNewline": true,
```
wycats, line 39: These already come from .editorconfig

.vscode/settings.json
```diff
@@ -115,120 +42,239 @@
...
+    ],
+    "fixme": [
+      "@fixme(\\s|$)"
+    ],
+    "warn": [
+      "@premerge(\\s|$)"
+    ]
+  },
+  "inline-bookmarks.view.showVisibleFilesOnly": true,
+  "javascript.preferences.importModuleSpecifier": "project-relative",
```
wycats, line 259: auto-import from other workspace packages directly (i.e. `from "@glimmer/vm"`) rather than importing
(incorrectly) from a relative path (which would escape the package boundary).

benchmark/benchmarks/krausest/lib/components/Application.d.hbs.ts
```diff
@@ -0,0 +1,4 @@
+import type { SerializedTemplateWithLazyBlock } from '@glimmer/interfaces';
```
wycats, line 1: There are two changes in all files in this directory:

1. Importing from `@glimmer/interfaces` rather than a relative path.
2. Using `.d.hbs.ts`, which is the new TypeScript default for `allowArbitraryExtensions` (see [the docs](https://www.typescriptlang.org/tsconfig/#allowArbitraryExtensions))

benchmark/benchmarks/krausest/lib/components/Application.ts
```diff
@@ -1,6 +1,9 @@
-import { swapRows, type Item, updateData, buildData } from '@/utils/data';
```
wycats, line 1: Throughout the benchmarks, use [subpath
imports](https://nodejs.org/api/packages.html#subpath-imports) (the new npm standard) rather than
path mappings.

benchmark/benchmarks/krausest/lib/utils/data.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/restrict-plus-operands */
```
wycats, line 1: The biggest change across the entire PR is the addition of `eslint-disable` directives to account
for the more consistent application of eslint rules in this PR.

benchmark/benchmarks/krausest/.eslintrc.cjs
```diff
@@ -1,13 +0,0 @@
-/** @type {import("eslint").Linter.Config} */
```
wycats, line 1: We don't need individual .eslintrc files anymore.

bin/build-verify.mjs
```diff
@@ -45,7 +46,9 @@ for (let filePath of files) {
   let content = file.toString();
 
   for (let searchFor of FORBIDDEN) {
-    if (content.match(searchFor)) {
+    const match = typeof searchFor === 'string' ? content === searchFor : searchFor.test(content);
```
wycats, line 49: Bug (?) caught by linting. Even if it somehow works, it seems good to make the code match the types.

bin/link-all.mts
```diff
@@ -1,12 +1,14 @@
-import { readFile, writeFile, unlink } from 'node:fs/promises';
+import { writeFileSync } from 'node:fs';
```
wycats, line 1: The PR has a lot of changes to imports to make them consistent.

bin/link-all.mts
```diff
@@ -55,7 +57,7 @@ const unpack = pkgs.map(async (pkg) => {
     const packageJsonPath = join(pkgDest, 'package.json');
     const packageJson = JSON.parse(await readFile(packageJsonPath, { encoding: 'utf8' }));
     delete packageJson.devDependencies;
-    await writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2), {
+    writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2), {
```
wycats, line 60: `writeFileSync` is not async, so no need for `await`. There are a number of cases of this in the PR.

bin/opcodes.mts
```diff
@@ -16,11 +16,11 @@ const ALL: (string | null)[] = [
 const TYPES = ['interface', 'code', 'debug', 'all'] as const;
 
 if (emitter.options.help) {
-  throw usage();
+  usage();
```
wycats, line 19: An old version of TS didn't properly handle functions returning `never`, and we worked around that
by throwing the `never`. But now that's fixed and it's a lint error to throw `never`, so there are a
number of places where this is fixed.

bin/package.json
```diff
@@ -1,11 +1,23 @@
 {
   "name": "@glimmer-workspace/bin",
+  "type": "module",
```
wycats, line 3: All packages have `"type": "module"` to reflect reality.

bin/package.json
```diff
@@ -1,11 +1,23 @@
...
   "private": true,
+  "repo-meta": {
```
wycats, line 5: This PR  introduces a way of specifying package-specific tweaks to the overall linting
configuration. The details are documented in @glimmer-workspace/eslint-plugin/README.md

bin/package.json
```diff
@@ -1,11 +1,23 @@
...
+      "node",
+      "console"
+    ],
+    "lint": [
+      "*"
+    ]
+  },
   "scripts": {
-    "test:lint": "eslint .",
+    "test:lint": "eslint . --quiet",
```
wycats, line 16: This PR uses `warning` level lints for situations where we want to annotate a line with a `disable`
as documentation for the future (when we enable stricter lints), but don't want the lint to fail
yet. It's a temporary measure, and we should switch back to error-only lints once we've moved
everything to stricter lints.

bin/run-browserstack-tests.mjs
```diff
@@ -36,13 +38,13 @@ function run(command, args = []) {
...
       console.log('success');
-      // eslint-disable-next-line n/no-process-exit
-      process.exit(0);
     } finally {
       if (process.env['GITHUB_RUN_ID']) {
         await run('ember', ['browserstack:results']);
       }
       await run('ember', ['browserstack:disconnect']);
+      // eslint-disable-next-line n/no-process-exit
+      process.exit(0);
```
wycats, line 47: The previous code ran process.exit() in the `try`, which would mean that the `finally` would never
run. I'm not sure why that worked (perhaps because we never relied on browserstack to run
properly?), but I moved the `exit` to the end of the `finally` rather than completely delete it,
since it presumably exists for a reason?

bin/run-tests.mjs
```diff
@@ -50,10 +51,10 @@ const browser = await puppeteer.launch({
 console.log('[ci] puppeteer launched');
 
 try {
-  await /** @type {Promise<void>} */ (
-    new Promise(async (fulfill, reject) => {
-      const page = await browser.newPage();
+  const page = await browser.newPage();
```
wycats, line 54: This code is adjusted to run the asynchronous part of the work outside of the `Promise` constructor.
While keeping it inside the `Promise` constructor *works* in this situation, it doesn't behave the
way it looks like it does.

bin/run-tests.mjs
```diff
@@ -70,13 +71,14 @@ try {
...
+          // eslint-disable-next-line @typescript-eslint/prefer-promise-reject-errors
           reject();
         }
       });
-
-      await page.goto('http://localhost:60173?hidepassed&ci');
     })
   );
+
+  await page.goto('http://localhost:60173?hidepassed&ci');
```
wycats, line 81: This is a change in behavior (the old code was fire-and-forget because of the confusing way it used
the `Promise` constructor). This is my best guess, and we need to verify that it still works
correctly.

In general, there's a bunch of "fire and forget" code in the `bin` directory, and we should document
when we do that, and why.

bin/setup-bench.mjs
```diff
@@ -1,7 +1,12 @@
-import 'zx/globals';
```
wycats, line 1: Why import globals when you can just import `$`?

bin/tsconfig.json
```diff
@@ -1,23 +0,0 @@
-{
```
wycats, line 1: No more project references.

packages/@glimmer/compiler/lib/builder/builder.ts
```diff
@@ -38,7 +38,7 @@ import {
   SPLAT_HEAD,
   THIS_VAR,
 } from '@glimmer/constants';
-import { assert, exhausted, expect, isPresentArray } from '@glimmer/debug-util';
+import { exhausted, expect, isPresentArray,localAssert } from '@glimmer/debug-util';
```
wycats, line 41: Renamed `assert` to `localAssert` to make it super-clear that it's not included in the build output.

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/utils/if-unless.ts
```diff
@@ -67,7 +67,7 @@ function assertIfUnlessInlineKeyword(type: string) {
             inverted ? 'false' : 'true'
           }, and 3. the value to return if the condition is ${
             inverted ? 'true' : 'false'
-          }. Received ${positional?.size ?? 0} parameters`,
+          }. Received ${positional.size} parameters`,
```
wycats, line 70: There's quite a bit of code in the repo that checked for things that can't happen. Sometimes, this
is because it's handling code from untyped JS users. In other cases, like this one, it's purely
internal code interacting with other typed code, and the check is a holdover from a previous time
when the check was necessary.

The current linting setup warns about situations like this. In some situations, like this one, I removed the unnecessary
check. In other situations, I disabled the lint and identified the fact that the check is for JS callers.

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/impl.ts
```diff
@@ -8,8 +8,8 @@ import type { NormalizationState } from '../context';
 import { Err } from '../../../shared/result';
 
 export interface KeywordDelegate<Match extends KeywordMatch, V, Out> {
-  assert(options: Match, state: NormalizationState): Result<V>;
-  translate(options: { node: Match; state: NormalizationState }, param: V): Result<Out>;
+  assert: (options: Match, state: NormalizationState) => Result<V>;
```
wycats, line 11: Since `assert` and `translate` are extracted, this documents that they are arrows without a `this` binding.

packages/@glimmer/compiler/lib/passes/1-normalization/keywords/impl.ts
```diff
@@ -115,20 +115,9 @@ export type KeywordNode =
   | ASTv2.InvokeBlock
   | ASTv2.ElementModifier;
 
-export function keyword<
```
wycats, line 118: This code is only used in one place and abstracts very little, and the bulk of the implementation is
type annotations. ✂️

packages/@glimmer/debug/lib/dism/opcode.ts
```diff
@@ -29,7 +29,7 @@ export function describeOp(
   program: Program,
   meta: Nullable<BlockMetadata>
 ): Fragment {
-  const { name, params } = debugOp(program, op, meta)!;
+  const { name, params } = debugOp(program, op, meta);
```
wycats, line 32: There's a decent amount of totally unnecessary `!` assertions, and they were removed via autofix.

packages/@glimmer/debug/lib/render/fragment.ts
```diff
@@ -30,31 +30,31 @@ export class Fragment<T extends FragmentType = FragmentType> {
   static integer(
     this: void,
     value: number,
-    options?: Omit<IntegerFragment, 'kind' | 'value'> | undefined
+    options?: Omit<IntegerFragment, 'kind' | 'value'>
```
wycats, line 33: You don't need `| undefined` in parameter positions, and this (and many other cases like it) were
caught by the new lint rules.

packages/@glimmer/debug/lib/render/logger.ts
```diff
@@ -53,12 +53,10 @@ export class DebugLogger {
   #lines(type: 'log' | 'debug' | 'group' | 'groupCollapsed', lines: FlushedLines): void {
     const [first, ...rest] = lines;
 
-    if (first) {
```
wycats, line 56: The whole point of `FlushedLines` is to require at least one lint, so this check is unnecessary. ✂️

packages/@glimmer/debug/lib/vm/snapshot.ts
```diff
@@ -172,6 +173,7 @@ export class VmSnapshotArrayDiff<N extends string | number, T extends unknown[]>
         post = value(after, { ref: `${i}:push` });
       } else if (op === 'retain') {
         post = value(after, { ref: `${i}:after` });
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- exhaustion check
```
wycats, line 176: There's possibly a better way to do this, but the issue is: the linter knows that there's only one
kind of `op` left (`'pop'`) so  it thinks that the check is unnecessary (and could be handled by a
simple `else`). However, it's here to ensure that if `op` gets a new member, the type will fail and
force us to add a new clause.

There are a few places like this, and it would be good to consolidate on a single good solution, but
I think ad-hoc solutions are fine for now, since there aren't a ton of these cases, and the "no
unnecessary check" rule is quite valuable and caught a lot of places where the extra check was
either a mistake or caused confusion.

packages/@glimmer/debug/lib/debug.ts
```diff
@@ -66,9 +70,8 @@ export function describeOpcode(type: string, params: Dict<SomeDisassembledOperan
   }
 }
 
-function stringify(value: number, type: 'constant'): string;
+function stringify(value: number, type: 'constant' | 'variable' | 'pc'): string;
```
wycats, line 73: I generally took the linter's advice and consolidated overloads when possible.

packages/@glimmer/debug/lib/debug.ts
```diff
@@ -183,7 +186,8 @@ export type OperandOptionsA<O extends AnyOperand> = O extends [
   options: infer Options,
 ]
   ? Options
-  : {};
+  : // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
wycats, line 189: I didn't always try to figure out what `{}` could be replaced with without creating more type
errors, but it's a good thing to follow-up with.

packages/@glimmer/debug/lib/debug.ts
```diff
@@ -354,7 +358,7 @@ export function debugOp(program: Program, op: OpSnapshot, meta: BlockMetadata |
   if (LOCAL_DEBUG) {
     let metadata = opcodeMetadata(op.type);
 
-    let out: Dict<RawDisassembledOperand> = Object.create(null);
+    let out = Object.create(null) as Dict<RawDisassembledOperand>;
```
wycats, line 361: This is a common change: there are a number of places where we were using `const x: Y = z` as a type
assertion, but the safety lint rules (correctly) treat that as an `any` assignment. While this is
just as unsafe, an `as` cast more idiomatically communicates an unsafe cast.

An `:` assignment *normally* represents a safe assignment, but it was only allowed *because* the
expression was `any`. I think this is a nice improvement to clarity.

packages/@glimmer/debug/lib/metadata.ts
```diff
@@ -54,11 +54,8 @@ export type RawOperandFormat = OperandName | PresentArray<OperandName>;
 export function normalize(key: string, input: RawOperandMetadata): NormalizedMetadata {
   let name: ShorthandOperand;
 
-  if (input.format === undefined) {
```
wycats, line 57: This is impossible, and it's internal, well-typed code.

packages/@glimmer/debug/lib/metadata.ts
```diff
@@ -54,11 +54,8 @@ export type RawOperandFormat = OperandName | PresentArray<OperandName>;
...
-    throw new Error(`Missing format in ${JSON.stringify(input)}`);
-  }
-
   if (Array.isArray(input.format)) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- @fixme
```
wycats, line 58: I left quite a few of these `!` assertions alone with the intent of going back and triaging them. I
still want to do a quick pass for cases where an alternative pattern would be better, but I
generally tried to break ties by leaving the behavior alone. I think it would be good to follow up
and try to be more precise about what's going on in these situations, but I think it's ok to do so
incrementally.

packages/@glimmer/debug/test/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
wycats, line 1: Some of the test packages were missing `package.json`s for some reason, which broke the assumptions
in the linting setup.

packages/@glimmer/debug-util/lib/simple-cast.ts
```diff
@@ -23,13 +23,14 @@ type BrowserTag = keyof BrowserTags;
...
 type NodeCheck<N extends Node> = (node: Node) => node is N;
 type SugaryNodeCheck<K extends BrowserTag = BrowserTag> = NodeCheck<BrowserTags[K]> | K | K[];
-type NodeForSugaryCheck<S extends SugaryNodeCheck<BrowserTag>> = S extends NodeCheck<infer N>
-  ? N
-  : S extends keyof BrowserTags
-    ? BrowserTags[S]
-    : S extends (keyof BrowserTags)[]
-      ? BrowserTags[S[number]]
-      : never;
+type NodeForSugaryCheck<S extends SugaryNodeCheck> =
```
wycats, line 26: This is one of a number of places where the default value for a type parameter was explicitly
included. They were removed by autofix.

packages/@glimmer/debug-util/test/debug-to-string-test.ts
```diff
@@ -55,6 +55,7 @@ if (import.meta.env.DEV) {
     );
   });
   QUnit.test('should return debug name for class', (assert) => {
+    // eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
wycats, line 58: I don't know how much we care about this lint, but I disabled it for now rather than make too many
changes to the default strict settings in `@typescript-eslint`.

packages/@glimmer/destroyable/index.ts
```diff
@@ -115,8 +115,9 @@ export function registerDestructor<T extends Destroyable>(
 
   let meta = getDestroyableMeta(destroyable);
 
-  let destructorsKey: 'eagerDestructors' | 'destructors' =
-    eager === true ? 'eagerDestructors' : 'destructors';
+  let destructorsKey: 'eagerDestructors' | 'destructors' = eager
```
wycats, line 118: `x === true` boolean checks are autofixed to `x` boolean checks if `x` is a boolean.

packages/@glimmer/interfaces/lib/compile/index.d.ts
```diff
@@ -1,4 +1,4 @@
-export * from './encoder.js';
-export * from './instruction-encoder.js';
-export * from './operands.js';
-export * from './wire-format/api.js';
+export type * from './encoder.js';
```
wycats, line 1: Even though this is a `.d.ts` file, it seemed fine to follow the linting rules and `export type`.
The whole `@glimmer/interfaces` story could probably use a clear plan and overhaul, but I think it's
fine to follow the normal TypeScript lints here unless there's a good reason not to.

packages/@glimmer/interfaces/lib/template.d.ts
```diff
@@ -120,7 +120,7 @@ export interface BlockMetadata {
   symbols: BlockSymbolNames;
   scopeValues: unknown[] | null;
   isStrictMode: boolean;
-  moduleName: string;
+  moduleName: string | undefined;
```
wycats, line 123: This is actually allowed to be `undefined`, and when the auto-fixer tried to assume otherwise, it
produced failing code.

packages/@glimmer/manager/lib/internal/api.ts
```diff
@@ -5,6 +5,7 @@ import type {
   Owner,
 } from '@glimmer/interfaces';
 import { debugToString } from '@glimmer/debug-util';
+import { debugAssert } from '@glimmer/global-context';
```
wycats, line 8: Lots of little clarifying tweaks in this file, including using asserts instead of errors.

packages/@glimmer/manager/lib/public/helper.ts
```diff
@@ -21,9 +22,12 @@ export function helperCapabilities<Version extends keyof HelperCapabilitiesVersi
   managerAPI: Version,
   options: Partial<HelperCapabilities> = {}
 ): HelperCapabilities {
-  if (import.meta.env.DEV && managerAPI !== '3.23') {
-    throw new Error('Invalid helper manager compatibility specified');
-  }
+  debugAssert(
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- JS-only check
```
wycats, line 26: An example of an "unnecessary" check that exists for JS-only Ember users.

packages/@glimmer/manager/lib/public/template.ts
```diff
@@ -1,33 +1,26 @@
 import type { TemplateFactory } from '@glimmer/interfaces';
 import { debugToString } from '@glimmer/debug-util';
+import { debugAssert } from '@glimmer/global-context';
```
wycats, line 3: `debugAssert`s *do* show up in the build output. They're still nicer than explicit `import.mneta.env.DEV` checks.

packages/@glimmer/manager/test/package.json
```diff
@@ -1,9 +1,12 @@
 {
   "private": true,
   "name": "@glimmer-test/manager",
+  "repo-meta": {
+    "strictness": "loose"
+  },
   "dependencies": {
-    "@glimmer/manager": "workspace:*",
     "@glimmer/interfaces": "workspace:*",
+    "@glimmer/manager": "workspace:*",
```
wycats, line 9: There's a decent amount of missing dependencies in test packages, which worked kind of by accident.
It's nice to document things properly.

packages/@glimmer/opcode-compiler/lib/opcode-builder/helpers/shared.ts
```diff
@@ -39,7 +39,7 @@ export function CompileArgs(
 
   if (atNames) flags |= 0b1000;
 
-  if (blocks) {
+  if (blocks.hasAny) {
```
wycats, line 42: I think this meant that all component invocations were tagged as having args. I'm not sure how much
that matters, but if we're going to do a check, we may as well do it properly.

packages/@glimmer/program/lib/program.ts
```diff
@@ -194,17 +194,7 @@ export class ProgramImpl implements Program {
 }
 
 function slice(arr: Int32Array, start: number, end: number): Int32Array {
-  if (arr.slice !== undefined) {
```
wycats, line 197: No supported browsers or engines are missing `.slice` anymore. This entire function could probably
be removed.

wycats, line 197: No supported engines are missing `slice` anymore. Further, this whole function only existed for
eager mode, which doesn't exist anymore.

packages/@glimmer/reference/test/utils/platform.ts
```diff
@@ -1,8 +0,0 @@
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
wycats, line 1: This function is never used

wycats (commented):
I've started documenting the reasons for changes in comments. Since a lot of the changes are repetitive, I only document the first time a change occurred in filename order.

NullVoxPopuli:
Can you fix the conflicts and the benchmark?

bin/package.json
NullVoxPopuli: this means we can start using publint and arethetypeswrong to ensure we don't have publishing mishaps!

NullVoxPopuli: yay

bin/setup-bench.mjs
```diff
@@ -1,7 +1,10 @@
...
+/* eslint-disable n/no-process-exit */
+import { readFile, writeFile } from 'node:fs/promises';
 import os from 'node:os';
 import { join } from 'node:path';
-import { readFile, writeFile } from 'node:fs/promises';
+
+import { ensureDirSync, writeFileSync } from 'fs-extra';
```
NullVoxPopuli, line 6: the default import for fs-extra is actually cjs, so you have to do:

```js
import fse from 'fs-extra';

const { ensureDirSync, writeFileSync } = fse;

```

guides/workspace/vscode.md
```diff
@@ -0,0 +1,94 @@
...
+This repository is configured to use [code
+actions](https://code.visualstudio.com/docs/editor/refactoring#_code-actions-quick-fixes-and-refactorings)
+to autofix ESLint issues and format the source code.
+
+The Format Code Action extension allows us to run prettier as a code action, so we can specify the
+order of formatting and lint fixing.
+
+> [!TIP]
+>
+> We intentionally don't configure ESLint to run prettier as an autofix, for [reasons described by
```
NullVoxPopuli, line 28: yaaaaas!

packages/@glimmer-workspace/build/lib/config.js
```diff
@@ -58,56 +56,65 @@ const EXTERNAL = true;
...
  */
-export function typescript(pkg, config) {
-  const typeScriptConfig = {
-    ...config,
-    paths: {
-      '@glimmer/interfaces': [resolve(pkg.root, '../@glimmer/interfaces/index.d.ts')],
-      '@glimmer/*': [resolve(pkg.root, '../@glimmer/*/src/dist/index.d.ts')],
+export function typescript(pkg) {
+  return [
+    rollupSWC({
```
NullVoxPopuli, line 85: noice

packages/@glimmer/compiler/lib/builder/builder-interface.ts
```diff
@@ -322,7 +322,7 @@ function normalizeVerboseStatement(statement: VerboseStatement): NormalizedState
 }
 
 function extractBlockHead(name: string): NormalizedHead {
-  const result = /^(#|!)(.*)$/u.exec(name);
+  const result = /^(#|!)(.*)$/v.exec(name);
```
NullVoxPopuli, line 325: why should any of these be using unicode?

image


This lint made everything slower

NullVoxPopuli: bench here: https://github.com/NullVoxPopuli/node-benchmarks/blob/main/tests/glimmer-vm-isElement.js

NullVoxPopuli: the `/u` variant is pretty close to native (sometimes)
image

packages/@glimmer/debug-util/lib/platform-utils.ts
```diff
@@ -16,7 +16,7 @@ export function unwrap<T>(val: Maybe<T>): T {
  * existing `LOCAL_DEBUG` check.
  */
 export function dev<T>(val: Optional<() => T>): T {
-  if (val === null || val === undefined) {
```
NullVoxPopuli, line 19: is it no longer possible for val to be null?

packages/@glimmer/destroyable/index.ts
```diff
@@ -159,11 +161,17 @@ export function destroy(destroyable: Destroyable) {
   meta.state = DESTROYING_STATE;
 
   iterate(children, destroy);
-  iterate(eagerDestructors, (destructor) => destructor(destroyable));
-  iterate(destructors, (destructor) => scheduleDestroy(destroyable, destructor));
+  iterate(eagerDestructors, (destructor) => {
```
NullVoxPopuli, line 164: why add bodies to these functions?

packages/@glimmer/compiler/lib/builder/builder-interface.ts
NullVoxPopuli: finally found a tool that expresses the information in a way I like:
image

packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts
```diff
@@ -123,11 +132,16 @@ export abstract class HandlebarsNodeVisitors extends Parser {
 
     // Backfill block params loc for the default block
     let blockParams: ASTv1.VarHead[] = [];
+    let repairedBlock: HBS.BlockStatement;
 
     if (block.program.blockParams?.length) {
       // Start from right after the hash
       let span = hash.loc.collapse('end');
 
+      // if (!block.program.loc) {
```
NullVoxPopuli, line 141: needs cleanup

NullVoxPopuli (changes requested):
Some minor things left:
- conflicts
- benchmark doesn't run
- https://github.com/glimmerjs/glimmer-vm/pull/1677#discussion_r1917078269 -- looks like we want to use either `/u` or no flag for runtime regex if we don't need to explicitly be matching _for_ unicode (unicode characters aren't valid elements, ya?)
- leftover commented out code: https://github.com/glimmerjs/glimmer-vm/pull/1677#discussion_r1917226979

bin/setup-bench.mjs
wycats: Ooh good point!

bin/package.json
wycats: Yes exactly. This PR also sets up publint (for all published packages) and publint *should* be clean
right now.

packages/@glimmer/destroyable/index.ts
wycats: I could have used the `void` prefix, but TL;DR if the expression returns non-void, the stricter lint settings don't let you implicitly return the value via an arrow.

packages/@glimmer/debug-util/lib/platform-utils.ts
wycats: It was never possible based on the type, and this code is only used internally by other typed code.

packages/@glimmer/compiler/lib/builder/builder-interface.ts
wycats: I think `/u` is a good general practice (because non-unicode regexes have surprising behavior and `/u` is a pretty free-and-clear upgrade without huge compat headaches). See https://eslint.org/docs/latest/rules/require-unicode-regexp.

`/v` is also touted as a free-and-clear upgrade to `/u`, but it's quite new, seems to have perf implications, and is a huge change to the character class syntax. With time, I think it will be the right default, but I agree that that time is not now, and plan to revert this change and update the lint.

packages/@glimmer-workspace/build/lib/config.js
wycats: I thought you would like it 😄

NullVoxPopuli:
Current performance boost on my lil'linux laptop:
```
duration phase estimated improvement -383ms [-528ms to -245ms] OR -7.39% [-10.18% to -4.72%]
renderEnd phase estimated improvement -1ms [-2ms to 0ms] OR -5.11% [-11.49% to -0.12%]
render1000Items1End phase no difference [-30ms to 4ms]
clearItems1End phase no difference [-5ms to 3ms]
render1000Items2End phase estimated improvement -17ms [-35ms to -1ms] OR -6.19% [-12.67% to -0.19%]
clearItems2End phase no difference [-2ms to 2ms]
render5000Items1End phase estimated improvement -117ms [-165ms to -79ms] OR -8.74% [-12.4% to -5.92%]
clearManyItems1End phase no difference [-5ms to 9ms]
render5000Items2End phase estimated improvement -114ms [-150ms to -69ms] OR -9.61% [-12.68% to -5.87%]
clearManyItems2End phase no difference [-16ms to 0ms]
render1000Items3End phase no difference [-42ms to 2ms]
append1000Items1End phase no difference [-58ms to 1ms]
append1000Items2End phase estimated improvement -23ms [-47ms to -2ms] OR -7.43% [-15.36% to -0.65%]
updateEvery10thItem1End phase no difference [-11ms to 1ms]
updateEvery10thItem2End phase no difference [-10ms to 0ms]
selectFirstRow1End phase no difference [-11ms to 0ms]
selectSecondRow1End phase no difference [-16ms to 0ms]
removeFirstRow1End phase no difference [-15ms to 1ms]
removeSecondRow1End phase no difference [-17ms to 0ms]
swapRows1End phase no difference [-1ms to 1ms]
swapRows2End phase no difference [-1ms to 0ms]
clearItems4End phase no difference [-3ms to 5ms]
paint phase no difference [0ms to 1ms]
```

this is very nice!

I wonder how much has to do with there just being less code in the output now

turbo.json
```diff
@@ -21,8 +21,7 @@
     "packages/@glimmer-workspace/build/lib/config.js"
   ],
   "tasks": {
-    "build": {
-      "dependsOn": ["^build"],
+    "prepare": {
```
NullVoxPopuli, line 24: prepack?

wycats: Yeah, I already figured that out :)

packages/@glimmer/syntax/package.json
```diff
@@ -2,38 +2,33 @@
...
+    "url": "git+https://github.com/glimmerjs/glimmer-vm.git",
+    "directory": "packages/@glimmer/syntax"
+  },
   "type": "module",
   "exports": "./index.ts",
   "publishConfig": {
     "access": "public",
     "exports": {
-      ".": {
-        "require": {
```
NullVoxPopuli, line 12: removing require here breaks prettier.
let's fix in a followup though

NullVoxPopuli (approved):

PR #1682, opened 2025-01-19
[ Pending RFC ] create Cell

NullVoxPopuli:
I want to RFC Resources, but we don't really have an implemented way to have reactivity in them right now.

We have:
 - `trackedData` in `@glimmer/validator` -- this whole package should not be used by ember users tho, and most of its exports are not-RFC'd and should be considered private API 
 - `@glimmer/tracking/primitives/storage` -- this would work as an implementation of "Cell", but isn't implemented yet
 - `@tracked` - requires creating class state, which many would feel is too cumbersome / verbose when for the "simple" cases, like having one value

### DIFF
--- a/packages/@glimmer/validator/lib/cell.ts
+++ b/packages/@glimmer/validator/lib/cell.ts
@@ -1,0 +1,54 @@
+import type { UpdatableTag } from '@glimmer/interfaces';
+
+import { consumeTag } from './tracking';
+import { createUpdatableTag, DIRTY_TAG } from './validators';
+
+type Equality<T> = (a: T, b: T) => boolean;
+
+export class Cell<Value> {
+  static create<T>(value: T, equals: Equality<T> = Object.is): Cell<T> {
+    return new Cell(value, equals);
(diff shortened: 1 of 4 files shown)


packages/@glimmer/validator/lib/cell.ts
```diff
@@ -0,0 +1,54 @@
...
+    return this.#value;
+  }
+
+  read(): Value {
+    consumeTag(this.#tag);
+
+    return this.#value;
+  }
+
+  set(value: Value): boolean {
```
josemarluedke, line 35: Would it be worth making the set a bounded func? For example:
```ts
  set = (value: Value) =>  boolean {
```

This way we could pass this into templates and not need to create a func just to set the value of a cell.

NullVoxPopuli: how do you mean?

it's a bug in the vm that this doesn't work:
```hbs
<button {{on 'click' (fn cell.set 'hi')}}></button>
```

PR #1689, opened 2025-01-23
Prevent #1688 (and other accidental prettier breakages)

### LINKED ISSUE #1688: `@glimmer/syntax` 0.94.0 is broken
```
> (await import('@glimmer/syntax')).preprocess('<h1></h1>')
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at P (<CWD>/node_modules/@glimmer/syntax/dist/prod/index.js:368:12)
    at U.matchFor (<CWD>/node_modules/@glimmer/syntax/dist/prod/index.js:436:12)
```

NullVoxPopuli:
Fixes: https://github.com/glimmerjs/glimmer-vm/issues/1688

This repo is going through a lot of infra changes, so because we haven't had any integration tests with prettier (since they consume `@glimmer/syntax` for hbs and glimmer parsing), we've accidentally broke them now 2 or 3 times recently 😬 

Takes the same approach as: 
- https://github.com/emberjs/data/pull/9633

Where we use our real build in a real project outside the monorepo.
Every shortcut we take to make in-repo development easier needs to have extra testing to ensure we don't mess up publishing.

Both ember-data/warp-drive and glimmer-vm are not testing the code they ship to NPM.

--------------------

In debugging, I copied the prod index.js and formatted it with prettier so that I could debug it:
```js
function P(t) {
  return t.length > 0;
}
```
via:
```js
console.log((await import('./syntax-prod.js')).preprocess('<h1></h1>'));
```
and the VS Code JavaScript Debug Terminal


package.json
```diff
@@ -37,7 +37,7 @@
     "test:babel-plugins": "yarn workspace @glimmer/vm-babel-plugins test",
     "test:browserstack": "ember test --test-port=7774 --host 127.0.0.1 --config-file=testem-browserstack.js",
     "test:lint": "eslint . --quiet",
-    "test:node": "node bin/run-node-tests.mjs",
```
NullVoxPopuli, line 40: nothing was calling `test:node` prior to this PR, and run-node-tests called global `ember`, which isn't used in CI

package.json
```diff
@@ -37,7 +37,7 @@
...
+    "test:node": "pnpm -r test:node",
```
wycats, line 40: Shouldn't this still go through Turbo? Right now, we only have one node test, but putting it in turbo would allow us to add package-local node tests easily if we wanted to.

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * See: https://github.com/glimmerjs/glimmer-vm/issues/1688
+ *
+ * Requires the root package.json#pnpm#overrides point at our internal
+ * copy of @glimmer/syntax, or else prettier brings its own already published
+ * copy of @glimmer/syntax
+ */
+describe('Prettier', () => {
+  it(`SANITY: we've symlinked to the in-repo copy of @glimmer/syntax`, () => {
```
wycats, line 21: I try to use the term "SMOKE" rather than "SANITY" in general.

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,207 @@
...
+const require = createRequire(import.meta.url);
+
+/**
+ * See: https://github.com/glimmerjs/glimmer-vm/issues/1688
+ *
+ * Requires the root package.json#pnpm#overrides point at our internal
+ * copy of @glimmer/syntax, or else prettier brings its own already published
+ * copy of @glimmer/syntax
+ */
+describe('Prettier', () => {
```
wycats, line 20: Love having formal prettier tests.

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      wireFormat: `glimmer-wire-format-${await versionOf('@glimmer/wire-format')}`,
+    };
+
+    let file = `console.log((await import('@glimmer/syntax')).preprocess('<h1></h1>'));`;
+    let manifest = JSON.stringify({
+      name: 'real-test',
+      type: 'module',
+      private: true,
+      devDependencies: {
+        '@glimmer/syntax': `file://./${tarballs.syntax}`,
```
wycats, line 78: Should we be consolidating this with the equivalent work I did for the benchmark? It would be awesome to have reusable infra for making tests with the published packages, and we need it already for benchmarks. (not a blocker, and I could do the follow-up work if you think it's a good idea).

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * pnpm packs tgz's with the name format ${hyphenated-package-name}-${version}.tgz
+     */
+    async function versionOf(name) {
+      let manifestPath = join(monorepoRoot, 'packages', name, 'package.json');
+      let manifestString = await readFile(manifestPath);
+      return JSON.parse(manifestString.toString()).version;
+    }
+
+    let tarballs = {
```
wycats, line 66: One thing to note about the benchmark setup: I ran `pnpm pack` with an `--out` that removed the version. In that case, it allowed me to have a checked-in `package.json` that wouldn't need to be changed as the versions change. In general, it seems nice to avoid having to care about the version by creating a version of the tarball without the version in the first place.

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    //////////
+    // 3 create a project that represents real consumer usage
+    await writeFile(join(tmp, 'package.json'), manifest);
+    await writeFile(join(tmp, 'index.js'), file);
+
+    //////////
+    // 4 install the tarballs using stable names so we don't have to
+    //   dynamically build the package.json
+    let packToTemp = `pnpm pack --pack-destination ${tmp}`;
```
wycats, line 135: Yeah, I see. This is why I bumped to `pnpm@10`. It's [not documented](https://pnpm.io/cli/pack), but they [added an `--out` flag](https://github.com/pnpm/pnpm/pull/8900) in pnpm 10, which makes it possible to pack directly into the exact file name you want without hassle.

packages/@glimmer/debug-util/lib/present.ts
```diff
@@ -14,8 +14,8 @@ export function assertPresent<T>(value: T, message?: string): asserts value is P
   }
 }
 
-export function isPresentArray<T>(list: readonly T[]): list is PresentArray<T> {
-  return list.length > 0;
+export function isPresentArray<T>(list?: readonly T[]): list is PresentArray<T> {
```
wycats, line 17: Hm... what uses `isPresentArray` with an optional array?

NullVoxPopuli: nothing -- it's a minification bug!

input:
```ts
  private matchFor(
    left: OffsetKind,
    right: OffsetKind
  ): (left: PositionData, right: PositionData) => Out {
    const nesteds = this._whens.match(left);

    localAssert(
      isPresentArray(nesteds),
      `no match defined for (${left}, ${right}) and no AnyMatch defined either`
    );

    const callback = new WhenList(nesteds).first(right);

    localAssert(
      callback !== null,
      `no match defined for (${left}, ${right}) and no AnyMatch defined either`
    );

    return callback;
  }
```

output: 
```js
 matchFor(t, e) {
    const s = this._whens.match(t);
    return P(), new I(s).first(e);
  }
```

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
NullVoxPopuli: I like this plan, would clean up my test a bit. will update

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
NullVoxPopuli: I like this plan, would clean up my test a bit. will update

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
NullVoxPopuli: updated!

package.json
NullVoxPopuli: updated!

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  it('Underlynig preprocess API works', async () => {
+    let result = (await import('@glimmer/syntax')).preprocess('<h1></h1>');
+
+    expect(result, `It can be await import()'d, and doesn't error`).toBeTruthy();
+  });
+
+  it('Prettier can call preprocess', async () => {
+    let result = await prettier.format(`     <div>\n</div>`, { parser: 'glimmer' });
```
fisker, line 34: This PR has been settled for a long time, sorry that I didn't see this earlier, but this test is meaningless, since Prettier uses a bundled version of `@glimmer/syntax`.

If we want a reliable test, we need

1. Clone prettier/prettier
2. Upgrade `@glimmer/*` to use the changed version
3. Run Prettier test suite.

packages/@glimmer-workspace/integration-node-tests/test/prettier.test.ts
```diff
@@ -0,0 +1,41 @@
...
+ * copy of @glimmer/syntax
+ *
+ * NOTE: that this test alone is insufficient to test our built outputs.
+ *       the smoke-tests/* folders are for that purpose.
+ */
+describe('Prettier', () => {
+  it(`SMOKE: we've symlinked to the in-repo copy of @glimmer/syntax`, () => {
+    let workspacePath = require.resolve('@glimmer/syntax');
+    let prettierPath = require.resolve('prettier');
+    let prettierGlimmer = require.resolve('@glimmer/syntax', { paths: [prettierPath] });
```
fisker, line 22: This is not useful, since prettier bundles all dependencies. https://www.npmjs.com/package/prettier?activeTab=dependencies

NullVoxPopuli: Ah i see. I'll use the clone approach you suggested earlier

fisker: Or you can install `prettier@prettier/prettier` (our codebase can be installed and run directly), if you only want run a simple test.

PR #1690, opened 2025-01-23
[WIP] Begin moving towards invokable wire format

wycats:


reteps:
Any way that I can contribute? In the context of #1666, this would be very helpful.

NullVoxPopuli:
idk, basically what's left here is that I need to add tests, and make sure CI goes green

wycats:
> I hope I'm wrong, but this PR doesn't seem to be headed into that direction (yet?).

@vlascik "(yet?)" is right. I'm trying to do this a little at a time, and even this incremental piece has been a somewhat significant project.

That said, the goal here is to get to a point where, first, the wire format itself is not needed. Next, we'd migrate more of the opcode compiler to vanilla-js (although I think you might be underestimating what's needed to make "vanilla js" support updating). If we can get to the end of _this_ PR, I'll feel pretty optimistic about the direction you're suggesting.

And apologies for how long it's taken!

PR #1717, opened 2025-02-21
Fix source slicing for whitespace-stripping comments

chancancode:
Re-apply #1430 with a more robust implementation and tests

Also turns the warning into a `localAssert()` so it gets stripped before reaching downstream consumers.

This relates to #1431 in that this removes the warning, so to the extent that #1431 was motivated by sliencing the warning in Ember builds, it avoids the issue. But I haven't investigated the root cause of that issue and whether it is indicative of an actual bug somewhere deeper.

Fixes emberjs/ember.js#19392


NullVoxPopuli:
https://github.com/glimmerjs/glimmer-vm/actions/runs/13466445316/job/37633202706?pr=1717#step:4:5753

image


Same as on main, so we're good here :tada: 

thanks for fixing this!

NullVoxPopuli (approved):

PR #1722, opened 2025-02-28
Printer bug: empty string literal args are dropped

ef4:
The printer (in all modes, not just the hacked-on "codemod mode") loses empty string literal arguments:

```diff
-<Hello @world="" />
+<Hello @world />
```

### DIFF
--- a/packages/@glimmer/syntax/lib/generation/printer.ts
+++ b/packages/@glimmer/syntax/lib/generation/printer.ts
@@ -285,7 +285,9 @@ export default class Printer {
     let { name, value } = attr;
 
     this.buffer += name;
-    if (value.type !== 'TextNode' || value.chars.length > 0) {
+    const isAttribute = !name.startsWith('@');
+    const shouldElideValue = isAttribute && value.type == 'TextNode' && value.chars.length === 0;
+    if (!shouldElideValue) {
       this.buffer += '=';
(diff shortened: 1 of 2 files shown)


ef4:
Followup investigation: the existing behavior isn't strictly incorrect, since a missing attribute or argument value is treated as the empty string.

There's not an obvious best thing to do here in all cases.

1. The printer doesn't currently make any distinction between arguments and attributes. It's not clear that it should, since at the syntax level they behave identically.

2. If we continue to print `@foo=""` as `@foo`, that runs afoul of a default [template-lint rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-valueless-arguments.md).

3. But if we go the other direction and normalize attributes to always show the empty string literal, that would change `<input disabled >` to `<input disabled="">`, which is also annoying.

To make both those cases do the expected thing, we'd need to make the printer treat attributes and arguments differently. Maybe that's OK? It seems like the least bad near-term option.

Ultimately we would prefer to have a printer designed from the ground up for format preservation, but that would also require changes all the way down through the glimmer parser and handlebars parser, because information gets lost in both those layers.

ef4:
I'm very confused about how I can have a seemingly-unrelated CI failure here that isn't reproducing on main.

PR #1742, opened 2025-03-18
No way it's this easy

NullVoxPopuli:
another thing from Devin. Kinda crazy it has so much confidence and ignores the tests so frequently, when repeatedly told not to ignore the tests.

Session here: https://app.devin.ai/sessions/5d35b429dc134491937a68c89a4e8e70


NullVoxPopuli:
so, with some more poking around -- _it is_ easy to implement via resolvers, but I don't think we want to implement via resolvers.

PR #1751, opened 2025-05-14
prove that per-item reactivity can work

NullVoxPopuli:
Noticed a potential issue in:
- https://github.com/glimmerjs/glimmer-vm/pull/1748

But since this test works, it means the issue is in the trackedArray implementation

### DIFF
--- a/packages/@glimmer-workspace/integration-tests/test/keywords/each-test.ts
+++ b/packages/@glimmer-workspace/integration-tests/test/keywords/each-test.ts
@@ -1,3 +1,4 @@
+import { consumeTag, createUpdatableTag, dirtyTag } from '@glimmer/validator';
 import {
   defineComponent,
   jitSuite,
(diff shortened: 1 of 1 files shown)


packages/@glimmer-workspace/integration-tests/test/keywords/each-test.ts
```diff
@@ -30,6 +31,39 @@ class Each extends RenderTest {
 
     this.assertHTML('  ');
   }
+
+  @test
+  'each with array of tags is reactive per item'(assert: Assert) {
+    let array = [createUpdatableTag(), createUpdatableTag()];
```
lifeart, line 37: I'm not sure we testing array here, because we dirty only items...
I think if we need to verify atomic array reactivity we may need to mutate array values by indexes and assert for amount of opcode operations in list-sync logic.

packages/@glimmer-workspace/integration-tests/test/keywords/each-test.ts
```diff
@@ -30,6 +31,39 @@ class Each extends RenderTest {
...
+    );
+
+    this.renderComponent(Foo);
+    this.assertHTML(' 0  1 ');
+    assert.verifySteps(['0', '1']);
+
+    dirtyTag(array[0]!);
+    this.rerender();
+    this.assertHTML(' 0  1 ');
+    assert.verifySteps(['0']);
```
lifeart, line 60: I think steps count should be `0` (empty array), for following reasons: `item` not used inside block scope, so, there is no tag should be created for it, and because array is not changed, nothing should be recomputed. It may be recomputed in only one case: `item` used inside `step` and `step` consume tag. Like: `{{step item index}}`

PR #1771, opened 2025-09-05
Bugfix: support Array as parent in associateDestroyableChild

ef4:
The included test throws an assertion without the accompanying bugfix.

The issue here is that the DestroyableMeta is trying to be clever about avoiding extra allocations, so it uses the OneOrMany type a lot. But this means it cannot accurately distinguish between the case where your parent is an array vs the case where you have multiple parents.

My fix adds a Symbol to the internally-created array that represents multiple parents so it's distinguishable from a single parent that happens to be an array.

### DIFF
--- a/packages/@glimmer/destroyable/index.ts
+++ b/packages/@glimmer/destroyable/index.ts
@@ -7,7 +7,7 @@ const DESTROYING_STATE = 1;
 const DESTROYED_STATE = 2;
 type DestroyableState = 0 | 1 | 2;
 
-type OneOrMany<T> = null | T | T[];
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
We'll need https://github.com/glimmerjs/glimmer-vm/pull/1772 first :see_no_evil: 

Silly deprecation means non-working forever thing :melting_face:

NullVoxPopuli:
https://github.com/glimmerjs/glimmer-vm/actions/runs/17483617066/job/49658813526#step:6:552

Oof bit rot

